PR #300, opened 2025-05-19
refactor(inkless): pipelining committer threading

jeqo:
At the moment, the threads are managed by the jobs. Specially, the commit job blocks on upload future.
Instead we can use CompletableFuture to compose the thread coordination. This should help with better resource utilization.


jeqo:
WriteIntegrationTest is getting flaky which may indicate the proposed approach is breaking ordering on the response. Have to look a bit deeper into this.

PR #301, opened 2025-05-19
refactor(inkless): avoid alloc by passing bytebuffer around

jeqo:
This removes the need to allocate the bytes. It passes byteBuffer to uploader, so a ByteBufferInputStream can be used to upload; and passes a byteBuffer to the cache, so a slice array can be passed to the cache.


ivanyu:
@jeqo this seems to need rebasing

PR #324, opened 2025-06-12
Introduce materialize log size [INK-251]

ivanyu:
This commit introduces a new piece of materialized state of a log: its size in bytes. While them are usually uncomfortable to have and work with (too easy to forget to update, etc.), this one is, unfortunately, necessary. Without it, retention-by-size would be too costly to perform.

The commit also introduces the control plane function to get the log information. It's used for testing, but in future will be useful for observability.


storage/inkless/src/main/resources/db/migration/V3__Add_logs_size.sql
```diff
@@ -0,0 +1,300 @@
...
+SET byte_size = aggregated.total_byte_size
+FROM (
+    SELECT topic_id, partition, SUM(byte_size) AS total_byte_size
+    FROM batches
+    GROUP BY topic_id, partition
+) AS aggregated
+WHERE logs.topic_id = aggregated.topic_id
+    AND logs.partition = aggregated.partition;
+
+-- 2. Update functions to support `byte_size`.
```
ivanyu, line 18: There are two ways to affect the size of the log (not taking into account the deletion of the whole log, because in this case the very record of it is deleted from the DB):
1. Append records/batches, done by `commit_file_v1`.
2. Delete records/batches, done by `delete_records_v1`.

storage/inkless/src/main/resources/db/migration/V3__Add_logs_size.sql
```diff
@@ -0,0 +1,300 @@
...
+                        AND producer_id = l_request.producer_id
+                    ORDER BY row_id DESC
+                    LIMIT 1
+                    OFFSET 5
+                );
+        END IF;
+
+        UPDATE logs_tmp
+        SET high_watermark = high_watermark + (l_request.last_offset - l_request.base_offset + 1),
+            byte_size = byte_size + l_request.byte_size
```
ivanyu, line 165: An essential change

PR #327, opened 2025-06-18
Fix deadlock in delete_records_v1 [INK-260]

ivanyu:
The function did the deadlock-preventing ordering on a wrong query.

The job test was enhanced to check different request orders work correctly.

`EnforceRetentionJob` already had the code for processing results of such a PG function, which was generalized to serve both.

### DIFF
--- a/storage/inkless/src/main/resources/db/migration/V5__Fix_deadlock_in_delete_records_v1.sql
+++ b/storage/inkless/src/main/resources/db/migration/V5__Fix_deadlock_in_delete_records_v1.sql
@@ -1,0 +1,90 @@
+-- Copyright (c) 2025 Aiven, Helsinki, Finland. https://aiven.io/
+
+CREATE OR REPLACE FUNCTION delete_records_v1(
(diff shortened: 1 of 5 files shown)


storage/inkless/src/main/resources/db/migration/V5__Fix_deadlock_in_delete_records_v1.sql
```diff
@@ -0,0 +1,90 @@
...
+    DROP TABLE IF EXISTS affected_files;
+    CREATE TEMPORARY TABLE affected_files (
+        file_id BIGINT PRIMARY KEY
+    )
+    ON COMMIT DROP;
+
+    FOR l_request IN
+        SELECT *
+        FROM unnest(arg_requests)
+        ORDER BY topic_id, partition  -- ordering is important to prevent deadlocks
```
ivanyu, line 24: This line is the main change in this PR, moved from the `SELECT` below. The rest of the function is unchanged.

PR #337, opened 2025-06-24
refactor(inkless): add artificial latency to metadata update

jeqo:
When refreshing metadata, new requests to next broker may be processed _before_ the last request on the previous broker, causing OutOfOrderSequence.
These errors are retried and data finally lands in order. However to avoid this error, this proposal adds artificial latency to the metadata response when it includes inkless topics. This way it gives enough time for the previous broker to complete processing requests, before the new one takes on.


storage/inkless/src/main/java/io/aiven/inkless/metadata/InklessTopicMetadataTransformer.java
```diff
@@ -65,6 +70,11 @@ public void transformClusterMetadata(
                 partition.setLeaderEpoch(LeaderAndIsr.INITIAL_LEADER_EPOCH);
             }
         }
+        if (hasInklessTopics) {
+            // Introduce artificial latency to avoid a race condition between the metadata update and the producer
+            // causing OutOfOrderSequenceException.
+            time.sleep(500);
```
ivanyu, line 76: I'm not sure it's a good idea for several reasons:
1. I don't see how waiting for fixed delay of 500 ms may solve the problem... As produce and metadata updates happen asynchronously, they may interleave in arbitrary order with or without the sleep.
2. Blocking the handler thread isn't great.

I think the problem is theoretically possible on classic Kafka, too, it's juts unlikely because of more leadership stability + the `NOT_LEADER_OR_FOLLOWER` error. How bad would be it if we just do nothing and rely on existing recovery mechanism?

jeqo: > As produce and metadata updates happen asynchronously, they may interleave in arbitrary order with or without the sleep.

From what I found, metadata updates are not async (e.g. https://github.com/apache/kafka/blob/3c1f965c60789dcc8ee14ebabcbb4e16ebffc5ee/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L642) hence 500ms (choose arbitrarily but aiming to cope with a rotation on the broker side, default 250ms) would give room to avoid the race condition; though I'd agree any other implementation could potentially handle this update asynchronously and this solution may not be enough.

> I think the problem is theoretically possible on classic Kafka, too

I don't think it is with partition leaders. The producer state is cached on the leader and if leadership changes, it starts accepting requests after updating the state.

> How bad would be it if we just do nothing and rely on existing recovery mechanism?

OutOfOrderSequence errors are retried: https://github.com/apache/kafka/blob/6f783f85362071f82da3dcef706c7e6b89b86c2a/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java#L829-L834
We could just document that this error is expected and retried, and users could safely ignore it on their logs.
Another side effect we have observed is that retries lead to latency spikes (as the retry is handled within producer machinery so from the request time it just takes longer) -- though again this could be documented.

PR #342, opened 2025-07-07
Inkless generated config docs

juha-aiven:
Create inkless config HTML documentation and include it the docs.

Screenshot 2025-07-07 at 9 38 17

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1196,7 +1196,8 @@ project(':core') {
                                ':connect:runtime:genConnectMetricsDocs', ':connect:runtime:genConnectOpenAPIDocs',
                                ':connect:mirror:genMirrorSourceConfigDocs', ':connect:mirror:genMirrorCheckpointConfigDocs',
                                ':connect:mirror:genMirrorHeartbeatConfigDocs', ':connect:mirror:genMirrorConnectorConfigDocs',
-                               ':storage:genRemoteLogManagerConfigDoc', ':storage:genRemoteLogMetadataManagerConfigDoc'], type: Tar) {
+                               ':storage:genRemoteLogManagerConfigDoc', ':storage:genRemoteLogMetadataManagerConfigDoc',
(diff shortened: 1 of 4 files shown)


jeqo (commented):
Thanks for looking into this! Unfortunately, unless we host the HTML site somewhere, it would be very hard to navigate this. 

Instead, we could use RST that is rendered on GitHub (so, no need to host it). We did similar on Tiered Storage plugin: https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/blob/21795a6f4b214fe671bb9406eb5b9949bcbdb38e/docs/src/main/java/io/aiven/kafka/tieredstorage/misc/ConfigsDocs.java#L49

Could you have a look and see if you could have a similar approach here?

juha-aiven:
I'll check if there's a simple way we can host the docs site somewhere. If that proves to be too complicated, I understand what you mean with generating the docs and serving from GitHub.

I think the diff between inkless and and TS is that inkless is part of `kafka` and thus, it's natural to have inkless documentation embedded in the kafka docs. `tiered-storage-for-apache-kafka` is not part of kafka and it's docs need to be "elsewhere".

jlprat:
The problem is that Apache Kafka docs are relying on Apache Server's rewrite rules, so it would be hard to host this without a separated repo and an Apache Server running.

juha-aiven:
OK, I see now. Need to rewind back 20 years. Things like this

```
<!--#include virtual="generated/inkless_config.html" -->
```

Are processed by the Apache HTTP response module `mod_include`. We'd need to host that server somewhere, build the necessary CI etc.

So, the effort / benefit ratio of using the native Kafka documentation mechanism for inkless doesn't look good at all. Right now. If one day merged to `apache/kafka`, then yes. And if the docs were at that point already in the "correct" location, less work would be needed. But that's a future problem not worth dealing with now.

PR #343, opened 2025-07-08
Create configs.rst

juha-aiven:
Add Gradle task :storage:inkless:genInklessConfigDoc that generates the file docs/inkless/configs.rst documenting Inkless-specific configuration parameters. The task executes aiven.inkless.doc.ConfigsDocs.

Make :storage:inkless:build depend on genInklessConfigDoc to automatically generate the config.rst.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2468,6 +2468,19 @@ project(':storage:inkless') {
     outputs.dir("${projectDir}/build/generated/main/java/io/aiven/inkless/generated")
   }
 
+  task genInklessConfigDoc(type: JavaExec) {
+    classpath = sourceSets.main.runtimeClasspath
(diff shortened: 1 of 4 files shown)


juha-aiven:
> This a good starting point. I wonder how we could add the Topic config as well.
> Maybe let's add a CONFIG.md that documents manually the topic configs (inkless.enable and server default log.inkless.enable), and links to the broker-configs configs.rst; wdyt?

Sounds quite good. I'll have a look at this. Also, the metrics need to be documented, I'll check that too.

jeqo (approved):

jlprat:
I'm late to the party, but why an `rst` file and not and `md` one?

PR #344, opened 2025-07-08
Create docs/inkless/topic_configs.rst

juha-aiven:
Add a new class TopicConfigsDocs to document Inkless topic-level configurations.

Add genInklessTopicConfigDoc Gradle task and make build depend on it.

Note: in order to add now Inkless-specific topic configs, the config needs to be manually added in TopicConfigsDocs.

Also note, the server level default key is manually specified in the inkless.enable topic config key documentation. It's not very straightforward to automatically create that. The rst-file generator doesn't seem to support that right now. The HTML creator would likely work, if LogConfigDef was used instead of ConfigDef.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2481,6 +2481,19 @@ project(':storage:inkless') {
     }
   }
 
+  task genInklessTopicConfigDoc(type: JavaExec) {
+    classpath = project(':storage:inkless').sourceSets.main.runtimeClasspath
+    mainClass = 'io.aiven.inkless.doc.TopicConfigsDocs'
+
+    // Define the outputs formally
+    outputs.file("$rootDir/docs/inkless/topic_configs.rst")
+
+    // Set up the output in the execution phase, not configuration, and avoid removing on clean
+    doFirst {
+      standardOutput = new File("$rootDir/docs/inkless/topic_configs.rst").newOutputStream()
+    }
+  }
+
   tasks {
     generateJooqClasses {
       withContainer {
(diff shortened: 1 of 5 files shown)


storage/inkless/src/main/java/io/aiven/inkless/config/InklessTopicConfig.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package io.aiven.inkless.config;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.common.config.TopicConfig;
+
+/**
+ * Configuration definitions for Inkless topic-level properties.
+ */
+public class InklessTopicConfig {
```
jeqo, line 26: Not a bad idea to have this to generate only the inkless topic config docs, but maybe let's have this within `TopicConfigsDocs` instead to avoid confusing that this may be used as the actual config.

Also, somehow let's add the server-side topic defaults? Maybe it's enough to update `INKLESS_ENABLE_DOC` to include that the default is defined at the server level with log.inkless.enable

juha-aiven: Everything's now in `TopicConfigsDocs`.

The server default is manually mentioned in the `inkless.enable` documentation. It seems more difficult than it should be to automatically create the "Server Default Property" in the documentation. The mechanism in the codebase seems to support creating that for HTML but not for rst.

PR #346, opened 2025-07-10
Inkless storage backend metrics

juha-aiven:
Add a tool that creates `docs/inkless/metrics.rst`.

To get started add the storage backend metrics there. Upcoming PRs will add more metrics.

Some manual testing. In the image below on left the generated file, the AWS S3 section and on the right in a screenshot of Jconsole. The metrics seem to match.

Screenshot 2025-07-10 at 16 54 07


clients/src/main/java/org/apache/kafka/clients/consumer/internals/SensorBuilder.java
juha-aiven, line 1: I wonder if it's a good idea to depend on this one. Or should we resort to copy-pasting?

jeqo (approved):

PR #361, opened 2025-07-21
fix broken tests after changes in DelayedFetch

giuseppelillo:
Fix the following tests:

DelayedFetchTest.testDelayedFetchWithMessageOnlyHighWatermark endOffset=0
DelayedFetchTest.testDelayedFetchWithMessageOnlyHighWatermark endOffset=500
DelayedFetchTest.testDivergingEpoch()
DelayedFetchTest.testFetchWithFencedEpoch()
DelayedFetchTest.testNotLeaderOrFollower()
ReplicaManagerQuotasTest.testCompleteInDelayedFetchConsumerFetch()
ReplicaManagerQuotasTest.testCompleteInDelayedFetchWithReplicaThrottling()

### DIFF
--- a/core/src/main/scala/kafka/server/DelayedFetch.scala
+++ b/core/src/main/scala/kafka/server/DelayedFetch.scala
@@ -170,9 +170,9 @@ class DelayedFetch(
       val fetchOffset = fetchStatus.startOffsetMetadata
       new FindBatchRequest(topicIdPartition, fetchOffset.messageOffset, fetchStatus.fetchInfo.maxBytes)
     }
+    if (requests.isEmpty) { return Some(0) }
 
     val response = replicaManager.findInklessBatches(requests, Int.MaxValue)
-    if (response.isEmpty) { return Some(0) }
     val r = response.get.asScala
     r.foreach { response =>
       response.errors() match {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/server/DelayedFetch.scala
```diff
@@ -170,9 +170,9 @@ class DelayedFetch(
       val fetchOffset = fetchStatus.startOffsetMetadata
       new FindBatchRequest(topicIdPartition, fetchOffset.messageOffset, fetchStatus.fetchInfo.maxBytes)
     }
+    if (requests.isEmpty) { return Some(0) }
 
     val response = replicaManager.findInklessBatches(requests, Int.MaxValue)
-    if (response.isEmpty) { return Some(0) }
```
giuseppelillo, line 175: Useless because if the response is empty nothing will be executed anyways

core/src/test/scala/integration/kafka/server/DelayedFetchTest.scala
```diff
@@ -225,6 +237,67 @@ class DelayedFetchTest {
     }
   }
 
+  @Test
+  def testWithInkless(): Unit = {
```
giuseppelillo, line 241: New test that checks that min bytes are accumulated both with classic and inkless partitions

PR #367, opened 2025-08-13
refactor(inkless): reduce the control-plane's findBatch on inkless fetching path

jeqo:
This PR includes a set of refactoring changes to reduce the findBatch fan-out when consuming from Inkless topics, and expands the test coverage and comments to clarify how classic and inkless topics are fetched together.

It reduces the fan-out by:
- Removing the initial inkless find fetch, and only executing find fetches through delayed fetch
- Not trying to complete on delayed fetch on appends completion to avoid unbounded number of find fetch requests
- Piggy-backing on the remote.fetch.max.wait.ms to enforce a lower-bound on the fetch timout. This should reduce how frequent fetches would be when no much throughout to serve requests; therefore reducing the find fetch requests.

When applied, this change will reduce consumer fetch rate overall for inkless topics, and reduce the fan-out to up to two find fetches, all within the DelayedFetch:
- One on the first try
- One on the force complete, as part of actually serving the fetch request.


AnatolyPopov (approved):
LGTM

PR #376, opened 2025-08-22
fix(inkless:ci): regenerate jooq classes with jdk23 on kafka ci

jeqo:
Had to downgrade the checks JDK to 17, as it is for some reason always generating (or using cached) classes with this-escape.

I tried different workarounds but will just leave it as a comment for now, as the tests _are_ running on both JDKs 17 and 23.


jlprat (approved):
LGTM

PR #384, opened 2025-08-29
Handle fetch requests with null topic id

giuseppelillo:
Previously every fetch request that did not include the topic id resulted in a UNKNOWN_TOPIC_OR_PARTITION error.
Fix this behaviour by finding batches using the topic name when topic id is not available.

### DIFF
--- a/storage/inkless/src/main/resources/db/migration/V8__Find_batches_handles_null_topic_id.sql
+++ b/storage/inkless/src/main/resources/db/migration/V8__Find_batches_handles_null_topic_id.sql
@@ -1,0 +1,120 @@
+-- Copyright (c) 2025 Aiven, Helsinki, Finland. https://aiven.io/
+
+CREATE DOMAIN topic_id_nullable_t AS UUID;
+
+DROP TYPE find_batches_request_v1 CASCADE;
+
+CREATE TYPE find_batches_request_v1 AS (
+    topic_id topic_id_nullable_t,
+    topic_name topic_name_t,
+    partition partition_t,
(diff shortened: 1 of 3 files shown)


storage/inkless/src/main/resources/db/migration/V8__Find_batches_handles_null_topic_id.sql
```diff
@@ -0,0 +1,120 @@
...
+    partition partition_t,
+    starting_offset BIGINT,
+    max_partition_fetch_bytes INT
+);
+
+DROP TYPE find_batches_response_v1 CASCADE;
+
+CREATE TYPE find_batches_response_v1 AS (
+    topic_id topic_id_t,
+    topic_name topic_name_t,
```
giuseppelillo, line 19: Also return topic_name now, to match the requests with responses when topic id is not available

storage/inkless/src/main/resources/db/migration/V8__Find_batches_handles_null_topic_id.sql
```diff
@@ -0,0 +1,120 @@
...
+        aggregated_batches AS (
+            SELECT
+                idx,
+                array_agg(batch_data ORDER BY base_offset) AS batches
+            FROM final_batch_set
+            GROUP BY idx
+        )
+    SELECT
+        COALESCE(r.topic_id, r.requested_topic_id),
+        COALESCE(r.topic_name, r.requested_topic_name),
```
giuseppelillo, line 111: Return the requested topic id and name when no topic is found

PR #404, opened 2025-09-16
MINOR: Document compression

stanislavkozlovski:
This patch adds a question regarding compression. I frequently confuse the behavior, so I thought to document it.


stanislavkozlovski:
Just to clarify - the compression data is stored in the RecordBatch as per usual.

1. In a normal classic Kafka partition log segment, it is possible to have both compressed and uncompressed data - eg in the case where we have producer A with compression enabled and producer B without compression - right?
2. In a mixed diskless segment, it is also possible to have compressed and uncompressed record batches. The only difference is that they can now belong to different partitions too

PR #407, opened 2025-09-16
feat(inkless): run fetch on fixed thread pools

jeqo:
Define fixed thread-pools for fetch data and metadata to define
an upper-bound to consumption related tasks.

Drop executors where there is no IO.
Replace passing futures to compose CompletableFuture instead.

It updates the exception handling to report metrics properly.

The goal is to allow better allocating resources on fetching side.


ivanyu (changes requested):
The main change LGTM, one small comment

ivanyu (approved):

PR #413, opened 2025-09-26
refactor(inkless): add a batches per partition limit configuration

jeqo:
Introduce another fetch-side configuration to control how many batches to return per partition when serving a single request.

This configuration is pass down to the Control-plane back-end implementation.
PostgreSQL control-plane implementation will use this to limit the number of batches per partition at the DB function level.

Most changes are aligning tests to the new configuration. I've highlighted the main changes to help with the review.

### DIFF
--- a/storage/inkless/src/main/resources/db/migration/V8__Find_batches_function_with_limit.sql
+++ b/storage/inkless/src/main/resources/db/migration/V8__Find_batches_function_with_limit.sql
@@ -1,0 +1,88 @@
+-- Copyright (c) 2025 Aiven, Helsinki, Finland. https://aiven.io/
+CREATE OR REPLACE FUNCTION find_batches_v2(
+    arg_requests find_batches_request_v1[],
+    fetch_max_bytes INT,
+    max_batches_per_partition INT DEFAULT 0
+)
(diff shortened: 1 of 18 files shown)


storage/inkless/src/main/resources/db/migration/V8__Find_batches_function_with_limit.sql
```diff
@@ -0,0 +1,88 @@
...
+                AND b.last_offset >= r.starting_offset
+                AND b.base_offset < r.high_watermark
+        ),
+        per_partition_limited_batches AS (
+            SELECT idx, batch_data, byte_size, base_offset, rn
+            FROM all_batches_with_metadata
+            WHERE (rn = 1  -- each partition gets always at least one batch
+                -- include also last batch, even if it overflows max.partition.fetch.bytes
+                OR (partition_cumulative_bytes - byte_size) < max_partition_fetch_bytes
+                ) AND (max_batches_per_partition = 0 OR rn <= max_batches_per_partition)
```
jeqo, line 58: main change: same as before if new config == 0, otherwise, limit the number of partitions.

PR #453, opened 2025-11-20
refactor(inkless): reuse storage metrics instance on cache

jeqo:
Instead of creating yet another metrics instance, reuse the storage metrics instance used by storage layer, as caching is a related module.

This will change the metric name prefix from:
`io.aiven.inkless.cache.caffeine` to `io.aiven.inkless.storage`

Related fixes:
- Add missing headers to cache classes
- Add tests to ensure metric names do not change without notice


jjaakola-aiven (dismissed):
Naming of the metric is now unclear in the PR. It would be good to have a test to validate that metrics names are expected and not changed by mistake in future as the change will propagate to ops dashboards everywhere.

For example the cache size metrics name would be `io.aiven.inkless.storage.wal-segment-cache.size`, right?

PR #470, opened 2026-01-05
fix(cache): respect max-idle=-1 (disable idle eviction)

AnatolyPopov:
Respect max-idle=-1 (disable idle eviction) setting in CaffeineCache.


jeqo (commented):
Good catch! Could we add a test to validate/document this behavior?

jeqo (approved):

PR #525, opened 2026-03-03
perf(produce): add async commit pipeline to reduce FileCommitWaitTime

Mwea:
## Summary

Replace blocking upload wait with CompletableFuture callbacks. The commit thread no longer blocks on `uploadFuture.get()` - instead, commits chain via `thenCombine()` to preserve submission order while allowing parallel uploads.

**Key changes:**
- FileCommitter now supports async (callback-based) and sync (blocking) modes, controlled by `produce.async.commit.pipeline.enabled` config
- Async mode chains commits via CompletableFuture to preserve order without blocking the single commit thread
- Add error isolation: failed commits don't block subsequent commits (using `handle()` to convert failures for ordering purposes only)
- Add proper shutdown with `awaitTermination()` in `close()`
- Add `markReadyToCommit()` so FileCommitWaitTime measures only the executor queue wait, not the cumulative chain wait

**Tests:**
- Add CacheStoreJobTest for async callback behavior
- Add tests for error isolation in both async and sync modes
- Add tests for awaitTermination shutdown behavior
- Update FileCommitterTest to use real executors with Awaitility

## Benchmark results

| Metric              | Before (sync) | After (async) | Improvement |
|---------------------|---------------|---------------|-------------|
| FileCommitWaitTime p99 | 227ms      | ~8ms          | 96.5%       |

## Test plan

- [x] Unit tests pass for FileCommitterTest
- [x] Unit tests pass for CacheStoreJobTest
- [x] Integration test passes for WriterIntegrationTest
- [x] Manual testing with feature flag enabled in dev-env environment

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/produce/FileCommitter.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/produce/FileCommitter.java
@@ -35,7 +35,6 @@ import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
@@ -98,6 +97,10 @@ class FileCommitter implements Closeable {
 
     private final AtomicInteger totalFilesInProgress = new AtomicInteger(0);
     private final AtomicInteger totalBytesInProgress = new AtomicInteger(0);
+
+    // Tracks the previous commit to ensure commits happen in submission order, not upload completion order.
+    // Each new commit waits for both its upload AND the previous commit to complete before starting.
+    private CompletableFuture<?> previousCommitFuture = CompletableFuture.completedFuture(null);
 
     @DoNotMutate
     FileCommitter(final int brokerId,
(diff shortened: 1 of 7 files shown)


Mwea:
Addressing review comments:

**Re: FileCommitter.java:90 (pipelining futures)**


**Re: InklessConfig.java:135 (feature flag)**

Agreed. Removed the feature flag entirely. Async is now the default and only behavior. The sync code path has been deleted.

storage/inkless/src/main/java/io/aiven/inkless/produce/FileCommitter.java
```diff
@@ -157,83 +162,143 @@ void commit(final ClosedFile file) throws InterruptedException {
...
+        // Chain commit to run after upload completes AND previous commit completes
+        final CompletableFuture<List<CommitBatchResponse>> commitFuture = uploadFuture
+                .thenCombine(prevCommitBarrier, (objectKey, ignored) -> objectKey)
+                .whenComplete((objectKey, error) -> {
+                    // Reset the submit time now that we're ready to commit (success or failure).
+                    // This ensures FileCommitWaitTime measures only the executor queue wait,
+                    // not the time waiting for upload completion or previous commits in the chain.
+                    commitJob.markReadyToCommit();
+                })
+                .handleAsync(commitJob, executorServiceCommit)
```
jeqo: Similarly, using thenAplyAsync can simplify the commitJob.

jeqo (commented):
Changes look good but I think this is mixing a bunch of refactoring and feature work. We can clean this up moving the refactoring to a prep PR that can be quickly merged and do the feature review separately.

PR #562, opened 2026-04-01
refactor(inkless): SharedState to own StorageBackend lifecycle

jeqo:
## Summary

- `SharedState` now creates and manages 4 dedicated `StorageBackend` instances (`fetchStorage`, `maybeLaggingFetchStorage`, `produceStorage`, `backgroundStorage`) instead of exposing a `buildStorage()` factory method
- Centralizes resource lifecycle with proper cleanup on partial initialization failure and ordered shutdown
- Consumers (`Reader`, `Writer`, `FileCleaner`, `FileMerger`) no longer close storage backends — `SharedState` owns their lifecycle

## Changes

- **SharedState**: private constructor, `initialize()` creates backends with try/catch cleanup, `close()` shuts down all backends in reverse creation order using `Utils.closeQuietly`
- **FetchHandler/Reader**: use `fetchStorage()` and `maybeLaggingFetchStorage()` (Optional) from SharedState
- **AppendHandler/Writer**: use `produceStorage()`, Writer no longer stores or closes a StorageBackend field
- **FileCleaner/FileMerger**: use `backgroundStorage()` (shared, thread-safe)
- **New `SharedStateTest`**: verifies close ordering, lagging-disabled path, and failure cleanup

## Motivation

Previously each consumer called `buildStorage()` to create its own backend and was responsible for closing it. This led to:
- Distributed ownership making it hard to reason about lifecycle
- Risk of double-close or missed-close bugs
- No cleanup on partial initialization failure

## Test plan

- [x] All 900 inkless storage tests pass
- [x] New `SharedStateTest` covers close ordering, disabled lagging path, and partial init failure cleanup
- [x] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
@@ -181,7 +181,6 @@ public final class SharedState implements Closeable {
             Utils.closeQuietly(storageMetrics, "storageMetrics");
             Utils.closeQuietly(batchCoordinateCache, "batchCoordinateCache");
             Utils.closeQuietly(objectCache, "objectCache");
-            Utils.closeQuietly(controlPlane, "controlPlane");
             throw new RuntimeException("Failed to initialize SharedState", e);
         }
     }
@@ -195,7 +194,6 @@ public final class SharedState implements Closeable {
         Utils.closeQuietly(storageMetrics, "storageMetrics");
         Utils.closeQuietly(batchCoordinateCache, "batchCoordinateCache");
         Utils.closeQuietly(cache, "objectCache");
-        Utils.closeQuietly(controlPlane, "controlPlane");
     }
 
     public Time time() {
--- a/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
@@ -44,6 +44,7 @@ import io.aiven.inkless.cache.ObjectCache;
 import io.aiven.inkless.config.InklessConfig;
 import io.aiven.inkless.control_plane.ControlPlane;
 import io.aiven.inkless.control_plane.MetadataView;
+import io.aiven.inkless.storage_backend.common.ObjectFetcher;
 import io.aiven.inkless.storage_backend.common.StorageBackend;
 
 public final class SharedState implements Closeable {
@@ -58,7 +59,7 @@ public final class SharedState implements Closeable {
     // Separate storage client for lagging consumers to:
     // 1. Isolate connection pool usage (lagging consumers shouldn't exhaust connections for hot path)
     // 2. Allow independent tuning of timeouts/retries for cold storage access patterns
-    private final Optional<StorageBackend> maybeLaggingFetchStorage;
+    private final Optional<ObjectFetcher> maybeLaggingFetchStorage;
     private final StorageBackend produceStorage;
     // backgroundStorage is used by FileCleaner executor.
     // A dedicated backend instance guarantees it doesn't contend with hot-path fetch/produce clients.
@@ -78,7 +79,7 @@ public final class SharedState implements Closeable {
         final MetadataView metadata,
         final ControlPlane controlPlane,
         final StorageBackend fetchStorage,
-        final Optional<StorageBackend> maybeLaggingFetchStorage,
+        final Optional<ObjectFetcher> maybeLaggingFetchStorage,
         final StorageBackend produceStorage,
         final StorageBackend backgroundStorage,
         final Metrics storageMetrics,
@@ -161,7 +162,7 @@ public final class SharedState implements Closeable {
                 metadata,
                 controlPlane,
                 fetchStorage,
-                Optional.ofNullable(laggingFetchStorage),
+                Optional.<ObjectFetcher>ofNullable(laggingFetchStorage),
                 produceStorage,
                 backgroundStorage,
                 storageMetrics,
@@ -253,7 +254,7 @@ public final class SharedState implements Closeable {
      * <p>When {@code fetch.lagging.consumer.thread.pool.size == 0}, the lagging consumer
      * path is disabled and this storage backend is not created.</p>
      */
-    public Optional<StorageBackend> maybeLaggingFetchStorage() {
+    public Optional<ObjectFetcher> maybeLaggingFetchStorage() {
         return maybeLaggingFetchStorage;
     }
 
--- a/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
@@ -60,9 +60,8 @@ public final class SharedState implements Closeable {
     // 2. Allow independent tuning of timeouts/retries for cold storage access patterns
     private final Optional<StorageBackend> maybeLaggingFetchStorage;
     private final StorageBackend produceStorage;
-    // backgroundStorage is shared by FileCleaner and FileMerger executors which run concurrently.
-    // Kafka storage backends are required to be thread-safe (they share the same Metrics instance).
-    // A dedicated backend instance guarantees they don't contend with hot-path fetch/produce clients.
+    // backgroundStorage is used by FileCleaner executor.
+    // A dedicated backend instance guarantees it doesn't contend with hot-path fetch/produce clients.
     private final StorageBackend backgroundStorage;
     private final ObjectKeyCreator objectKeyCreator;
     private final KeyAlignmentStrategy keyAlignmentStrategy;
--- a/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
@@ -31,6 +31,7 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Supplier;
 
 import io.aiven.inkless.cache.BatchCoordinateCache;
@@ -53,6 +54,16 @@ public final class SharedState implements Closeable {
     private final InklessConfig config;
     private final MetadataView metadata;
     private final ControlPlane controlPlane;
+    private final StorageBackend fetchStorage;
+    // Separate storage client for lagging consumers to:
+    // 1. Isolate connection pool usage (lagging consumers shouldn't exhaust connections for hot path)
+    // 2. Allow independent tuning of timeouts/retries for cold storage access patterns
+    private final Optional<StorageBackend> maybeLaggingFetchStorage;
+    private final StorageBackend produceStorage;
+    // backgroundStorage is shared by FileCleaner and FileMerger executors which run concurrently.
+    // Kafka storage backends are required to be thread-safe (they share the same Metrics instance).
+    // A dedicated backend instance guarantees they don't contend with hot-path fetch/produce clients.
+    private final StorageBackend backgroundStorage;
     private final ObjectKeyCreator objectKeyCreator;
     private final KeyAlignmentStrategy keyAlignmentStrategy;
     private final ObjectCache cache;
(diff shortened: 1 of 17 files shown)


storage/inkless/src/test/java/io/aiven/inkless/common/SharedStateTest.java
```diff
@@ -0,0 +1,199 @@
...
+        inOrder.verify(fourthBackend).close();   // backgroundStorage
+        inOrder.verify(thirdBackend).close();    // produceStorage
+        inOrder.verify(firstBackend).close();    // fetchStorage
+        inOrder.verify(controlPlane).close();
+        // secondBackend was never created, so never closed.
+        verify(secondBackend, never()).close();
+    }
+
+    @Test
+    void shouldCleanupCreatedResourcesInReverseOrderOnInitFailure() throws Exception {
```
EelisK: Nit: we could consider using the `test` prefix for naming these cases, since it's the most dominant style in the codebase 👀

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
```diff
@@ -108,33 +123,79 @@ public static SharedState initialize(
...
+            );
+        } catch (Exception e) {
+            Utils.closeQuietly(backgroundStorage, "backgroundStorage");
+            Utils.closeQuietly(produceStorage, "produceStorage");
+            Utils.closeQuietly(laggingFetchStorage, "laggingFetchStorage");
+            Utils.closeQuietly(fetchStorage, "fetchStorage");
+            Utils.closeQuietly(storageMetrics, "storageMetrics");
+            Utils.closeQuietly(batchCoordinateCache, "batchCoordinateCache");
+            Utils.closeQuietly(objectCache, "objectCache");
+            Utils.closeQuietly(controlPlane, "controlPlane");
```
EelisK: issue: I think there's now a double close of the control plane. The `controlPlane` object is owned by `SharedServer`  and closed in it's `close` method
https://github.com/aiven/inkless/blob/08ea36e537b39e09b44d39bb838e890316f2634d/core/src/main/scala/kafka/server/SharedServer.scala#L390

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
```diff
@@ -108,33 +123,79 @@ public static SharedState initialize(
...
+                backgroundStorage,
+                storageMetrics,
+                objectKeyCreator,
+                keyAlignmentStrategy,
+                objectCache,
+                batchCoordinateCache,
+                brokerTopicStats,
+                defaultTopicConfigs
+            );
+        } catch (Exception e) {
```
EelisK, line 176: This is a pretty broad catch statement 👀 is it possible to limit the scope to only the statements that can realistically raise exceptions?

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
```diff
@@ -108,33 +123,79 @@ public static SharedState initialize(
...
+            storageMetrics = new Metrics(
+                new MetricConfig(), List.of(reporter), Time.SYSTEM,
+                new KafkaMetricsContext(STORAGE_METRIC_CONTEXT)
+            );
+            fetchStorage = config.storage(storageMetrics);
+            // If thread pool size is 0, lagging consumer support is disabled — don't create a separate client.
+            // Enabling lagging consumer support requires a broker restart so that a new storage client can be created.
+            laggingFetchStorage = config.fetchLaggingConsumerThreadPoolSize() > 0 ? config.storage(storageMetrics) : null;
+            produceStorage = config.storage(storageMetrics);
+            backgroundStorage = config.storage(storageMetrics);
```
EelisK, line 155: Note: I think this is now shared between `FileCleaner` and `FileMerger` and could be problematic with multithreaded use, but it should be fine because `FileMerger` is not used and is pending removal?

storage/inkless/src/test/java/io/aiven/inkless/common/SharedStateTest.java
```diff
@@ -0,0 +1,199 @@
...
+                () -> mock(LogConfig.class)
+        )).isInstanceOf(RuntimeException.class)
+                .hasMessageContaining("Failed to initialize SharedState");
+
+        final InOrder inOrder = inOrder(secondBackend, firstBackend, controlPlane, thirdBackend, fourthBackend);
+        inOrder.verify(secondBackend).close();
+        inOrder.verify(firstBackend).close();
+        inOrder.verify(controlPlane).close();
+        inOrder.verify(thirdBackend, times(0)).close();
+        inOrder.verify(fourthBackend, times(0)).close();
```
EelisK: Nit: `never` could be a bit more readable :)

```suggestion
        inOrder.verify(thirdBackend, never()).close();
        inOrder.verify(fourthBackend, never()).close();
```

EelisK (commented):
Nice work, left a few comments and questions 👍 

I think it could be good to have another pair of eyes validate the approach. I fear that `SharedState` could become a god object, which is not good either. Have we considered passing the storage dependencies directly via constructors to each component instead?

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
jeqo: Good catch -- It's currently the case, but agree it should not be needed.

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
jeqo: It's ok to share these resources when related to similar jobs -- in this case background jobs -- as these are thread-safe. Now that they are managed by SharedState, they don't need to race for who needs to close the resource. 
Or do you mean another type of problem could be created here?

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
jeqo: Yeah, it's a bit long. The other option is using Utils.closeAll within (yet another) try-catch but I'm not sure that's much clearer. 
I'd try to move it into a new function but it would need to take parameters as resources are not the same and cannot be hidden.

Using closeAll would be around the same lines:

```java
        } catch (Exception e) {
            try {
                Utils.closeAll(
                    backgroundStorage,
                    produceStorage,
                    laggingFetchStorage,
                    fetchStorage,
                    storageMetrics,
                    batchCoordinateCache,
                    objectCache
                );
            } catch (IOException ex) {
                throw new RuntimeException("Failed to close all resources when initializing SharedState", ex);
            }
            throw new RuntimeException("Failed to initialize SharedState", e);
        }
```

If you feel it's much readable I can to reconsider and use this approach.

jeqo:
@EelisK good catch on the god object. Agree SharedState has high cohesion by design — it's the composition root for the inkless module. Its job is to own the lifecycle of shared infrastructure (storage backends, caches, metrics) and provide them to components.

When upstream this should be broken up and move the initialization into respective managers (e.g. ReplicaManager)

storage/inkless/src/main/java/io/aiven/inkless/consume/Reader.java
```diff
@@ -119,7 +121,7 @@ public Reader(
         BrokerTopicStats brokerTopicStats,
         int fetchMetadataThreadPoolSize,
         int fetchDataThreadPoolSize,
-        ObjectFetcher laggingObjectFetcher,
+        Optional<StorageBackend> maybeLaggingFetchStorage,
```
AnatolyPopov: Nit: Why do we need `StorageBackend` here? It still implements `ObjectFetcher` and I would suspect that we do not need any methods that are not `ObjectFetcher` specific. Or am I missing anything?

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
```diff
@@ -53,6 +54,16 @@ public final class SharedState implements Closeable {
...
     private final ControlPlane controlPlane;
+    private final StorageBackend fetchStorage;
+    // Separate storage client for lagging consumers to:
+    // 1. Isolate connection pool usage (lagging consumers shouldn't exhaust connections for hot path)
+    // 2. Allow independent tuning of timeouts/retries for cold storage access patterns
+    private final Optional<StorageBackend> maybeLaggingFetchStorage;
+    private final StorageBackend produceStorage;
+    // backgroundStorage is shared by FileCleaner and FileMerger executors which run concurrently.
+    // Kafka storage backends are required to be thread-safe (they share the same Metrics instance).
+    // A dedicated backend instance guarantees they don't contend with hot-path fetch/produce clients.
```
AnatolyPopov: I did not manage to find how FIleMerger uses this backgroundStorage, is this comment correct or maybe I'm missing something?

storage/inkless/src/main/java/io/aiven/inkless/produce/Writer.java
```diff
@@ -236,7 +232,7 @@ public void close() throws IOException {
             // Rotate file before closing the uploader so the file gets into the queue first.
             rotateFile(true);
             fileCommitter.close();
-            storage.close();
+            // SharedState owns the storage backend lifecycle.
```
AnatolyPopov: Do we need this comment? Without having the `storage.close()` call here it seems lacking the context.

storage/inkless/src/main/java/io/aiven/inkless/produce/Writer.java
```diff
@@ -79,7 +79,6 @@ class Writer implements Closeable {
 
     private final Lock lock = new ReentrantLock();
     private ActiveFile activeFile;
-    private final StorageBackend storage;
```
AnatolyPopov, line 82: Do I understand correctly that it was just unused here?

storage/inkless/src/main/java/io/aiven/inkless/consume/Reader.java
jeqo: Good catch. I missed we can use `? extends ObjectFetcher` and was using StorageBackend out of convenience. Should be fixed now.

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
jeqo: Yeah, was dropped recently. I've updated the comment

storage/inkless/src/main/java/io/aiven/inkless/produce/Writer.java
jeqo: Pretty much, yes. Probably a left over from initial iterations

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
```diff
@@ -53,6 +54,15 @@ public final class SharedState implements Closeable {
     private final InklessConfig config;
     private final MetadataView metadata;
     private final ControlPlane controlPlane;
+    private final StorageBackend fetchStorage;
+    // Separate storage client for lagging consumers to:
+    // 1. Isolate connection pool usage (lagging consumers shouldn't exhaust connections for hot path)
+    // 2. Allow independent tuning of timeouts/retries for cold storage access patterns
+    private final Optional<StorageBackend> maybeLaggingFetchStorage;
```
AnatolyPopov: Nit: I think this guy can also be ObjecterFetcher now, I do not think there is a need to expose StorageBacknd here since this one is dedicated for fetching.

storage/inkless/src/main/java/io/aiven/inkless/common/SharedState.java
AnatolyPopov: Another issue with this eager initalization we make another issue a bit worse: in BrokerServer SharedState is closed only as part of ReplicaManager shutdown but SharedState is initalized _before_ and the failure can happen in between. In that case SharedState and all the StorageBackends will not be closed.

jeqo: I've changed BrokerServer to close sharedState when replicaManager fails/is empty.

AnatolyPopov (approved):
LGTM now, thanks @jeqo for addressing the comments.

PR #563, opened 2026-04-07
feat(inkless): Init diskless log on Control Plane

giuseppelillo:
Complete the InitDisklessLogManager orchestration with the final transition from `AwaitingMetadata` to `Done`.

### DIFF
--- a/core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
+++ b/core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
@@ -237,8 +237,10 @@ abstract class RetriableInitDisklessLogBatchQueue[S <: InitDisklessLogState](
     }
   }
 
-  private def completeAndRemovePromise(tp: TopicPartition, accepted: Boolean): Unit = {
-    Option(resultPromiseByTp.remove(tp)).foreach(_.trySuccess(accepted))
+  private def completeAndRemovePromise(tp: TopicPartition, accepted: Boolean): Unit = withQueueLock {
+    if (!queuedByTp.containsKey(tp)) {
+      Option(resultPromiseByTp.remove(tp)).foreach(_.trySuccess(accepted))
+    }
   }
 
   private def computeRetryDelayMs(): Long = {
--- a/core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
+++ b/core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
@@ -114,8 +114,11 @@ abstract class RetriableInitDisklessLogBatchQueue[S <: InitDisklessLogState](
   }
 
   def remove(tp: TopicPartition): Unit = {
-    withQueueLock { queuedByTp.remove(tp) }
-    completeAndRemovePromise(tp, accepted = false)
+    val promise = withQueueLock {
+      queuedByTp.remove(tp)
+      Option(resultPromiseByTp.remove(tp))
+    }
+    promise.foreach(_.trySuccess(false))
   }
   
   private def task(): Unit = {
--- a/core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
+++ b/core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
@@ -67,7 +67,7 @@ abstract class RetriableInitDisklessLogBatchQueue[S <: InitDisklessLogState](
 
   private case class Attempt(state: S, attemptNumber: Int)
   private var queuedByTp = new java.util.LinkedHashMap[TopicPartition, Attempt]()
-  private val resultPromiseByTp = new java.util.HashMap[TopicPartition, Promise[Boolean]]()
+  private val resultPromiseByTp = new java.util.concurrent.ConcurrentHashMap[TopicPartition, Promise[Boolean]]()
   private val retryBackoff = new ExponentialBackoff(retryPeriodMs, 2, maxRetryTimeMs, 0.0)
   
   private sealed trait TaskStatus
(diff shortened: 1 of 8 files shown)


core/src/main/scala/kafka/server/InitDisklessLogState.scala
```diff
@@ -197,8 +204,102 @@ object SendingToController {
...
 
-}   
\ No newline at end of file
+  def onSuccess: Done = {
+    info(s"Metadata application successful. Advancing to Done.")
+    Done(partition, topicId)
+  }
+}
+
+/** Terminal state used for observability once metadata init succeeds. */
```
jeqo: nit: as state is reached and immediately deleted; unless is meant to be exposed somehow?
```suggestion
/** Terminal state once metadata init succeeds. */
```

core/src/main/scala/kafka/server/InitDisklessLogState.scala
```diff
@@ -197,8 +204,102 @@ object SendingToController {
...
+    val outcomes = stateSeq.zipWithIndex.map { case (state, index) =>
+      retryOutcomeByIndex.getOrElse(index, {
+        responseResult match {
+          case Left(t) =>
+            // If the single batched call fails, retry all.
+            state.warn(s"Control-plane InitDisklessLog for ${state.tp} failed, scheduling retry", t)
+            retriable(state)
+          case Right(_) =>
+            (if (responseIterator.hasNext) Some(responseIterator.next()) else None) match {
+              case Some(r) if r.error() == Errors.NONE || r.error() == Errors.INVALID_REQUEST =>
```
jeqo, line 291: just to clarify:
```suggestion
              // INVALID_REQUEST = partition already initialized (idempotent success)
              case Some(r) if r.error() == Errors.NONE || r.error() == Errors.INVALID_REQUEST =>
```

core/src/test/scala/unit/kafka/server/InitDisklessLogManagerTest.scala
```diff
@@ -913,6 +917,94 @@ class InitDisklessLogManagerTest {
...
+      partition = partition,
+      topicId = topicId,
+      topicName = tp0.topic(),
+      disklessStartOffset = 100L,
+      producerStates = util.List.of()
+    )
+
+    scheduler.tick()
+
+    verify(controlPlane, atLeastOnce()).initDisklessLog(any())
```
jeqo: I agree with the copilot comment, but instead of testing idempotency, let's rename the test from "IsIdempotent" to "IsSafe". The manager doesn't promise idempotency--it promises safety under repeated calls, with CP as the dedup layer.
```suggestion
    // CP is called twice due to MockScheduler firing delay=0 tasks synchronously —
    // the first completes and removes tp0 before the second re-inserts it.
    // In production the batch would coalesce both into one call.
    // Either way safe: CP treats already-initialized as terminal success (INVALID_REQUEST)
    verify(controlPlane, times(2)).initDisklessLog(any())
    assertTrue(manager.getTrackedPartitions.isEmpty)
```

jeqo (commented):
Looks good to me, only a few minor improvements suggested.

jeqo (dismissed):

core/src/main/scala/kafka/server/InitDisklessLogState.scala
giuseppelillo: Not exposed right, i'll add some metrics later though.

core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
```diff
@@ -97,32 +100,22 @@ abstract class RetriableInitDisklessLogBatchQueue[S <: InitDisklessLogState](
...
-    
     promise.future
   }
 
-  def remove(tp: TopicPartition): Unit = withQueueLock {
-    queuedByTp.remove(tp)
-    Option(resultPromiseByTp.remove(tp)).foreach(_.trySuccess(false))
+  def remove(tp: TopicPartition): Unit = {
+    withQueueLock { queuedByTp.remove(tp) }
+    completeAndRemovePromise(tp, accepted = false)
```
viktorsomogyi: why do you move it to outside the queueLock? It seems like `resultPromiseByTp` in `completeAndRemovePromise` is used without the lock, while in `enqueue()` it's in the lock. That way a promise can be added legitimately but another thread may remove it before it was supposed to be removed. Is this intended?

core/src/test/scala/unit/kafka/server/InitDisklessLogManagerTest.scala
giuseppelillo: I forgot to update the test, so I now realized that AwaitingMetadataBatchQueue was not using lingerMs. I've updated that and the test so now the deduplication logic is actually tested, checking that `controlPlane.initDisklessLog` is only called once.

core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
giuseppelillo: You're right, this was the result of a refactor using a ConcurrentHashMap, but the race condition is still there. I've changed it now so the removal is done within the lock, while the promise completion can be done outside.

core/src/main/scala/kafka/server/InitDisklessLogBatchQueue.scala
```diff
@@ -231,28 +237,15 @@ abstract class RetriableInitDisklessLogBatchQueue[S <: InitDisklessLogState](
     }
   }
 
-  private def completeAndRemovePromise(tp: TopicPartition, accepted: Boolean): Unit = withQueueLock {
+  private def completeAndRemovePromise(tp: TopicPartition, accepted: Boolean): Unit = {
```
viktorsomogyi: I think that this way the following scenario is possible:
1. Thread A runs `task()` and consumes the queue
2. Thread B enqueues the same topic-partition and reuses the same promise
3. Thread A may run into `shouldSend(oldState) == false` and call `completeAndRemovePromise(tp, false)`, which completes that shared promise.

We could introduce tracking the attempt number (if a topic is reenqueued just increase) so we can detect whether the correct promise is removed, but perhaps for now we can just check if we're not removing a promise if the topic is reenqueued:
```suggestion
  private def completeAndRemovePromise(tp: TopicPartition): Unit = withQueueLock {
    if (!queuedByTp.containsKey(tp)) {
      Option(resultPromiseByTp.remove(tp)).foreach(_.trySuccess(false))
    }
  }

giuseppelillo: Nice catch, I've applied the fix you suggested.

jeqo (approved):
LGTM. deferring to @viktorsomogyi for a final look and merge

viktorsomogyi (approved):

PR #567, opened 2026-04-09
feat(inkless): POD-2392 Implement consolidating partition tracking

viktorsomogyi:
This commit implements ConsolidationFetcherManager and ConsolidationFetcherThread in ReplicaManager to track diskless partitions that are being consolidated into classic Kafka logs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -312,6 +312,7 @@ if (repo != null) {
         'inkless-benchmarks/**',
         'tests/kafkatest/tests/inkless/**',
         'inkless-sync/**',
+        'core/src/main/scala/io/aiven/inkless/**',
         // end of inkless ignored files
         'licenses/*',
         '**/generated/**',
--- a/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
@@ -1,0 +1,120 @@
+/*
+ * Inkless
+ * Copyright (C) 2024 - 2026 Aiven OY
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.aiven.inkless.consolidation
+
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.common.message.{FetchResponseData, OffsetForLeaderEpochRequestData, OffsetForLeaderEpochResponseData}
+import org.apache.kafka.common.requests.FetchRequest
+import org.apache.kafka.server.common.OffsetAndEpoch
+import org.apache.kafka.server.network.BrokerEndPoint
+import org.apache.kafka.server.{LeaderEndPoint, PartitionFetchState, ReplicaFetch, ResultWithPartitions}
+
+import java.util
+import java.util.Optional
+
+class DisklessLeaderEndPoint(brokerEndPoint: BrokerEndPoint) extends LeaderEndPoint {
+
+  /**
+   * A boolean specifying if truncation when fetching from the leader is supported
+   */
+  override def isTruncationOnFetchSupported: Boolean = false
+
+  /**
+   * Initiate closing access to fetches from leader.
+   */
+  override def initiateClose(): Unit = {
+  }
+
+  /**
+   * Closes access to fetches from leader.
+   * `initiateClose` must be called prior to invoking `close`.
+   */
+  override def close(): Unit = {
+  }
+
+  /**
+   * The specific broker (host:port) we want to connect to.
+   */
+  override def brokerEndPoint(): BrokerEndPoint = {
+    brokerEndPoint
+  }
+
+  /**
+   * Given a fetchRequest, carries out the expected request and returns
+   * the results from fetching from the leader.
+   *
+   * @param fetchRequest The fetch request we want to carry out
+   * @return A map of topic partition -> fetch data
+   */
+  override def fetch(fetchRequest: FetchRequest.Builder): util.Map[TopicPartition, FetchResponseData.PartitionData] =
+    util.Map.of[TopicPartition, FetchResponseData.PartitionData]
+
+  /**
+   * Fetches the epoch and log start offset of the given topic partition from the leader.
+   *
+   * @param topicPartition     The topic partition that we want to fetch from
+   * @param currentLeaderEpoch An int representing the current leader epoch of the requester
+   * @return An OffsetAndEpoch object representing the earliest offset and epoch in the leader's topic partition.
+   */
+  override def fetchEarliestOffset(topicPartition: TopicPartition, currentLeaderEpoch: Int): OffsetAndEpoch = {
+    return new OffsetAndEpoch(0, 0)
+  }
+
+  /**
+   * Fetches the epoch and log end offset of the given topic partition from the leader.
+   *
+   * @param topicPartition     The topic partition that we want to fetch from
+   * @param currentLeaderEpoch An int representing the current leader epoch of the requester
+   * @return An OffsetAndEpoch object representing the latest offset and epoch in the leader's topic partition.
+   */
+  override def fetchLatestOffset(topicPartition: TopicPartition, currentLeaderEpoch: Int): OffsetAndEpoch = {
+    return new OffsetAndEpoch(0, 0)
+  }
+
+  /**
+   * Fetches offset for leader epoch from the leader for each given topic partition
+   *
+   * @param partitions A map of topic partition -> leader epoch of the replica
+   * @return A map of topic partition -> end offset for a requested leader epoch
+   */
+  override def fetchEpochEndOffsets(partitions: util.Map[TopicPartition, OffsetForLeaderEpochRequestData.OffsetForLeaderPartition]): util.Map[TopicPartition, OffsetForLeaderEpochResponseData.EpochEndOffset] = {
+    util.Map.of[TopicPartition, OffsetForLeaderEpochResponseData.EpochEndOffset]()
+  }
+
+  /**
+   * Fetches the epoch and local log start offset from the leader for the given partition and the current leader-epoch
+   *
+   * @param topicPartition     The topic partition that we want to fetch from
+   * @param currentLeaderEpoch An int representing the current leader epoch of the requester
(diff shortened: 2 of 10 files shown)


build.gradle
```diff
@@ -312,6 +312,7 @@ if (repo != null) {
         'inkless-benchmarks/**',
         'tests/kafkatest/tests/inkless/**',
         'inkless-sync/**',
+        'core/src/main/scala/io/aiven/inkless/**',
```
giuseppelillo, line 315: Why is this needed?

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Given a fetchRequest, carries out the expected request and returns
+   * the results from fetching from the leader.
+   *
+   * @param fetchRequest The fetch request we want to carry out
+   * @return A map of topic partition -> fetch data
+   */
+  override def fetch(fetchRequest: FetchRequest.Builder): util.Map[TopicPartition, FetchResponseData.PartitionData] =
+    util.Map.of[TopicPartition, FetchResponseData.PartitionData]
```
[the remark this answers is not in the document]
viktorsomogyi: Dismissed, DisklessLeaderEndPoint is implemented in #568 , hence these changes here are irrelevant.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Fetches the epoch and log start offset of the given topic partition from the leader.
+   *
+   * @param topicPartition     The topic partition that we want to fetch from
+   * @param currentLeaderEpoch An int representing the current leader epoch of the requester
+   * @return An OffsetAndEpoch object representing the earliest offset and epoch in the leader's topic partition.
+   */
+  override def fetchEarliestOffset(topicPartition: TopicPartition, currentLeaderEpoch: Int): OffsetAndEpoch = {
+    return new OffsetAndEpoch(0, 0)
+  }
```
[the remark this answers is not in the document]
viktorsomogyi: Dismissed, DisklessLeaderEndPoint is implemented in #568 , hence these changes here are irrelevant.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * Builds a fetch request, given a partition map.
+   *
+   * @param partitions A map of topic partitions to their respective partition fetch state
+   * @return A ResultWithPartitions, used to create the fetchRequest for fetch.
+   */
+  override def buildFetch(partitions: util.Map[TopicPartition, PartitionFetchState]): ResultWithPartitions[Optional[ReplicaFetch]] = {
+    new ResultWithPartitions[Optional[ReplicaFetch]](Optional.empty(), util.Set.of())
```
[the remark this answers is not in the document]
viktorsomogyi: Dismissed, DisklessLeaderEndPoint is implemented in #568 , hence these changes here are irrelevant.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2824,36 +2834,45 @@ class ReplicaManager(val config: KafkaConfig,
...
-      if (!isDiskless) {
+      if (!isDiskless || isConsolidatingDisklessTopic) {
         getOrCreatePartition(tp, delta, info.topicId).foreach { case (partition, isNew) =>
           try {
             val partitionAssignedDirectoryId = directoryIds.find(_._1.topicPartition() == tp).map(_._2)
             partition.makeLeader(info.partition, isNew, offsetCheckpoints, Some(info.topicId), partitionAssignedDirectoryId)
 
             changedPartitions.add(partition)
+            if (isConsolidatingDisklessTopic) {
+              partition.createLogIfNotExists(isNew = true, isFutureReplica = false, offsetCheckpoints = offsetCheckpoints, topicId = partition.topicId, targetLogDirectoryId = partitionAssignedDirectoryId)
```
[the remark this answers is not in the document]
viktorsomogyi: Valid comment, applying it.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2824,36 +2834,45 @@ class ReplicaManager(val config: KafkaConfig,
...
+              consolidatingDisklessPartitionsToStartFetching.put(tp, partition)
+            }
```
giuseppelillo, line 2864: This could be done in its own `else if` condition to improve readability, even though it requires duplicating the code to get `partitionAssignedDirectoryId`.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2872,6 +2891,37 @@ class ReplicaManager(val config: KafkaConfig,
         }
       }
     }
+
+    if (consolidatingDisklessPartitionsToStartFetching.nonEmpty) {
+      consolidationFetcherManager.foreach(_.removeFetcherForPartitions(consolidatingDisklessPartitionsToStartFetching.keySet))
+      stateChangeLogger.info(s"Stopped consolidating diskless fetchers as part of become-leader for ${consolidatingDisklessPartitionsToStartFetching.size} partitions")
```
giuseppelillo: This always restart the consolidation fetcher thread right? Is it to follow the same behavior of the replica fetcher threads?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -278,6 +279,10 @@ class ReplicaManager(val config: KafkaConfig,
   private val inklessFileCleaner: Option[FileCleaner] = inklessSharedState.map(new FileCleaner(_))
   // FIXME: FileMerger is having issues with hanging queries. Disabling until fixed.
   private val inklessFileMerger: Option[FileMerger] = None // inklessSharedState.map(new FileMerger(_))
+  private val consolidationFetcherManager: Option[ConsolidationFetcherManager] = if (config.disklessRemoteStorageConsolidationEnabled)
+    Some(new ConsolidationFetcherManager(config, this, quotaManagers.follower))
+  else
+    None
```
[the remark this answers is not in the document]
viktorsomogyi: Done.

build.gradle
viktorsomogyi: The new classes (ConsolidatingFetcherThread, ConsolidatedFetcherManager and DisklessLeaderEndPoint) are in a new package in core (io.aiven.inkless.consolidation) and the RAT check runs on them. As I noticed the Inkless license headers aren't added to RAT, just ignored for now, so I followed it.

This though raises the question which I actually wanted to check: are these files in a good place? I didn't want to mix them with upstream classes to make them easier to distinguish, so I used the package mentioned above.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Yes, I didn't want to deviate from the replica fetcher threads.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Yea, it can be separated if you think that's better for readability. I also considered that, however I wanted the least amount of deviation from the classic code, so it seemed safer and easier to hook into this branch.
If we duplicate this code part, it carries the risk that we might miss changes in this branch in case of rebases.

build.gradle
jeqo: Good idea. I think the placement is good to distinguish things on upstream syncs.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2824,36 +2834,44 @@ class ReplicaManager(val config: KafkaConfig,
...
     delta: TopicsDelta,
     offsetCheckpoints: OffsetCheckpoints,
     localLeaders: mutable.Map[TopicPartition, LocalReplicaChanges.PartitionInfo],
     directoryIds: mutable.Map[TopicIdPartition, Uuid]
   ): Unit = {
     stateChangeLogger.info(s"Transitioning ${localLeaders.size} partition(s) to " +
       "local leaders.")
-    replicaFetcherManager.removeFetcherForPartitions(localLeaders.keySet)
+    val classicLocalLeaders = localLeaders.keySet.filter(tp => !_inklessMetadataView.isDisklessTopic(tp.topic))
+    replicaFetcherManager.removeFetcherForPartitions(classicLocalLeaders)
```
jeqo: Was the filtering intentional here? The original code passed all leaders to `removeFetcherForPartitions` -- is there a reason consolidating leaders should keep their classic replica fetcher entry? 

This scenario feasible:
>  e.g. a partition that was a classic follower (tracked by the replica fetcher) transitions to consolidating-diskless leader:
> 1. Broker is follower of tp while topic is classic → replicaFetcherManager tracks tp
> 2. Topic becomes diskless+consolidating (isDisklessTopic → true)
> 3. Broker becomes leader → applyLocalLeadersDelta filters tp out of classicLocalLeaders (because isDisklessTopic), so replicaFetcherManager.removeFetcherForPartitions never sees tp
> 4. Consolidation fetcher starts for tp → both fetchers now track the same partition

removeFetcherForPartitions is a no-op for untracked partitions, so reverting to localLeaders.keySet (original behavior) should be safe.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2872,6 +2890,37 @@ class ReplicaManager(val config: KafkaConfig,
...
+              partition.getLeaderEpoch,
+              initialFetchOffset(log)
+            ))
+          case None =>
+            stateChangeLogger.trace(s"Unable to start fetching $topicPartition with topic ID ${partition.topicId} " +
+              s"from diskless subsystem ${partition.leaderReplicaIdOpt}.")
+        }
+      }
+
+      consolidationFetcherManager.foreach(_.addFetcherForPartitions(consolidatingPartitionAndOffsets))
```
jeqo, line 2921: Same concern as above: removeFetcherForPartitions is idempotent, so filtering by isConsolidatingDisklessTopic could add a race window — if the topic config changes before stopPartitions runs, the partition is missed and the consolidation fetcher is left orphaned. The other two managers on the lines above (replicaFetcherManager, replicaAlterLogDirsManager) both receive unfiltered partitions.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Good point, I'll revert this.

viktorsomogyi: The main idea here was that we should separate diskless and classic partitions although I indeed missed the transitionary partitions. It also raises the question that whether we should force an ISR change when a classic->diskless or diskless->classic transition happens, so they get picked up by the right fetcher. I'll address that problem in a later PR though to limit the scope in this one.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Good catch. I actually moved this below the `replicaFetcherManager` in line 2846.

giuseppelillo (approved):

PR #568, opened 2026-04-10
feat(inkless): POD-2393 Add DisklessLeaderEndPoint to handle fetch requests

viktorsomogyi:
This commit adds the implementation for DisklessLeaderEndPoint. It uses FetchHandler and FetchOffsetHandler to fetch from the diskless coordinator and append to the local log (which is handled by the fetcher thread).

### DIFF
--- a/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
@@ -18,103 +18,231 @@
 
 package io.aiven.inkless.consolidation
 
+import io.aiven.inkless.consume.{FetchHandler, FetchOffsetHandler}
+import kafka.server.{KafkaConfig, ReplicaManager, ReplicaQuota}
+import org.apache.kafka.common.Uuid
+import org.apache.kafka.common.errors.KafkaStorageException
+import org.apache.kafka.common.message.{FetchResponseData, OffsetForLeaderEpochRequestData}
+import org.apache.kafka.common.message.ListOffsetsRequestData.ListOffsetsPartition
+import org.apache.kafka.common.message.OffsetForLeaderEpochResponseData.EpochEndOffset
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.record.FileRecords.TimestampAndOffset
+import org.apache.kafka.common.requests.{FetchRequest, FetchResponse, ListOffsetsRequest, OffsetsForLeaderEpochResponse}
 import org.apache.kafka.common.TopicPartition
-import org.apache.kafka.common.message.{FetchResponseData, OffsetForLeaderEpochRequestData, OffsetForLeaderEpochResponseData}
-import org.apache.kafka.common.requests.FetchRequest
-import org.apache.kafka.server.common.OffsetAndEpoch
+import org.apache.kafka.metadata.LeaderAndIsr
+import org.apache.kafka.server.common.{MetadataVersion, OffsetAndEpoch}
 import org.apache.kafka.server.network.BrokerEndPoint
+import org.apache.kafka.server.storage.log.{FetchIsolation, FetchParams}
 import org.apache.kafka.server.{LeaderEndPoint, PartitionFetchState, ReplicaFetch, ResultWithPartitions}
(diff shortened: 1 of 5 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -279,10 +279,13 @@ class ReplicaManager(val config: KafkaConfig,
   private val inklessFileCleaner: Option[FileCleaner] = inklessSharedState.map(new FileCleaner(_))
   // FIXME: FileMerger is having issues with hanging queries. Disabling until fixed.
   private val inklessFileMerger: Option[FileMerger] = None // inklessSharedState.map(new FileMerger(_))
-  private val consolidationFetcherManager: Option[ConsolidationFetcherManager] = if (config.disklessRemoteStorageConsolidationEnabled)
-    Some(new ConsolidationFetcherManager(config, this, quotaManagers.follower))
-  else
-    None
+  private val consolidationFetcherManager: Option[ConsolidationFetcherManager] =
+    if (config.disklessRemoteStorageConsolidationEnabled)
+      inklessFetchHandler.zip(inklessFetchOffsetHandler).map { case (fetchHandler, fetchOffsetHandler) =>
```
jeqo, line 287: nit: maybe log a warning if any of these is None? (or when initialized?) -- just ensuring we don't silently by-pass this when that's the case.

jeqo (dismissed):
LGTM. just a minor comment, but not a blocker.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Something like this perhaps? I didn't want to be too specific and mention the exact fields.
```
  private val consolidationFetcherManager: Option[ConsolidationFetcherManager] =
    if (config.disklessRemoteStorageConsolidationEnabled) {
      if (inklessFetchHandler.isEmpty || inklessFetchOffsetHandler.isEmpty) {
        logger.warn("Remote storage consolidation is enabled, however Inkless doesn't seem to be configured properly.")
        None
      } else {
        inklessFetchHandler.zip(inklessFetchOffsetHandler).map { case (fetchHandler, fetchOffsetHandler) =>
          new ConsolidationFetcherManager(config, this, quotaManagers.follower, fetchHandler, fetchOffsetHandler)
        }
      }
    } else {
      None
    }

Comment (jeqo) [reply core/src/main/scala/kafka/server/ReplicaManager.scala:287]: Looks good! Ready to merge, but will wait to merge it in PR stack order 👍🏽

File: core/src/test/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPointTest.scala [hunk 2]
```diff
@@ -0,0 +1,483 @@
...
+    val holder = new FileRecordsOrError(
+      Optional.empty(),
+      Optional.of(new TimestampAndOffset(0L, 10L, Optional.of(5)))
+    )
+    when(fetchOffsetHandler.createJob()).thenReturn(job)
+    when(job.mustHandle(topicPartition.topic())).thenReturn(true)
+    when(job.add(eqTo(topicPartition), any())).thenReturn(CompletableFuture.completedFuture(holder))
+    when(job.start()).thenAnswer { _ =>
+      ()
+    }
```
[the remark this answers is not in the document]
viktorsomogyi: Will resolve in next patch.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -18,103 +18,231 @@
...
+              .setErrorCode(Errors.NONE.code)
+              .setLeaderEpoch(leaderEpoch)
+              .setEndOffset(endOffset)
+          }
+        } catch {
+          case t: Throwable =>
+            tp -> new EpochEndOffset()
+              .setPartition(tp.partition)
+              .setErrorCode(Errors.forException(t).code)
+        }
```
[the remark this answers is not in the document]
viktorsomogyi: For now FetchOffsetHandler doesn't support epochs and the current implementation matches what FetchOffsetHandler returns (initial epoch). POD-2419 should be implemented for correct fetch offset handling but that's a separate task.

jeqo (approved):

PR #572, opened 2026-04-20
refactor(inkless): Rename disklessStartOffset to classicToDisklessStartOffset

giuseppelillo:
Make explicit that the disklessStartOffset used in the KRaft metadata is only used for classic to diskless migrations.


jeqo (approved):
LGTM

PR #573, opened 2026-04-20
feat(inkless): Abort transactions on partition sealing

giuseppelillo:
If there are ongoing transactions in partitions being migrated from classic to diskless, add an ABORT transaction record.

### DIFF
--- a/core/src/main/scala/kafka/cluster/Partition.scala
+++ b/core/src/main/scala/kafka/cluster/Partition.scala
@@ -261,6 +261,10 @@ class Partition(val topicPartition: TopicPartition,
 
   def isSealed: Boolean = _sealed
 
+  /**
+   * Seal this Partition so that no more records can be appended locally.
+   * @throws KafkaStorageException If transaction abortion fails due to an I/O error.
+   */
   def seal(): Unit = inWriteLock(leaderIsrUpdateLock) {
     if (!_sealed) {
       val leaderLog = localLogOrException
@@ -276,6 +280,7 @@ class Partition(val topicPartition: TopicPartition,
    * Abort all ongoing transactions by appending ABORT markers directly to the log.
    * Diskless topics do not support transactions, so any in-flight transaction must
    * be resolved before migration proceeds.
+   * @throws KafkaStorageException If transaction abortion fails due to an I/O error.
    */
   private def abortOngoingTransactions(leaderLog: UnifiedLog): Unit = {
     val producerStateManager = leaderLog.producerStateManager()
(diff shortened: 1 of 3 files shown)


core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -263,8 +263,43 @@ class Partition(val topicPartition: TopicPartition,
...
+      val abortMarker = MemoryRecords.withEndTransactionMarker(
+        producerId,
+        producerEpoch,
+        new EndTransactionMarker(ControlRecordType.ABORT, coordinatorEpoch)
+      )
+      // Use TV_0 because its epoch validation (markerEpoch >= currentEpoch) allows
+      // writing the abort marker with the producer's current epoch. Higher transaction
+      // versions require a bumped epoch, which only the transaction coordinator can
+      // perform.
+      leaderLog.appendAsLeader(abortMarker, leaderEpoch, AppendOrigin.COORDINATOR,
```
jeqo, line 303: Before `seal` was only updating flag and printing logs -- but now it can have IO error here. 
Let's update `sealTopicPartitions` to catch errors similar to `applyLocalLeadersDelta`:

```scala
  def sealTopicPartitions(topic: String): Unit = {
      if (initDisklessLogManager.isEmpty) {
        error(s"Cannot seal partitions for topic $topic: InitDisklessLogManager is not enabled.")
        return
      }
      allPartitions.forEach { (tp, hostedPartition) =>
        if (tp.topic() == topic) {
          hostedPartition match {
            case HostedPartition.Online(partition) if partition.isLeader =>
              partition.topicId match {
                case Some(id) =>
                  try {
                    partition.seal()
                    initDisklessLogManager.foreach(_.registerPartition(partition, id))
                  } catch {
                    case e: KafkaStorageException =>
                      stateChangeLogger.error(s"Failed to seal partition ${partition.topicPartition} " +
                        s"for diskless migration due to a storage error ${e.getMessage}")
                      markPartitionOffline(tp)
                  }
                case None =>
                  error(s"Partition ${partition.topicPartition} has no topic ID, skipping seal and migration registration")
              }
            case _ =>
          }
        }
      }
```

so this doesn't block the whole migration process.

jeqo (commented):
Looks good. Just a comment on error handling.

core/src/main/scala/kafka/cluster/Partition.scala
giuseppelillo: Right, pushed a new commit for fixing this.

viktorsomogyi (approved):

PR #574, opened 2026-04-20
fix(inkless): Create Partition object even when diskless is enabled

giuseppelillo:
Create Partition for diskless topics with local classic data. This allows to recover local data after a broker restart.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2949,6 +2949,22 @@ class ReplicaManager(val config: KafkaConfig,
               s"with topic id ${info.topicId} due to a storage error ${e.getMessage}")
             markPartitionOffline(tp)
         }
+      } else if (logManager.getLog(tp).isDefined) {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3001,6 +3017,23 @@ class ReplicaManager(val config: KafkaConfig,
       if (_inklessMetadataView.isDisklessTopic(tp.topic())) {
         // Clean up migration tracking since only the leader drives classic -> diskless migration.
         initDisklessLogManager.foreach(_.removePartition(tp))
+        // Post-restart: diskless topic still has classic data on local disk.
+        // Create the Partition so classic data remains accessible for reads.
+        if (logManager.getLog(tp).isDefined) {
+          getOrCreatePartition(tp, delta, info.topicId).foreach { case (partition, isNew) =>
```
viktorsomogyi, line 3023: In case of consolidating diskless partitions this call will create the log sooner and then seal it. I think for now we should probably protect it with a `!isConsolidatingDisklessTopic` check.

viktorsomogyi (approved):

PR #580, opened 2026-04-23
feat(inkless): Set migrating partitions in KRaft metadata

giuseppelillo:
When service fetch requests, it's necessary to discern between a topic being migrated from classic to diskless but that has not yet committed the disklessStartOffset to KRaft (in this case fetch still need to go through UnifiedLog) and a full diskless topic (fetch goes through diskless read path).
Introduce a new sentinel value (-2) for `PartitionRegistration.classicToDisklessStartOffset` that indicates that the partition is being migrated from classic to diskless.

Before these changes:
1. Classic topic: `diskless.enable=false`, `PartitionRegistration.classicToDisklessStartOffset = -1`
1. Full diskless topic: `diskless.enable=true`, `PartitionRegistration.classicToDisklessStartOffset = -1`
1. Classic topic being migrated to diskless, that has not committed the diskless start offset yet to KRaft: `diskless.enable=true`, `PartitionRegistration.classicToDisklessStartOffset = -1`
1. Classic topic being migrated to diskless, that has committed the diskless start offset to KRaft: `diskless.enable=true`, `PartitionRegistration.classicToDisklessStartOffset = $disklessStartOffset`

*Case 2 and 3 are equal, even though 2 requires reading from UnifiedLog and 3 needs to read from diskless path.*

After the change:
1. Classic topic: `diskless.enable=false`, `PartitionRegistration.classicToDisklessStartOffset = -1`
1. Full diskless topic: `diskless.enable=true`, `PartitionRegistration.classicToDisklessStartOffset = -1`
1. Classic topic being migrated to diskless, that has not committed the diskless start offset yet to KRaft: `diskless.enable=true`, `PartitionRegistration.classicToDisklessStartOffset = -2`
1. Classic topic being migrated to diskless, that has committed the diskless start offset to KRaft: `diskless.enable=true`, `PartitionRegistration.classicToDisklessStartOffset = $disklessStartOffset`

*All cases can be individually distinguished.*

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
@@ -1936,7 +1936,7 @@ public final class QuorumController implements Controller {
                     List<ApiMessageAndVersion> allRecords = BoundedList.newArrayBacked(MAX_RECORDS_PER_USER_OP);
                     allRecords.addAll(result.records());
                     allRecords.addAll(migrationRecords);
-                    return ControllerResult.of(allRecords, result.response());
+                    return ControllerResult.atomicOf(allRecords, result.response());
                 }
                 return result;
             }
--- a/metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
@@ -121,6 +121,7 @@ import org.apache.kafka.server.common.KRaftVersion;
 import org.apache.kafka.server.common.OffsetAndEpoch;
 import org.apache.kafka.server.fault.FaultHandler;
 import org.apache.kafka.server.fault.FaultHandlerException;
+import org.apache.kafka.server.mutable.BoundedList;
 import org.apache.kafka.server.policy.AlterConfigPolicy;
 import org.apache.kafka.server.policy.CreateTopicPolicy;
 import org.apache.kafka.snapshot.SnapshotReader;
@@ -129,7 +130,6 @@ import org.apache.kafka.timeline.SnapshotRegistry;
 
 import org.slf4j.Logger;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.EnumSet;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


metadata/src/test/java/org/apache/kafka/controller/ReplicationControlManagerTest.java
```diff
@@ -6119,6 +6119,80 @@ public void testInitDisklessLogMultipleProducerStates() {
...
+            assertEquals(NONE.code(),
+                result.response().topics().get(0).partitions().get(0).errorCode());
+
+            ctx.replay(result.records());
+            PartitionRegistration updatedPartition = replicationControl.getPartition(topicId, 0);
+            assertEquals(100L, updatedPartition.classicToDisklessStartOffset);
+        }
+
+        @Test
+        public void testInitDisklessLogRejectsAlreadyInitializedPartition() {
```
AnatolyPopov: What is the difference between this test and `testInitDisklessLogAlreadyInitializedPartitionRejected`?

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -2875,6 +2875,47 @@ private void validatePartitionReplicationFactorUnchanged(PartitionRegistration p
         }
     }
 
+    List<ApiMessageAndVersion> markClassicToDisklessMigrationStarted(
```
AnatolyPopov, line 2878: Should we have some tests for this one?

giuseppelillo: Add a couple of tests

metadata/src/test/java/org/apache/kafka/controller/ReplicationControlManagerTest.java
giuseppelillo: Right they are basically duplicates, i removed the new testInitDisklessLogRejectsAlreadyInitializedPartition

metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
```diff
@@ -1929,6 +1930,14 @@ public CompletableFuture<Map<ConfigResource, ApiError>> incrementalAlterConfigs(
             if (validateOnly) {
                 return result.withoutRecords();
             } else {
+                List<ApiMessageAndVersion> migrationRecords =
+                    replicationControl.markClassicToDisklessMigrationStarted(configChanges, result.response());
+                if (!migrationRecords.isEmpty()) {
+                    List<ApiMessageAndVersion> allRecords = BoundedList.newArrayBacked(MAX_RECORDS_PER_USER_OP);
+                    allRecords.addAll(result.records());
+                    allRecords.addAll(migrationRecords);
+                    return ControllerResult.of(allRecords, result.response());
```
jeqo: Seems there is a subtle bug here, causing the JDK17 CI failure while JDK25 passes. 
`configurationControl.incrementalAlterConfigs` uses `ControllerResult.atomicOf` instead of `ControllerResult.of` -- trying to reproduce this locally.

giuseppelillo: Changed to use `ControllerResult.atomicOf`, let's see how the ci goes

AnatolyPopov: Seems all passed now.

AnatolyPopov (approved):
LGTM

jeqo (approved):

PR #582, opened 2026-04-24
feat(inkless:consume): Add request hedging for storage reads

jeqo:
to mitigate tail latency

When a storage fetch (S3/GCS/Azure) hasn't completed within a configurable threshold, a competing hedge request is fired. First to complete wins; the loser is best-effort cancelled. Applies to both hot (cached) and cold (lagging consumer) paths.

Two independent hedge triggers (at most one hedge per primary):
- TTFB: fetch.hedge.ttfb.threshold.ms — fires when first byte hasn't arrived, catching stuck connections early (default 0 = disabled)
- Total time: fetch.hedge.total.time.threshold.ms — fires when the entire fetch is slow (default 0 = disabled)

When both are enabled, total-time must be > TTFB (validated at startup).

Metrics: HedgeRequestRate, HedgeWonRate

### DIFF
--- a/docs/inkless/configs.rst
+++ b/docs/inkless/configs.rst
@@ -157,6 +157,22 @@ Under ``inkless.``
   * Valid Values: [0,...]
   * Importance: low
 
+``fetch.hedge.total.time.threshold.ms``
+  Total time threshold in milliseconds to trigger a hedge request. When a storage fetch has not completed within this threshold, a competing hedge request is submitted. The first request to complete wins; the other continues in the background and its result is ignored. Set to 0 to disable total-time-based hedging. When both hedging thresholds are enabled, this value must be strictly greater than fetch.hedge.ttfb.threshold.ms.
+
+  * Type: long
+  * Default: 0
+  * Valid Values: [0,...]
+  * Importance: low
+
+``fetch.hedge.ttfb.threshold.ms``
+  Time-to-first-byte threshold in milliseconds to trigger a hedge request. When a storage fetch has not received its first byte within this threshold, a competing hedge request is submitted. This catches stuck connections early, before the total-time threshold. Set to 0 to disable TTFB-based hedging. When both hedging thresholds are enabled, fetch.hedge.total.time.threshold.ms must be strictly greater than this value.
+
+  * Type: long
+  * Default: 0
+  * Valid Values: [0,...]
+  * Importance: low
+
 ``fetch.lagging.consumer.thread.pool.size``
   Thread pool size for lagging consumer fetch requests (consumers reading old data). Set to 0 to disable the lagging consumer feature (all requests will use the recent data path). The default value of 16 is designed as approximately half of the default fetch.data.thread.pool.size (32), providing sufficient capacity for typical cold storage access patterns while leaving headroom for the hot path. The queue capacity is automatically set to thread.pool.size * 100, providing burst buffering (e.g., 16 threads = 1600 queue capacity ≈ 8 seconds buffer at 200 req/s). Tune based on lagging consumer SLA and expected load patterns.
 
--- a/docs/inkless/configs.rst
+++ b/docs/inkless/configs.rst
@@ -40,7 +40,7 @@ Under ``inkless.``
   * Importance: high
 
 ``fetch.lagging.consumer.request.rate.limit``
-  Maximum requests per second for lagging consumer data fetches. Set to 0 to disable rate limiting. The upper bound of 10000 req/s is a safety limit to prevent misconfiguration. For high-throughput systems, consider the relationship between this rate limit, thread pool size, and storage backend capacity. At the default rate of 200 req/s with ~50ms per request latency, this allows ~10 concurrent requests.
+  Maximum requests per second for lagging consumer data fetches. Set to 0 to disable rate limiting. The upper bound of 10000 req/s is a safety limit to prevent misconfiguration. For high-throughput systems, consider the relationship between this rate limit, thread pool size, and storage backend capacity. At the default rate of 200 req/s with ~50ms per request latency, this allows ~10 concurrent requests. Note: hedge requests triggered by slow fetches are exempt from this limit. In the worst case, effective storage GET rate can reach up to 2x this value.
 
   * Type: int
   * Default: 200
@@ -158,7 +158,7 @@ Under ``inkless.``
   * Importance: low
 
 ``fetch.hedge.total.time.threshold.ms``
-  Total time threshold in milliseconds to trigger a hedge request. When a storage fetch has not completed within this threshold, a competing hedge request is submitted. The first request to complete wins; the other continues in the background and its result is ignored. Set to 0 to disable total-time-based hedging. When both hedging thresholds are enabled, this value must be strictly greater than fetch.hedge.ttfb.threshold.ms.
+  Total time threshold in milliseconds to trigger a hedge request. When a storage fetch has not completed within this threshold, a competing hedge request is submitted. The first request to complete wins; the other continues in the background and its result is ignored. Set to 0 to disable total-time-based hedging. When both hedging thresholds are enabled, this value must be strictly greater than fetch.hedge.ttfb.threshold.ms. Capacity impact: hedges submit to the same executor as primaries (fetch.data.thread.pool.size for hot path, fetch.lagging.consumer.thread.pool.size for cold path). Normal case: only tail-latency requests (exceeding threshold) trigger hedges — typically <5% of traffic. Worst case: if all in-flight requests exceed the threshold, effective storage GET rate doubles (one primary + one hedge per request), bounded by executor thread pool + queue capacity. Monitor HedgeRequestRate to detect excessive hedging. If hedge rate is too high, increase this threshold.
 
   * Type: long
   * Default: 0
--- a/storage/inkless/src/main/java/io/aiven/inkless/config/InklessConfig.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/config/InklessConfig.java
@@ -179,7 +179,9 @@ public class InklessConfig extends AbstractConfig {
         + "Set to 0 to disable rate limiting. "
         + "The upper bound of 10000 req/s is a safety limit to prevent misconfiguration. For high-throughput systems, "
         + "consider the relationship between this rate limit, thread pool size, and storage backend capacity. "
-        + "At the default rate of 200 req/s with ~50ms per request latency, this allows ~10 concurrent requests.";
+        + "At the default rate of 200 req/s with ~50ms per request latency, this allows ~10 concurrent requests. "
+        + "Note: hedge requests triggered by slow fetches are exempt from this limit. In the worst case, "
+        + "effective storage GET rate can reach up to 2x this value.";
     // Default 200 req/s: Conservative limit based on typical object storage GET request costs and latency.
     // At ~50ms per request, 200 req/s = ~10 concurrent requests, balancing throughput with cost control.
     // Tune based on storage backend capacity and budget constraints.
@@ -199,7 +201,13 @@ public class InklessConfig extends AbstractConfig {
         + "The first request to complete wins; the other continues in the background and its result is ignored. "
         + "Set to 0 to disable total-time-based hedging. "
         + "When both hedging thresholds are enabled, this value must be strictly greater than "
-        + FETCH_HEDGE_TTFB_THRESHOLD_MS_CONFIG + ".";
+        + FETCH_HEDGE_TTFB_THRESHOLD_MS_CONFIG + ". "
+        + "Capacity impact: hedges submit to the same executor as primaries (fetch.data.thread.pool.size for hot path, "
+        + "fetch.lagging.consumer.thread.pool.size for cold path). "
+        + "Normal case: only tail-latency requests (exceeding threshold) trigger hedges — typically <5% of traffic. "
+        + "Worst case: if all in-flight requests exceed the threshold, effective storage GET rate doubles "
+        + "(one primary + one hedge per request), bounded by executor thread pool + queue capacity. "
+        + "Monitor HedgeRequestRate to detect excessive hedging. If hedge rate is too high, increase this threshold.";
     private static final long FETCH_HEDGE_TOTAL_TIME_THRESHOLD_MS_DEFAULT = 0;
 
     public static final String FETCH_FIND_BATCHES_MAX_BATCHES_PER_PARTITION_CONFIG = "fetch.find.batches.max.per.partition";
--- a/storage/inkless/src/main/java/io/aiven/inkless/config/InklessConfig.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/config/InklessConfig.java
@@ -185,6 +185,23 @@ public class InklessConfig extends AbstractConfig {
     // Tune based on storage backend capacity and budget constraints.
     private static final int FETCH_LAGGING_CONSUMER_REQUEST_RATE_LIMIT_DEFAULT = 200;
 
+    public static final String FETCH_HEDGE_TTFB_THRESHOLD_MS_CONFIG = "fetch.hedge.ttfb.threshold.ms";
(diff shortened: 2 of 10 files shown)


storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -229,24 +244,27 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
             metrics.recordRecentDataRequest();
-            return cache.computeIfAbsent(
+            final AtomicBoolean firstByteReceived = new AtomicBoolean(false);
+            final CompletableFuture<FileExtent> primary = cache.computeIfAbsent(
                 request.toCacheKey(),
-                k -> fetchFileExtent(objectFetcher, request),
+                k -> fetchFileExtent(objectFetcher, request, firstByteReceived),
                 fetchDataExecutor
             );
+            return withHedge(primary, objectFetcher, request, fetchDataExecutor, firstByteReceived);
```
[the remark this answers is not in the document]
jeqo: Benign edge case. When `computeIfAbsent` returns an existing in-flight future (concurrent dedup), the load function isn't invoked so `firstByteReceived` stays `false`. Worst case: a TTFB hedge fires on a deduped request. This is acceptable because:

1. The CAS guard (`hedgeFired`) limits it to one hedge
2. The shared future will likely complete before the timer fires (it's already in progress)
3. An extra hedge on a deduped request is just wasted work, not incorrect behavior

Redesigning the cache value wrapper for this narrow edge case adds disproportionate complexity.

storage/inkless/src/main/java/io/aiven/inkless/config/InklessConfig.java
```diff
@@ -189,6 +189,20 @@ public class InklessConfig extends AbstractConfig {
...
+    public static final String FETCH_HEDGE_TTFB_THRESHOLD_MS_DOC = "Time-to-first-byte threshold in milliseconds to trigger a hedge request. "
+        + "When a storage fetch has not received its first byte within this threshold, a competing hedge request is submitted. "
+        + "This catches stuck connections early, before the total-time threshold. "
+        + "Set to 0 to disable TTFB-based hedging.";
+    private static final long FETCH_HEDGE_TTFB_THRESHOLD_MS_DEFAULT = 0;
+
+    public static final String FETCH_HEDGE_TOTAL_TIME_THRESHOLD_MS_CONFIG = "fetch.hedge.total.time.threshold.ms";
+    public static final String FETCH_HEDGE_TOTAL_TIME_THRESHOLD_MS_DOC = "Total time threshold in milliseconds to trigger a hedge request. "
+        + "When a storage fetch has not completed within this threshold, a competing hedge request is submitted. "
+        + "The first request to complete wins; the other is best-effort cancelled. "
```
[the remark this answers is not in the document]
jeqo: Outdated description. Fixing it.

docs/inkless/configs.rst
```diff
@@ -149,6 +149,22 @@ Under ``inkless.``
   * Valid Values: [0,...]
   * Importance: low
 
+``fetch.hedge.total.time.threshold.ms``
+  Total time threshold in milliseconds to trigger a hedge request. When a storage fetch has not completed within this threshold, a competing hedge request is submitted. The first request to complete wins; the other is best-effort cancelled. Set to 0 to disable hedging.
```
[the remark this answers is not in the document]
jeqo: Same

storage/inkless/src/test/java/io/aiven/inkless/consume/FetchPlannerTest.java
```diff
@@ -1631,4 +1644,616 @@ private void assertBatchPlan(Map<TopicIdPartition, FindBatchResponse> coordinate
...
+                return ByteBuffer.wrap(data);
+            });
+
+            // 1 thread + 1-slot queue: primary occupies the thread, NullCache's supplyAsync
+            // fills the queue slot, so the hedge submission is guaranteed to be rejected.
+            final ExecutorService tinyExecutor = new ThreadPoolExecutor(
+                1, 1, 0L, TimeUnit.MILLISECONDS,
+                new ArrayBlockingQueue<>(1),
+                new ThreadPoolExecutor.AbortPolicy()
+            );
```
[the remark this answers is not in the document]
jeqo: Good call, fixing it

storage/inkless/src/test/java/io/aiven/inkless/consume/FetchPlannerTest.java
```diff
@@ -1631,4 +1644,616 @@ private void assertBatchPlan(Map<TopicIdPartition, FindBatchResponse> coordinate
...
+                );
+
+                final List<FetchPlanner.FetchRequestWithFuture> results = planner.get();
+                assertThat(results).hasSize(1);
+
+                // Primary fails fast, result future should complete exceptionally
+                // (primary error propagates before hedge can fire)
+                assertThatThrownBy(() -> results.get(0).future().get(5, TimeUnit.SECONDS))
+                    .isInstanceOf(ExecutionException.class)
+                    .hasCauseInstanceOf(FileFetchException.class);
```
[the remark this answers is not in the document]
jeqo: Good call, fixing it.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -277,6 +296,126 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
+            triggerMetric.run();
+            try {
+                final CompletableFuture<FileExtent> hedge = CompletableFuture.supplyAsync(
+                    () -> fetchFileExtent(fetcher, request), executor
+                );
+                hedge.whenComplete((value, error) -> {
+                    if (error == null && primary.complete(value)) {
+                        metrics.recordHedgeWon();
+                    }
+                });
```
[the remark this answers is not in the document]
jeqo: Adding some comments to clarify this.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -277,6 +301,130 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
+        final ObjectFetcher fetcher,
+        final ObjectFetchRequest request,
+        final ExecutorService executor,
+        final Runnable triggerMetric
+    ) {
+        if (hedgeFired.compareAndSet(false, true)) {
+            metrics.recordHedgeRequest();
+            triggerMetric.run();
+            try {
+                final CompletableFuture<FileExtent> hedge = CompletableFuture.supplyAsync(
```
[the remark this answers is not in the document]
jeqo: ack with comment. no changes.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -277,6 +301,136 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
+
+        try {
+            // TTFB trigger: fire hedge if first byte hasn't arrived within threshold
+            if (hedgeTtfbThresholdMs > 0) {
+                timers.add(hedgeScheduler.schedule(() -> {
+                    if (!primary.isDone() && !firstByteReceived.get()) {
+                        tryFireHedge(hedgeFired, primary, fetcher, request, executor,
+                            metrics::recordHedgeTtfbTriggered);
+                    }
+                }, hedgeTtfbThresholdMs, TimeUnit.MILLISECONDS));
```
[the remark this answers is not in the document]
jeqo: Low prio. Noise.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -277,6 +301,136 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
+        final ObjectFetcher fetcher,
+        final ObjectFetchRequest request,
+        final ExecutorService executor,
+        final AtomicBoolean firstByteReceived
+    ) {
+        if (hedgeScheduler == null || primary.isDone()) {
+            return primary;
+        }
+
+        final AtomicBoolean hedgeFired = new AtomicBoolean(false);
```
AnatolyPopov: So this is created per caller and not per primary future? Or am I missing something?

jeqo: Per-caller, yes. But each caller maps to a single primary future.

AnatolyPopov: So it is possible to spawn multiple hedging futures here? But the thread pool is single threaded right? Should it be one per primary future?

jeqo: Right, not within the same request (this is guarded by the compare-and-set) but it can fire multiple hedge requests for the same primary (one per concurrent caller sharing the cached future). They all resolve correctly because primary.complete(value) is a compare-and-set as well -- first one wins, the rest are no-ops. Edge case, but should be handled gracefully. 

Scheduler is single-threaded yes. It only schedules the timer callbacks. No need to have one per primary future -- the job is done on the data executors, this only triggers tasks.

I'm improving the docs to clarify this a bit better, and adding a test to cover the scenario where multiple hedge requests are fired.

AnatolyPopov: So what will be the boundary for the number of concurrent hedge fetches if there will be multiple consumers or groups trying to fetch exactly the same offsets? Do I understand correctly that every request will spawn an object storage fetch without any particular limit to number of those? Or am I missing anything? Can this lead to to some kind of hedge storm in the worst case?

jeqo: You are right. I was considering this an edge case but given the cost implication that could carry, and the complexity needed for this is not much, it does not justify not having it -- I'm adding a concurrent map for hedge guards to map primary future to hedge fired flag, so the maximum additional object storage fetches for concurrent requests on the same key should bound to one (the primary fetch + at most one hedge, regardless of caller count).

Overall the maximum number of hedge requests will be now bound to one per in-flight primary request.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -265,6 +291,7 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
                         }
                     }
                 });
+                return withHedge(primary, laggingObjectFetcher, request, laggingFetchDataExecutor, firstByteReceived);
```
AnatolyPopov: Am I missing something or we do not apply rate limiting to hedge requests? It seems that when the hedge request is triggered we might already have facing throttling or any other problems with fetching, so bypassing the rate limiting does not sound as a good idea here.

jeqo: Ah good catch! We don't apply rate limit to hedge requests as it should only trigger hedge request for _running_ primary requests. There is a subtle issue where if primary request is rate-limited, it would appear slow, and trigger hedge prematurely (defeating the rate limit purpose).

In the last push hedge timers are now deferred until after primary request has actually started (e.g. after rate limit blocking).

So the flow is now:
```
  submit primary → [rate limit wait] → fetchStarted → [threshold] → hedge fires
                                        ^ timers start here
```

On whether hedges themselves should be rate-limited: intentionally not. 
A rate-limited hedge would be equally slow as the primary -- making it useless. 
Instead, hedges will be bounded by:
- Timers: only measure actual fetch time, not rate-limit wait
- Per-key dedup: at most one hedge per primary
- Executor capacity: full executor → RejectedExecutionException → silently dropped

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -277,6 +318,142 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
+            // Both thresholds are 0 — no hedging
+            return primary;
+        }
+
+        // Per-key dedup: all callers sharing the same primary (via cache dedup) share one guard.
+        // At most one hedge fires per primary, regardless of concurrent caller count.
+        final AtomicBoolean hedgeFired = hedgeGuards.computeIfAbsent(primary, k -> {
+            k.whenComplete((v, e) -> hedgeGuards.remove(k));
+            return new AtomicBoolean(false);
+        });
```
AnatolyPopov, line 364: As far as I understand there is a race here and according to JavaDoc the `computeIfAbsent()` mapping function should not modify the map during computation.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
AnatolyPopov: I agree the new fetchStarted deferral fixes the earlier bug but the remaining issue seems to be different: once a cold primary has passed the limiter and is slow in storage, the hedge is still an additional lagging storage read outside fetch.lagging.consumer.request.rate.limit. Does that mean the configured “maximum requests per second for lagging consumer data fetches” is no longer a maximum? in the worst case it can become roughly primary rate + hedge rate.
IMO If that is intentional, it needs an explicit separate hedge limit or very clear config/docs that lagging-consumer hedges are exempt and can increase storage GET rate beyond the configured limiter.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
jeqo: ConcurrentHashMap.computeIfAbsent prohibits recursive modification (e.g., calling put/remove on the same map synchronously inside the mapping function, since it holds a bin lock). Here the remove is deferred — it runs outside that.

Added a clarifying comment in the latest push.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
jeqo: I've added a clear doc on config to define this behavior. I don't see the need for a separate hedge limit. Worst case it should double the limit, and configs and metrics are in place to cope with this. Also, hedging is disabled by default so requires opt-in from users.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
AnatolyPopov: Does it run outside? I think whenCompleteAsync would run outside but whenComplete can pick the same thread and still run inside when the future is already completed.

jeqo: Ah good catch! switched to `whenCompleteAsync(…, hedgeScheduler)` that should be fine to run this

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -277,6 +318,143 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
+        if (hedgeTtfbThresholdMs <= 0 && hedgeTotalTimeThresholdMs <= 0) {
+            // Both thresholds are 0 — no hedging
+            return primary;
+        }
+
+        // Per-key dedup: all callers sharing the same primary (via cache dedup) share one guard.
+        // At most one hedge fires per primary, regardless of concurrent caller count.
+        // whenCompleteAsync ensures remove never runs inline during computeIfAbsent.
+        final AtomicBoolean hedgeFired = hedgeGuards.computeIfAbsent(primary, k -> {
+            k.whenCompleteAsync((v, e) -> hedgeGuards.remove(k), hedgeScheduler);
```
ivanyu, line 362: Technically entries may get stuck in the map forever if the provided lambda is not executed, but this can happen only if the executor is shut down, which in turns happens only when the broker is shut down, so I guess it's fine.

ivanyu (approved):
LGTM, thank you!

PR #583, opened 2026-04-27
feat(inkless): POD-2394 Implement read path for consolidated logs

viktorsomogyi:
Consolidated read will be done by prioritizing local logs whenever possible. This means that if the log is consolidating and if the fetched offset falls in the local offset space, meaning fetched offset < localLog.logEndOffset, then read from local logs. Otherwise read from diskless.

In case of diskless partitions that aren't consolidaing, read should be decided based on disklessStartOffset. That means that if there is a diskless offset, then read from the diskless offsets, otherwise read from the local log.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -1999,7 +1999,7 @@ class ReplicaManager(val config: KafkaConfig,
     }
 
     if (classicFetchInfos.isEmpty && disklessFetchInfosWithoutTopicId.isEmpty &&
-      invalidDisklessFetchResponses.isEmpty && invalidConsolidatingPartitionFetchResponses.nonEmpty) {
+      immediateFetchResponses.isEmpty && invalidConsolidatingPartitionFetchResponses.nonEmpty) {
       respond(Seq.empty)
       return
     }
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -1912,14 +1912,22 @@ class ReplicaManager(val config: KafkaConfig,
 
     fetchInfos.foreach { case (tp, partitionData) =>
       val fetchInfo = tp -> partitionData
-      if (!_inklessMetadataView.isDisklessTopic(tp.topic())) {
+      val isDiskless = _inklessMetadataView.isDisklessTopic(tp.topic)
+      if (!isDiskless) {
         classicFetchInfos += fetchInfo
       } else {
         val classicToDisklessStartOffset = _inklessMetadataView.getClassicToDisklessStartOffset(tp.topicPartition())
-        val migrationPending =
-          classicToDisklessStartOffset == PartitionRegistration.CLASSIC_TO_DISKLESS_MIGRATION_PENDING
-        val shouldReadFromUnifiedLog = migrationPending || (
-          classicToDisklessStartOffset >= 0 && partitionData.fetchOffset < classicToDisklessStartOffset)
+        var shouldReadFromUnifiedLog = classicToDisklessStartOffset == PartitionRegistration.CLASSIC_TO_DISKLESS_MIGRATION_PENDING
+        val isConsolidatingPartition =
+          _inklessMetadataView.isConsolidatingDisklessTopic(tp.topic) &&
+            config.disklessRemoteStorageConsolidationEnabled
+        if (isConsolidatingPartition) {
+          shouldReadFromUnifiedLog = shouldReadFromUnifiedLog ||
+            getPartitionOrException(tp).log.exists(partitionData.fetchOffset < _.logEndOffset)
+        } else {
+          shouldReadFromUnifiedLog = shouldReadFromUnifiedLog ||
+            (classicToDisklessStartOffset >= 0 && partitionData.fetchOffset < classicToDisklessStartOffset)
+        }
 
         if (params.isFromFollower && !shouldReadFromUnifiedLog && classicToDisklessStartOffset >= 0) {
           // The partition has fully migrated to diskless and the follower is asking for an offset at or beyond it.
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -1909,10 +1909,12 @@ class ReplicaManager(val config: KafkaConfig,
     val disklessFetchInfosWithoutTopicId = new mutable.ArrayBuffer[(TopicIdPartition, PartitionData)]()
     val classicFetchInfos = new mutable.ArrayBuffer[(TopicIdPartition, PartitionData)]()
     val immediateFetchResponses = new mutable.ArrayBuffer[(TopicIdPartition, FetchPartitionData)]()
+    val invalidConsolidatingPartitionFetchResponses = new mutable.ArrayBuffer[(TopicIdPartition, FetchPartitionData)]()
 
     fetchInfos.foreach { case (tp, partitionData) =>
       val fetchInfo = tp -> partitionData
       val isDiskless = _inklessMetadataView.isDisklessTopic(tp.topic)
+      var partitionAlreadyHandled = false
       if (!isDiskless) {
         classicFetchInfos += fetchInfo
       } else {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1912,14 +1897,22 @@ class ReplicaManager(val config: KafkaConfig,
...
-          classicToDisklessStartOffset == PartitionRegistration.CLASSIC_TO_DISKLESS_MIGRATION_PENDING
-        val shouldReadFromUnifiedLog = migrationPending || (
-          classicToDisklessStartOffset >= 0 && partitionData.fetchOffset < classicToDisklessStartOffset)
+        var shouldReadFromUnifiedLog = classicToDisklessStartOffset == PartitionRegistration.CLASSIC_TO_DISKLESS_MIGRATION_PENDING
+        val isConsolidatingPartition =
+          _inklessMetadataView.isConsolidatingDisklessTopic(tp.topic) &&
+            config.disklessRemoteStorageConsolidationEnabled
+        if (isConsolidatingPartition) {
+          shouldReadFromUnifiedLog = shouldReadFromUnifiedLog ||
+            getPartitionOrException(tp).log.exists(partitionData.fetchOffset < _.logEndOffset)
```
jeqo: should we use `getPartitionOrError` to handle this more gracefully if a partition happen to be offline?

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
```diff
@@ -247,6 +254,22 @@ var record = recordFactory.apply(i);
...
+        configs.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        var consumedRecordsCount = 0L;
+        try (var consumer = new KafkaConsumer<>(configs)) {
+            consumer.subscribe(Collections.singletonList(topicName));
+            for  (int i = 0; i < numRecords; i++) {
+                var records = consumer.poll(Duration.ofMillis(1000));
+                consumedRecordsCount += records.count();
+            }
+        }
+        assertEquals(numRecords, consumedRecordsCount);
```
jeqo: polling based on the number of records could lead to flaky tests. Maybe a more robust approach could be: 
```suggestion
        try (var consumer = new KafkaConsumer<>(configs)) {
            consumer.subscribe(Collections.singletonList(topicName));
            while (consumedRecordsCount < numRecords) {
                var records = consumer.poll(Duration.ofMillis(1000));
                consumedRecordsCount += records.count();
                if (/* exceeded deadline */) break;
            }
        }
        assertEquals(numRecords, consumedRecordsCount);
```
or similar using `TestUtils.waitForCondition`

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
```diff
@@ -247,6 +254,22 @@ var record = recordFactory.apply(i);
         log.info("Produced {} records in total size {}", numRecords, totalSize);
     }
 
+    private void consumeAndVerify(Map<String, Object> configs, int numRecords) {
+        configs.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class.getName());
+        configs.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class.getName());
+        configs.put(ConsumerConfig.GROUP_ID_CONFIG, "test-group-id");
+        configs.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        var consumedRecordsCount = 0L;
+        try (var consumer = new KafkaConsumer<>(configs)) {
```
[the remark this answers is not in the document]
viktorsomogyi: Will resolve in the next patch.

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
```diff
@@ -247,6 +254,22 @@ var record = recordFactory.apply(i);
         log.info("Produced {} records in total size {}", numRecords, totalSize);
     }
 
+    private void consumeAndVerify(Map<String, Object> configs, int numRecords) {
+        configs.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class.getName());
+        configs.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class.getName());
+        configs.put(ConsumerConfig.GROUP_ID_CONFIG, "test-group-id");
+        configs.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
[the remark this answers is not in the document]
viktorsomogyi: Will resolve in the next patch.

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
viktorsomogyi: I have refactored this yesterday evening but forgot to push, so it looks a bit different now. Your suggestion can be included though, so I did that. Will be there in the next patch.

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
```diff
@@ -247,6 +254,22 @@ var record = recordFactory.apply(i);
...
+        var consumedRecordsCount = 0L;
+        try (var consumer = new KafkaConsumer<>(configs)) {
+            consumer.subscribe(Collections.singletonList(topicName));
+            for  (int i = 0; i < numRecords; i++) {
+                var records = consumer.poll(Duration.ofMillis(1000));
+                consumedRecordsCount += records.count();
+            }
+        }
```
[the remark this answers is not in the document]
viktorsomogyi: Resolved.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Yes, probably a fallback to diskless would make sense.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1897,14 +1897,22 @@ class ReplicaManager(val config: KafkaConfig,
...
```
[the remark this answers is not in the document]
viktorsomogyi: Same comment as Jorge's, resolving.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: On a second thought I remember now that I couldn't really fall back to diskless because at this point I think there is no knowledge whether that offset is in the tiered space or the local one. I'll see if I can figure out something.

viktorsomogyi: I'll handle this now by using `getPartitionOrError` and returning the error in the fetch responses. We could in theory improve this with POD-1965 to fall back to diskless offsets if the fetchOffset is >= disklessStartOffset.

viktorsomogyi:
@jeqo resolved the comments, every comment is addressed in a separate commit to ease the review.

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
```diff
@@ -230,23 +238,64 @@ public void testNewConsolidatedDisklessTopics() throws Exception {
...
+                records.forEach(r -> {
+                    var tp = new TopicPartition(r.topic(), r.partition());
+                    offsetsByPartition.computeIfAbsent(tp, __ -> new ArrayList<>()).add(r.offset());
+                });
+                consumedCount.addAndGet(records.count());
+                return consumedCount.get() >= expectedTotalRecords;
+            }, 60_000, "Not all records have been consumed");
+            // Monotonicity + no gaps per partition
+            offsetsByPartition.forEach((tp, offsets) -> {
+                offsets.sort(Long::compareTo);
```
jeqo: ordering should be guaranteed by the log, so no need for reordering here -- more if ordering is being tested after. We can just drop this line I think.
```suggestion
```

jeqo (commented):
LGTM, just a minor comment on one of the tests.

core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
viktorsomogyi: Done, removed it in the latest patch.

AnatolyPopov (approved):
LGTM

jeqo (approved):

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: I'll handle this now by using `getPartitionOrError` and returning the error in the fetch responses. We could in theory improve this with POD-1965 to fall back to diskless offsets if the fetchOffset is >= disklessStartOffset.

PR #584, opened 2026-04-27
feat(inkless): Support ListOffsets for all cases of diskless partitions

giuseppelillo:
Introduce DisklessFetchOffsetRouter to dispatch ListOffsets queries between the classic UnifiedLog path and the diskless control plane, covering all four diskless-managed partition states:
  1. Pure diskless: answer from the diskless control plane.
  2. Hybrid (classic data still present alongside diskless): route by requested timestamp, with classic↔diskless fallback on no-match or empty results.
  3. Migration pending: classic path only.
  4. Follower ListOffsets on a migrated partition: classic path only.

### DIFF
--- a/core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
+++ b/core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
@@ -1,0 +1,348 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kafka.server
+
+import io.aiven.inkless.consume.FetchOffsetHandler
+import kafka.server.metadata.InklessMetadataView
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.common.errors.ApiException
+import org.apache.kafka.common.message.ListOffsetsRequestData.ListOffsetsPartition
+import org.apache.kafka.common.message.ListOffsetsResponseData.ListOffsetsPartitionResponse
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.record.{FileRecords, RecordBatch}
+import org.apache.kafka.common.requests.ListOffsetsRequest
+import org.apache.kafka.metadata.PartitionRegistration
+import org.apache.kafka.server.purgatory.{DelayedOperationPurgatory, DelayedRemoteListOffsets, ListOffsetsPartitionStatus, TopicPartitionOperationKey}
+import org.apache.kafka.storage.internals.log.AsyncOffsetReadFutureHolder
+import org.apache.kafka.storage.internals.log.OffsetResultHolder.FileRecordsOrError
+
+import java.util.Optional
+import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.{CompletableFuture, CopyOnWriteArrayList, Future}
+import scala.jdk.OptionConverters.RichOptional
+
+/**
+ * Routes ListOffsets queries for diskless-managed topics among the classic UnifiedLog path,
+ * the diskless control-plane path, and hybrid configurations that can answer from either.
+ *
+ * Distinguishes among:
+ *   1. Pure diskless partition (`classicToDisklessStartOffset <= 0` or managed replicas disabled):
+ *      fetch offset in the diskless path.
+ *   2. Hybrid diskless partition (`classicToDisklessStartOffset > 0` and managed replicas enabled):
+ *      fetch offset in classic and/or diskless path, depending on the requested timestamp.
+ *   3. Partition that is being migrated from classic to diskless
+ *      (`classicToDisklessStartOffset == CLASSIC_TO_DISKLESS_MIGRATION_PENDING` and managed
+ *      replicas enabled): fetch offset only in the classic path.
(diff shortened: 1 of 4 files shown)


core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
```diff
@@ -0,0 +1,348 @@
...
+    // Migrated partitions seal their classic local log: once classicToDisklessStartOffset is
+    // committed the LEO can no longer grow and every ISR replica has the same data on disk.
+    // Any replica can therefore safely answer ListOffsets from its local log, so we let
+    // followers serve the classic-side query as well.
+    val allowFromFollower = isMigratedWithClassicAccess
+    val isFollowerRequest = replicaId >= 0
+
+    def classicLookup(): ListOffsetsPartitionStatus = classicFetchOffset(topicPartition, partition, allowFromFollower)
+    def disklessLookup: Lookup = disklessLookupOnJob(job, topicPartition, partition)
+    def disklessLookupOnNewJob: Lookup = disklessLookupOnJob(inklessFetchOffsetHandler.get.createJob(), topicPartition, partition, startNow = true)
```
jeqo: nit: instead of .get here, pass FetchOffsetHandler as parameter and deal with option type on the caller.

core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
jeqo, line 1: Could we add a test class for the router specifically? can simplify review and maintenance

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
```diff
@@ -7915,79 +8258,113 @@ class ReplicaManagerTest {
...
-
-      verify(jobMock).add(ArgumentMatchers.eq(disklessTopicPartition.topicPartition()), any())
-      verify(replicaManager, never()).fetchOffsetForTimestamp(any(), anyLong(), any(), any(), anyBoolean())
+      // The diskless fallback's hit must be surfaced (offset 200 / timestamp 456), not the
+      // clamped classic remote hit (offset 100 / timestamp 123).
+      assertEquals(200L, partitionResponse.offset())
+      assertEquals(456L, partitionResponse.timestamp())
+      verify(batchJobMock).add(ArgumentMatchers.eq(disklessTopicPartition.topicPartition()), any())
+      verify(batchJobMock).start()
     }
```
jeqo, line 8948: We could add another test for the scenario where fetch offset happens on pure diskless, e.g.:

```scala
@Test
def testFetchOffsetPureDisklessRoutesToDisklessPathWhenNoMigration(): Unit = {
  val jobMock = Mockito.mock(classOf[FetchOffsetHandler.Job])
  when(jobMock.mustHandle(any())).thenReturn(true)
  doNothing().when(jobMock).start()

  val taskFuture = new CompletableFuture[OffsetResultHolder.FileRecordsOrError]()
  when(jobMock.add(any[TopicPartition], any[ListOffsetsPartition])).thenReturn(taskFuture)
  when(jobMock.cancelHandler()).thenReturn(CompletableFuture.completedFuture(null))

  val fetchOffsetHandlerCtorInit: MockedConstruction.MockInitializer[FetchOffsetHandler] = {
    case (handlerMock, _) =>
      when(handlerMock.createJob()).thenReturn(jobMock)
  }
  val fetchOffsetHandlerCtor = mockConstruction(classOf[FetchOffsetHandler], fetchOffsetHandlerCtorInit)

  val replicaManager = try {
    spy(createReplicaManager(List(disklessTopicPartition.topic()), disklessManagedReplicasEnabled = false))
  } finally {
    fetchOffsetHandlerCtor.close()
  }

  // No migration — pure diskless
  when(replicaManager.inklessMetadataView().getClassicToDisklessStartOffset(disklessTopicPartition.topicPartition()))
    .thenReturn(PartitionRegistration.NO_CLASSIC_TO_DISKLESS_START_OFFSET)

  val (responseOpt, _) = fetchOffset(replicaManager, disklessTopicPartition, ListOffsetsRequest.EARLIEST_TIMESTAMP)

  // Verify routed to diskless, not classic
  verify(jobMock).add(eq(disklessTopicPartition.topicPartition()), any[ListOffsetsPartition])
  verify(replicaManager, never()).fetchOffsetForTimestamp(any(), anyLong(), any(), any(), anyBoolean())
}

```

jeqo (commented):
Looks good. A couple of improvement suggestions

core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
```diff
@@ -0,0 +1,348 @@
...
+
+    if (migrationPending && disklessManagedReplicasEnabled) {
```
viktorsomogyi, line 106: Can we have a case where managed replicas is disabled but migration is pending? In this case we won't have a diskless offset, also `isMigratedWithClassicAccess` will be false, so we may choose the case 1 (`disklessLookup`) that may fail because we won't have a diskless offset. Is this possible?

giuseppelillo: This can't happen because classic to diskless migration requires managed replicas to be enabled. I can still see however some edge cases where the migration is started but not finished and then managed replicas are disabled. I can also add another case to cover this

core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
giuseppelillo: Tests added in a separate commit

core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
giuseppelillo: Done

core/src/main/scala/kafka/server/DisklessFetchOffsetRouter.scala
giuseppelillo: Added a test for this: https://github.com/aiven/inkless/pull/584/changes/d9a45bb2cac1f204c67bed9c7c0b4cdf7502bb43 The request will fail because there's no diskless data on the control plane yet. This is really a misconfiguration case though.

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
giuseppelillo: Added a new test: https://github.com/aiven/inkless/pull/584/changes/8d4214f3fd4fb10166999e7861070a553fa909cc

viktorsomogyi (approved):

PR #587, opened 2026-05-06
feat(inkless): POD-2395 Prune consolidated diskless offsets

viktorsomogyi:
Diskless logs which have been already consolidated to the remote
tier should be removed from the coordinator and the WAL. This commit
adds the functionality to do that:
  - control plane (in-memory and postgres) implementation
  - postgres routine
  - wiring into ReplicaManager

The pruning will be invoked as a periodic task in ReplicaManager
with a configurable cleanup period.


storage/inkless/src/main/java/io/aiven/inkless/control_plane/PruneDisklessLogsRequest.java
```diff
@@ -0,0 +1,6 @@
+package io.aiven.inkless.control_plane;
```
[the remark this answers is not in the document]
viktorsomogyi: resolved

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/PostgresControlPlane.java
```diff
@@ -17,6 +17,8 @@
  */
 package io.aiven.inkless.control_plane.postgres;
 
+import io.aiven.inkless.control_plane.PruneDisklessLogsRequest;
+import io.aiven.inkless.control_plane.PruneDisklessLogsResponse;
```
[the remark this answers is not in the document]
viktorsomogyi: done

storage/inkless/src/main/java/io/aiven/inkless/control_plane/PruneDisklessLogsResponse.java
```diff
@@ -0,0 +1,8 @@
+package io.aiven.inkless.control_plane;
```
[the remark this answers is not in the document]
viktorsomogyi: done

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/PruneDisklessLogsJob.java
```diff
@@ -0,0 +1,77 @@
+package io.aiven.inkless.control_plane.postgres;
```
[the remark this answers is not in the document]
viktorsomogyi: done

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
```diff
@@ -0,0 +1,65 @@
+package io.aiven.inkless.consolidation
```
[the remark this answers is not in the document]
viktorsomogyi: added it

storage/inkless/src/main/java/io/aiven/inkless/control_plane/InMemoryControlPlane.java
```diff
@@ -733,6 +733,11 @@ public synchronized List<GetProducerStateResponse> getProducerState(final List<G
         return result;
     }
 
+    @Override
+    public List<PruneDisklessLogsResponse> pruneDisklessLogs(List<PruneDisklessLogsRequest> pruneDisklessLogsRequests) {
+        throw new UnsupportedOperationException();
+    }
+
```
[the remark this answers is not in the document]
viktorsomogyi: ignoring this one

storage/inkless/src/main/java/io/aiven/inkless/control_plane/InMemoryControlPlane.java
```diff
@@ -733,6 +734,11 @@ public synchronized List<GetProducerStateResponse> getProducerState(final List<G
         return result;
     }
 
+    @Override
+    public List<PruneDisklessLogsResponse> pruneDisklessLogs(List<PruneDisklessLogsRequest> pruneDisklessLogsRequests) {
+        throw new UnsupportedOperationException();
+    }
```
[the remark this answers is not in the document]
viktorsomogyi: Change added.

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -259,6 +261,10 @@ class Partition(val topicPartition: TopicPartition,
    */
   def isAtMinIsr: Boolean = leaderLogIfLocal.exists { partitionState.isr.size == effectiveMinIsr(_) }
 
+  def setDisklessStartOffset(disklessStartOffset: Long) = this.disklessStartOffset = disklessStartOffset
+
+  def getDisklessStartOffset() = disklessStartOffset
```
jeqo: unused?

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -259,6 +261,10 @@ class Partition(val topicPartition: TopicPartition,
    */
   def isAtMinIsr: Boolean = leaderLogIfLocal.exists { partitionState.isr.size == effectiveMinIsr(_) }
 
+  def setDisklessStartOffset(disklessStartOffset: Long) = this.disklessStartOffset = disklessStartOffset
```
jeqo: Should this check that new start offset is always growing? Is moving back a valid scenario?
Also, maybe rename to "bumpDisklessStartOffset" or similar?

storage/inkless/src/main/resources/db/migration/V12__Prune_diskless_batches.sql
```diff
@@ -0,0 +1,103 @@
...
+        )
+        ORDER BY l.topic_id, l.partition  -- ordering is important to prevent deadlocks
+        FOR UPDATE;
+        FOREACH l_request IN ARRAY arg_requests LOOP
+            SELECT *
+            FROM logs
+            WHERE topic_id = l_request.topic_id
+              AND partition = l_request.partition
+            ORDER BY topic_id, partition
+            FOR UPDATE
```
jeqo, line 50: Are these needed? there is a `FOR UPDATE` before already, and results are filtered so ordering shouldn't be needed.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/PruneDisklessLogsResponse.java
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package io.aiven.inkless.control_plane;
+
+import org.apache.kafka.common.TopicIdPartition;
+
+public record PruneDisklessLogsResponse(
+    TopicIdPartition topicIdPartition,
+    Long disklessLogStartOffset,
```
jeqo: In other responses we tend to use primitive `long` and use -1 for invalid offsets, e.g., `EnforceRetentionResponse`

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
```diff
@@ -0,0 +1,95 @@
...
+            pruneDisklessLogsResponse.error)
+        } else {
+          inklessMetadataView.getTopicName(pruneDisklessLogsResponse.topicIdPartition.topicId).toScala match {
+            case Some(topicName) =>
+              val responseTopicIdPartition = new TopicIdPartition(pruneDisklessLogsResponse.topicIdPartition.topicId,
+                pruneDisklessLogsResponse.topicIdPartition.partition, topicName)
+              replicaManager.getPartitionOrError(responseTopicIdPartition.topicPartition) match {
+                case Right(partition) =>
+                  val o = pruneDisklessLogsResponse.disklessLogStartOffset
+                  if (o == null) {
```
jeqo: Is it possible to have error == NONE and no diskless log start offset?

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
```diff
@@ -0,0 +1,95 @@
...
+        case _ => None
+      }.toSeq.asJava
+    if (!requests.isEmpty) {
+      controlPlane.pruneDisklessLogs(requests).asScala.foreach { pruneDisklessLogsResponse =>
+        if (pruneDisklessLogsResponse.error != PruneDisklessLogsError.NONE) {
+          logger.warn("Prune diskless logs did not apply for {} (control plane reported {})",
+            pruneDisklessLogsResponse.topicIdPartition,
+            pruneDisklessLogsResponse.error)
+        } else {
+          inklessMetadataView.getTopicName(pruneDisklessLogsResponse.topicIdPartition.topicId).toScala match {
```
jeqo: is this lookup really needed?

core/src/main/scala/kafka/cluster/Partition.scala
viktorsomogyi: Yea, added it for possible future use cases but now working on my follow up PR to this, I'm not sure I'll use it. I can remove it for now and add whenever needed.

storage/inkless/src/main/resources/db/migration/V12__Prune_diskless_batches.sql
viktorsomogyi: well, it's there to catch an update that was made after the execution of the first one (if new log rows were added since). But maybe it would be better to create a temporary table from the first one and work on that? That way we may not need this and hopefully it would be cleaner.

core/src/main/scala/kafka/cluster/Partition.scala
viktorsomogyi: On a second thought, maybe I can use this for the offset monotonicity validation.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/PruneDisklessLogsResponse.java
viktorsomogyi: ack

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
viktorsomogyi: Good catch, it should really be a primitive type.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
viktorsomogyi: Yes, the response won't contain the topic name, however we need it for `replicaManager.getPartitionOrError(responseTopicIdPartition.topicPartition)` a bit below, however I might be able to fish it out from `eitherErrorOrLog._2` if you think the metadata query is too expensive. Wdyt?

viktorsomogyi: Update: I might be able to acquire it in the response mapper so ignore my previous comment. Will let you know if that doesn't work, until that I'll resolve this.

core/src/main/scala/kafka/cluster/Partition.scala
viktorsomogyi: I'll add a check.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
```diff
@@ -0,0 +1,88 @@
...
+      controlPlane.pruneDisklessLogs(requests).asScala.foreach { pruneDisklessLogsResponse =>
+        if (pruneDisklessLogsResponse.error != PruneDisklessLogsError.NONE) {
+          logger.warn("Prune diskless logs did not apply for {} (control plane reported {})",
+            pruneDisklessLogsResponse.topicIdPartition,
+            pruneDisklessLogsResponse.error)
+        } else {
+          replicaManager.getPartitionOrError(pruneDisklessLogsResponse.topicIdPartition.topicPartition) match {
+            case Right(partition) =>
+              val newDisklessLogStart = pruneDisklessLogsResponse.disklessLogStartOffset
+              if (partition.getDisklessLogStartOffset() > newDisklessLogStart) {
```
viktorsomogyi: @jeqo put the monotonicity check here because inside `Partition` it would've been weird because that way we wouldn't have used `getDisklessLogStartOffset` so we would have ended up with a setter that could potentially throw an exception (depending on the implementation for invalid new diskless start offset).
I have an upcoming PR where I can possibly further enhance the handling of these kind of issues and also add pre-consolidation truncation (because doing consolidation for a migrated topic might be unsafe), so I might be able to address this more completely in that.

storage/inkless/src/main/resources/db/migration/V12__Prune_diskless_batches.sql
viktorsomogyi: Ended up refactoring into a temp table. I could get rid of the `FOR UPDATE` though 😄 . Unfortunately we still need to iterate through arg_requests to be able to return `unknown_topic_or_partition` errors so I can't really simplify it more.

jeqo: I think I misunderstood the initial implementation and the `FOR UPDATE` wasn't redundant -- it's the lock that prevents concurrent mutations between your read and your UPDATE logs SET log_start_offset. Without it, there's a TOCTOU window. The ORDER BY on the locking query is also needed (deadlock prevention). Could you bring those back and add a comment on those? And given that FOR UPDATE must stay, the temp table becomes optional — a PERFORM ... FOR UPDATE + direct re-read from logs inside the loop achieves the same correctness without DDL overhead. Sorry for the confusion here.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
jeqo: What's the intended consumer of this field beyond the pruner's own monotonicity check? If it's just monotonicity enforcement, this should be:
```
  // Self-enforcing — no external caller needs to know the contract
  def advanceDisklessLogStartOffset(newOffset: Long): Boolean = {
    if (newOffset > disklessLogStartOffset) {
      disklessLogStartOffset = newOffset
      true
    } else {
      false
    }
  }
```
Then the pruner becomes:
```
  if (!partition.advanceDisklessLogStartOffset(newDisklessLogStart)) {
    logger.error("Diskless log start offset is non-monotonic...")
  }
```

If there is a future reader (e.g., metrics, etc.), then a proper getter makes sense but should be named to signal it's a cached/approximate value (e.g., cachedDisklessLogStartOffset()), since the source of truth is the control plane. For fetch routing the control plane already handled this I think.

jeqo: Instead of `advanceDisklessLogStartOffset` maybe something closer to Log's `maybeUpdateHighWatermark`, e.g. `maybeUpdateDisklessLogStartOffset`?

viktorsomogyi: > If there is a future reader (e.g., metrics, etc.), then a proper getter makes sense but should be named to signal it's a cached/approximate value (e.g., cachedDisklessLogStartOffset()), since the source of truth is the control plane. For fetch routing the control plane already handled this I think.

Yea, perviously thought that there will be, but it turns out I won't need to use the cached diskless start offset. I want to add a reconciliation step as a pre-requirement to consolidation to eliminate the possibility of the cases where the classic log LEO isn't truncated yet. For this classicToDiskless and diskless log start offsets might have been useful but at that point there's no cached value yet, so I'd have to issue a query to the control plane. But anyway, this is a future issue and since I won't end up using the disklessLogStartOffset, I agree that your proposal makes more sense.

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -259,6 +261,15 @@ class Partition(val topicPartition: TopicPartition,
    */
   def isAtMinIsr: Boolean = leaderLogIfLocal.exists { partitionState.isr.size == effectiveMinIsr(_) }
 
+  def maybeUpdateDisklessLogStartOffset(newDisklessLogStartOffset: Long): Boolean = {
+    if (newDisklessLogStartOffset > disklessLogStartOffset) {
```
viktorsomogyi:
```suggestion
    if (newDisklessLogStartOffset >= disklessLogStartOffset) {
```

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -259,6 +261,15 @@ class Partition(val topicPartition: TopicPartition,
    */
   def isAtMinIsr: Boolean = leaderLogIfLocal.exists { partitionState.isr.size == effectiveMinIsr(_) }
 
+  def maybeUpdateDisklessLogStartOffset(newDisklessLogStartOffset: Long): Boolean = {
```
jeqo: nit: should this be synchronized? I know it's called from a single thread, but worth checking/considering

storage/inkless/src/main/java/io/aiven/inkless/control_plane/MetadataView.java
```diff
@@ -33,6 +34,8 @@ public interface MetadataView {
 
     Uuid getTopicId(String topicName);
 
+    Optional<String> getTopicName(Uuid topicId);
```
jeqo: only used in tests -- worth keeping?

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
```diff
@@ -0,0 +1,85 @@
...
+        if (pruneDisklessLogsResponse.error != PruneDisklessLogsError.NONE) {
+          logger.warn("Prune diskless logs did not apply for {} (control plane reported {})",
+            pruneDisklessLogsResponse.topicIdPartition,
+            pruneDisklessLogsResponse.error)
+        } else {
+          replicaManager.getPartitionOrError(pruneDisklessLogsResponse.topicIdPartition.topicPartition) match {
+            case Right(partition) =>
+              val newDisklessLogStart = pruneDisklessLogsResponse.disklessLogStartOffset
+              if (!partition.maybeUpdateDisklessLogStartOffset(newDisklessLogStart)) {
+                logger.error("Diskless log start offset is non-monotonic. The old one ({}) is greater than the new ({}).")
```
jeqo: missing args?
```suggestion
                logger.error("Diskless log start offset is non-monotonic for {}. The new value ({}) is not greater than current.",
                  pruneDisklessLogsResponse.topicIdPartition.topicPartition,
                  newDisklessLogStart: java.lang.Long)

```

jeqo (commented):
Few small suggestions -- mostly LGTM

storage/inkless/src/main/java/io/aiven/inkless/control_plane/MetadataView.java
viktorsomogyi: Makes sense, I'll remove it.

core/src/main/scala/kafka/cluster/Partition.scala
viktorsomogyi: given the many other methods use the ISR lock, it might be expected by callers that this does it too, so I think it wouldn't hurt adding it.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidatedDisklessLogPruner.scala
viktorsomogyi: What if I just moved this logging inside Partition? That way we could log both offsets.

viktorsomogyi: Thinking more about it, logging an error here doesn't make sense, we should enforce this sooner in the control plane and log an error there if it doesn't work. So I'll probably refactor this.

viktorsomogyi: Refactored this and ended up moving this method into Partition. Now both postgres and the in-memory implementations are monotonic, so logging here only means that we prevented a stale update going through (so I recuced it to warn too).

viktorsomogyi:
@jeqo I have applied your suggestions + rebased the commit on the latest main. The most recent changes have a few other updates too:
- ensure monotonicity of log start offset in postgres
- update byte_size after pruning
- address logging problem with diskless start offset
- construct pruner only when consolidation is enabled

jeqo (approved):

PR #590, opened 2026-05-11
feat(inkless): add lag metrics for consolidated topic partitions

EelisK:
Introduce per-partition metrics to monitor the consolidation process:
- `inkless.remote.consolidation.lag`: diskless log end offset minus remote log end offset
- `inkless.remote.consolidation.local.lag`: diskless log end offset minus local log end offset
- `inkless.remote.consolidation.deletable.messages`: messages consolidated to tiered storage that can be removed

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -23,7 +23,7 @@ import io.aiven.inkless.control_plane.{BatchInfo, FindBatchRequest, FindBatchRes
 import io.aiven.inkless.delete.{DeleteRecordsInterceptor, FileCleaner, RetentionEnforcer}
 import io.aiven.inkless.merge.FileMerger
 import io.aiven.inkless.produce.AppendHandler
-import io.aiven.inkless.consolidation.ConsolidationFetcherManager
+import io.aiven.inkless.consolidation.{ConsolidationFetcherManager, ConsolidationMetrics}
 import kafka.cluster.Partition
 import kafka.log.LogManager
 import kafka.server.HostedPartition.Online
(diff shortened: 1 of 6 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -487,8 +492,11 @@ class ReplicaManager(val config: KafkaConfig,
     val partitions = partitionsToStop.map(_.topicPartition)
     replicaFetcherManager.removeFetcherForPartitions(partitions)
     replicaAlterLogDirsManager.removeFetcherForPartitions(partitions)
-    consolidationFetcherManager.foreach(_.removeFetcherForPartitions(
-      partitions.filter(p => _inklessMetadataView.isConsolidatingDisklessTopic(p.topic))))
+    val consolidatingPartitionsToStop = partitions.filter(p => _inklessMetadataView.isConsolidatingDisklessTopic(p.topic))
```
viktorsomogyi: Please add the` && config.disklessRemoteStorageConsolidationEnabled` check too to filter out whether the feature has been enabled or not.

viktorsomogyi (commented):
Would you mind creating a few tests on the ConsolidationFetcherThread level?

core/src/main/scala/kafka/server/ReplicaManager.scala
EelisK: Added this outside of the filter scope since it's not a property of partitions:
```scala
    val consolidatingPartitionsToStop = if (config.disklessRemoteStorageConsolidationEnabled)
      partitions.filter(p => _inklessMetadataView.isConsolidatingDisklessTopic(p.topic))
      else Set[TopicPartition]()
```

EelisK:
@viktorsomogyi Thanks for the review!
> Would you mind creating a few tests on the ConsolidationFetcherThread level?

Added a few tests :+1:

viktorsomogyi (approved):
Thanks for the updates, LGTM. We can merge it after the CI completes green.

PR #591, opened 2026-05-12
fix(inkless): POD-1965 Use local log start in DisklessLeaderEndPoint

viktorsomogyi:
Fetching with FetchHandler returns the log_start_offset from the coordinator which is actually the diskless start offset. If we pass this on, then ReplicaManager will believe that log_start_offset is the first offset and therefore enforces retention.
Instead of this behavior we should use the local log start offset to properly respect the offset boundaries and retention.

### DIFF
--- a/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
@@ -119,7 +119,9 @@ class DisklessLeaderEndPoint(
         case Right(partition) =>
           val logStartOffset = Try(partition.localLogOrException).toOption match {
             case Some(localLog) => localLog.logStartOffset
-            case None => UnifiedLog.UNKNOWN_OFFSET
+            case None => 
+                    logger.warn("Local log unavailable for topic-partition {}, returning unknown log start offset", tp.topicPartition)
+                    UnifiedLog.UNKNOWN_OFFSET
           }
           if (fetchResponseData.errorCode == Errors.NONE.code) {
             // in case of an inconsistency log an error, set an unknown offset and also return unknown server error
--- a/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
@@ -124,7 +124,7 @@ class DisklessLeaderEndPoint(
           if (fetchResponseData.errorCode == Errors.NONE.code) {
             // in case of an inconsistency log an error, set an unknown offset and also return unknown server error
             if (logStartOffset > data.highWatermark) {
-              logger.error("Local log start offset ({}) is higher than high watermark ({}) for topic-partition {}",
+              logger.error("Local log start offset ({}) is higher than high watermark ({}) for topic-partition {}, this may indicate a transient inconsistency during migration",
                 logStartOffset, data.highWatermark, partition.topicPartition)
               fetchResponseData.setLogStartOffset(UnifiedLog.UNKNOWN_OFFSET)
               fetchResponseData.setErrorCode(Errors.UNKNOWN_SERVER_ERROR.code)
--- a/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
@@ -20,6 +20,7 @@ package io.aiven.inkless.consolidation
 
 import io.aiven.inkless.consume.{FetchHandler, FetchOffsetHandler}
 import kafka.server.{KafkaConfig, ReplicaManager, ReplicaQuota}
+import kafka.utils.Logging
 import org.apache.kafka.common.Uuid
 import org.apache.kafka.common.errors.KafkaStorageException
 import org.apache.kafka.common.message.{FetchResponseData, OffsetForLeaderEpochRequestData}
@@ -35,12 +36,14 @@ import org.apache.kafka.server.network.BrokerEndPoint
 import org.apache.kafka.server.storage.log.{FetchIsolation, FetchParams}
 import org.apache.kafka.server.{LeaderEndPoint, PartitionFetchState, ReplicaFetch, ResultWithPartitions}
 import org.apache.kafka.storage.internals.log.OffsetResultHolder.FileRecordsOrError
+import org.apache.kafka.storage.internals.log.UnifiedLog
 
 import java.util
 import java.util.Optional
 import java.util.concurrent.CompletableFuture
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
+import scala.util.Try
 
 /**
  * Leader endpoint for consolidation fetching from Inkless (object storage) on this broker.
@@ -58,7 +61,7 @@ class DisklessLeaderEndPoint(
   quota: ReplicaQuota,
   metadataVersionSupplier: () => MetadataVersion,
   brokerEpochSupplier: () => Long
-) extends LeaderEndPoint {
+) extends LeaderEndPoint with Logging {
 
   private val replicaId = brokerConfig.brokerId
   private val maxWait = brokerConfig.replicaFetchWaitMaxMs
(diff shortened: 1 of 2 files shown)


core/src/test/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPointTest.scala
```diff
@@ -113,6 +115,9 @@ class DisklessLeaderEndPointTest {
     val fetchHandler = mock(classOf[FetchHandler])
     val fetchOffsetHandler = mock(classOf[FetchOffsetHandler])
     val replicaManager = mock(classOf[ReplicaManager])
+    val partition = mock(classOf[Partition])
+    when(partition.logStartOffset).thenReturn(55L)
+    when(replicaManager.getPartitionOrError(topicPartition)).thenReturn(Right(partition))
```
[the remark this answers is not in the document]
viktorsomogyi: Applied the suggestion.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -96,14 +98,35 @@ class DisklessLeaderEndPoint(
...
+          fetchResponseData.setLogStartOffset(UnifiedLog.UNKNOWN_OFFSET)
+        case Right(partition) =>
+          if (fetchResponseData.errorCode == Errors.NONE.code) {
+            // in case of an inconsistency log an error and set an unknown offset
+            if (partition.logStartOffset > data.highWatermark) {
+              logger.error("Local log start offset ({}) is higher than high watermark ({}) for topic-partition {}",
+                partition.logStartOffset, data.highWatermark, partition.topicPartition)
+              fetchResponseData.setLogStartOffset(UnifiedLog.UNKNOWN_OFFSET)
+            } else {
+              fetchResponseData.setLogStartOffset(partition.logStartOffset)
```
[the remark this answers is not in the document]
viktorsomogyi: Applied the comment.

storage/inkless/src/main/resources/db/migration/V1__Create_tables.sql
```diff
@@ -48,7 +48,7 @@ CREATE TABLE logs (
     topic_id topic_id_t,
     partition partition_t,
     topic_name topic_name_t,
-    log_start_offset offset_t,
+    log_start_offset offset_t, -- start of the diskless offset space
```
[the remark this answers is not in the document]
viktorsomogyi: Applied the suggestion.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -96,14 +99,43 @@ class DisklessLeaderEndPoint(
...
+          // the error in the partition, but otherwise don't mask it.
+          if (fetchResponseData.errorCode == Errors.NONE.code) {
+            fetchResponseData.setErrorCode(error.code)
+          }
+          fetchResponseData.setLogStartOffset(UnifiedLog.UNKNOWN_OFFSET)
+        case Right(partition) =>
+          val logStartOffset = Try(partition.localLogOrException).toOption match {
+            case Some(localLog) => localLog.logStartOffset
+            case None => UnifiedLog.UNKNOWN_OFFSET
+          }
```
giuseppelillo, line 125: Do we alway expect to have the local log here? Even at the very first fetch where for example there only are 10 offsets only in diskless format?

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -96,14 +99,43 @@ class DisklessLeaderEndPoint(
...
+            if (logStartOffset > data.highWatermark) {
+              logger.error("Local log start offset ({}) is higher than high watermark ({}) for topic-partition {}",
+                logStartOffset, data.highWatermark, partition.topicPartition)
+              fetchResponseData.setLogStartOffset(UnifiedLog.UNKNOWN_OFFSET)
+              fetchResponseData.setErrorCode(Errors.UNKNOWN_SERVER_ERROR.code)
+            } else {
+              fetchResponseData.setLogStartOffset(logStartOffset)
+            }
+          } else {
+            fetchResponseData.setLogStartOffset(data.logStartOffset)
```
giuseppelillo: Why is this needed?

viktorsomogyi: It's just there to expose the start offset of diskless to provide more info for the caller. Since this is on the error path, the log start won't be advanced. That said I have no preference here so I'm happy to set UNKNOWN_OFFSET if you think that's better.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
viktorsomogyi: Not necessarily and there are multiple reasons why the log can't exist (partition created but no log yet, the fetch arrived sooner or some kind of failure happened and logs aren't available).

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
viktorsomogyi: So in this case since local logs had no exception but only diskless, it might make sense to return the diskless log start offset, although `UNKNOWN_OFFSET` would be the most conservative or local `logStartOffset` would also make sense.

giuseppelillo: I guess nothing is currently reading that offset if the error is not NONE, but I would prefer to play it safe here and still return `UNKNOWN_OFFSET` in this case.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
giuseppelillo: I'm wondering if there exist a case where the log is not available but we still want to return the log start offset from the diskless log. But given that consolidation only works with managed replicas the Partition should always exist, and in a regular case the UnifiedLog should exist too, but it should be empty, so I guess this is fine. WDYT?

viktorsomogyi: Yes, I think we won't need to return the log start offset of the diskless log. As you said, a partition should always exist in the normal case and if it doesn't then that means it's not running the normal case so we should return unknown offset.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -96,14 +99,43 @@ class DisklessLeaderEndPoint(
...
+          // of the unified log. If there is no error coming from diskless, then we can propagate
+          // the error in the partition, but otherwise don't mask it.
+          if (fetchResponseData.errorCode == Errors.NONE.code) {
+            fetchResponseData.setErrorCode(error.code)
+          }
+          fetchResponseData.setLogStartOffset(UnifiedLog.UNKNOWN_OFFSET)
+        case Right(partition) =>
+          val logStartOffset = Try(partition.localLogOrException).toOption match {
+            case Some(localLog) => localLog.logStartOffset
+            case None => UnifiedLog.UNKNOWN_OFFSET
```
jeqo: let's maybe have a log here if not too noisy? 
```suggestion
            case None => 
                    logger.warn("Local log unavailable for topic-partition {}, returning unknown log start offset", tp.topicPartition)
                    UnifiedLog.UNKNOWN_OFFSET
```

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -96,14 +99,43 @@ class DisklessLeaderEndPoint(
...
+          }
+          if (fetchResponseData.errorCode == Errors.NONE.code) {
+            // in case of an inconsistency log an error, set an unknown offset and also return unknown server error
+            if (logStartOffset > data.highWatermark) {
+              logger.error("Local log start offset ({}) is higher than high watermark ({}) for topic-partition {}",
```
jeqo: let's maybe add a hit on when could this happen?
```suggestion
              logger.error("Local log start offset ({}) is higher than high watermark ({}) for topic-partition {}, this may indicate a transient inconsistency during migration",
```

jeqo (dismissed):
LGTM, just a couple of minor suggestions but your call if actually needed.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
viktorsomogyi: Sure, I think this doesn't hurt.

jeqo (approved):
LGTM. Leaving @giuseppelillo to complete its review

giuseppelillo (approved):

PR #596, opened 2026-05-13
feat(inkless): POD-2457 Allow transactional offset commits for Diskless sources

AnatolyPopov:
Permit TxnOffsetCommitRequest to reference Diskless topic-partitions as offset keys, enabling transactional offset commits for consumers reading from Diskless topics.


viktorsomogyi:
Do you have some broader context here? Why can we allow transactional offset commits?

ivanyu (approved):

ivanyu:
@viktorsomogyi it turns out it prevents diskless topics from participating in transactional offset commits, even if there were not writes to them. Particularly, Iceberg connector cannot work on a diskless topic as source (i.e. without writes).

PR #603, opened 2026-05-21
fix(inkless:switch): Init Diskless Log on Control Plane on leader changes

giuseppelillo:
Currently diskless init on Control Plane is only executed by leaders when they see a change in `classicToDisklessStartOffset` from <0 to >=0, which means that the offset was just committed to the metadata log. In case the init on control plane was not successful and a new leader is elected, the new leader will not retry the init in cases where the metadata record that committed `classicToDisklessStartOffset` was already processed previously by the new leader.
Change this behavior so that diskless init on Control Plane is done also when there's a leader change.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2949,23 +2949,31 @@ class ReplicaManager(val config: KafkaConfig,
       }
     }
 
-    initDisklessLogOnControlPlane(delta)
+    initDisklessLogOnControlPlane(delta, localChanges.leaders.asScala)
   }
 
-  private def initDisklessLogOnControlPlane(delta: TopicsDelta): Unit = {
+  private def initDisklessLogOnControlPlane(
+    delta: TopicsDelta,
+    localLeaders: mutable.Map[TopicPartition, LocalReplicaChanges.PartitionInfo]
+  ): Unit = {
     initDisklessLogManager.foreach { manager =>
-      delta.changedTopics().forEach { (topicId, topicDelta) =>
-        val topicName = topicDelta.name()
-        topicDelta.partitionChanges().forEach { (partitionId, partitionRegistration) =>
-          val previousPartition = Option(delta.image().getTopic(topicId)).flatMap { topicImage =>
-            Option(topicImage.partitions().get(partitionId))
+      localLeaders.foreachEntry { (tp, info) =>
+        val partitionRegistration = info.partition
+        if (partitionRegistration.classicToDisklessStartOffset >= 0) {
+          val previousPartition = Option(delta.image().getTopic(info.topicId)).flatMap { topicImage =>
+            Option(topicImage.partitions().get(tp.partition))
           }
-          val shouldInitOnControlPlane = previousPartition.exists { previous =>
-            previous.classicToDisklessStartOffset < 0 &&
-              partitionRegistration.classicToDisklessStartOffset >= 0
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2949,23 +2949,31 @@ class ReplicaManager(val config: KafkaConfig,
...
+          val disklessStartOffsetJustCommitted = previousPartition.exists { previous =>
+            previous.classicToDisklessStartOffset < 0
           }
 
-          val tp = new TopicPartition(topicName, partitionId)
+          val becameLocalLeader = previousPartition.forall(_.leader != config.nodeId) &&
+            partitionRegistration.leader == config.nodeId
+          // Init Diskless Log on Control Plane if:
+          // - classicToDisklessStartOffset was just committed to the metadata log
+          // - this broker just became leader
```
jeqo: nit: clarify comment
```suggestion
          // Init Diskless Log on Control Plane if this broker is leader and either:
          // - classicToDisklessStartOffset was just committed to the metadata log (offset transition)
          // - this broker just became leader (failover with already-committed offset)
```

core/src/test/scala/unit/kafka/server/metadata/DisklessSwitchFlowTest.scala
```diff
@@ -793,6 +793,167 @@ class DisklessSwitchFlowTest {
...
+      val disklessImage = withClusterBrokers(disklessDelta.apply(MetadataProvenance.EMPTY))
+      broker0Ctx.metadataPublisher.onMetadataUpdate(disklessDelta, disklessImage, metadataManifest())
+      broker1Ctx.metadataPublisher.onMetadataUpdate(disklessDelta, disklessImage, metadataManifest())
+
+      broker0Ctx.time.sleep(broker0Ctx.initDisklessLogManager.lingerMs)
+      broker0Ctx.scheduler.tick()
+      assertEquals(1, broker0Ctx.channelManager.requests.size())
+      assertEquals(0, broker1Ctx.channelManager.requests.size())
+      broker0Ctx.channelManager.requests.poll().complete(new org.apache.kafka.common.message.InitDisklessLogResponseData().setTopics(util.List.of(
+        new org.apache.kafka.common.message.InitDisklessLogResponseData.TopicResponse()
```
jeqo: nit: maybe import InitDisklessLogResponseData separately?

core/src/test/scala/unit/kafka/server/metadata/DisklessSwitchFlowTest.scala
```diff
@@ -793,6 +793,167 @@ class DisklessSwitchFlowTest {
...
+
+      ctx.time.sleep(ctx.initDisklessLogManager.lingerMs)
+      ctx.scheduler.tick()
+      verify(ctx.controlPlane, times(1)).initDisklessLog(any())
+
+      val sameLeaderDelta = new MetadataDelta(pcrImage)
+      sameLeaderDelta.replay(new PartitionChangeRecord()
+        .setTopicId(topicId)
+        .setPartitionId(0)
+        .setLeader(ctx.config.brokerId)
```
jeqo: nit: changing the leader to no leader change here will not bump the epoch and just shrink the ISR.
```suggestion
        .setLeader(RecordFieldHelpers.NO_LEADER_CHANGE)
```

jeqo (commented):
LGTM, just a couple of minor suggestions.

jeqo (approved):

PR #604, opened 2026-05-21
refactor(inkless:switch): drop sealing and registering from BrokerMetadataPublisher

giuseppelillo:
The controller already emits the `diskless.enable=true` ConfigRecord and the per-partition `CLASSIC_TO_DISKLESS_SWITCH_PENDING` marker in a single atomic op, so the broker no longer needs a separate config-delta scan to seal existing leaders.
Sealing and InitDisklessLogManager registration are now driven entirely from `applyLocalLeadersDelta`, gated on the switch-pending marker so plain leader changes on already-diskless topics don't re-trigger init.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -33,7 +33,6 @@ import kafka.server.metadata.{InklessMetadataView, KRaftMetadataCache}
 import kafka.server.share.DelayedShareFetch
 import kafka.utils._
 import org.apache.kafka.common.{IsolationLevel, Node, TopicIdPartition, TopicPartition, Uuid}
-import org.apache.kafka.common.config.{ConfigResource, TopicConfig}
 import org.apache.kafka.common.errors._
 import org.apache.kafka.common.internals.{Plugin, Topic}
 import org.apache.kafka.common.message.DeleteRecordsResponseData.DeleteRecordsPartitionResult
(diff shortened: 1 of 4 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -590,54 +589,6 @@ class ReplicaManager(val config: KafkaConfig,
     }
   }
 
-  private def sealTopicPartitions(topic: String): Unit = {
-    if (initDisklessLogManager.isEmpty) {
-      error(s"Cannot seal partitions for topic $topic: InitDisklessLogManager is not enabled.")
-      return
-    }
```
jeqo, line 597: We have lost this check on the `applyLocalLeadersDelta`, should be worth adding a logging to detect misconfiguration.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3052,15 +3003,25 @@ class ReplicaManager(val config: KafkaConfig,
           }
         }
       } else if (existingPartition.isDefined && !isConsolidatingDisklessTopic) {
```
jeqo, line 3005:
```suggestion
      } else if (existingPartition.isDefined && !isConsolidatingDisklessTopic) {
          if (initDisklessLogManager.isEmpty) {
              error(s"Cannot seal partitions for topic $topic: InitDisklessLogManager is not enabled.")
              return
          }
```

jeqo (approved):

PR #612, opened 2026-05-26
feat(inkless): KC-72 Reconcile stale records after diskless switch

viktorsomogyi:
When a partition switches to diskless, a replica can still hold uncommitted records above the high watermark. If the switch is quick and consolidation is enabled, the consolidation fetchers would start appending on top of that stale tail. This commit implements the following:

- Truncate a newly switched partition's local log down to the just-committed seal (classic-to-diskless start offset). This runs after makeLeader/makeFollower and before any fetcher starts, so catch-up and consolidation always initialize from the truncated LEO.
- Add ConsolidationReconciler to decide when a consolidating partition joins the consolidation fetcher, covering both the initial switch and resume after failover, restart, or reassignment. ReplicaManager now delegates consolidation fetcher startup to it.
- Hand off from the classic ReplicaFetcher to consolidation once a follower reaches the seal, via the fetcher's self-eviction path.
- Gate pruning of switched diskless partitions behind a prune floor and skip partitions whose switch is still pending; born-consolidated partitions keep pruning directly.
- Transform the per-partition diskless watermark into a single monotonic safeConsolidationPruningFloor (gate + progress tracker), replacing lastAppliedDisklessLogStartOffset.


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -281,6 +281,10 @@ class ReplicaManager(val config: KafkaConfig,
   private val inklessDeleteRecordsInterceptor: Option[DeleteRecordsInterceptor] = inklessSharedState.map(new DeleteRecordsInterceptor(_))
   private val inklessRetentionEnforcer: Option[RetentionEnforcer] = inklessSharedState.map(new RetentionEnforcer(_))
   private val inklessFileCleaner: Option[FileCleaner] = inklessSharedState.map(new FileCleaner(_))
+  // FIXME: FileMerger is having issues with hanging queries. Disabling until fixed.
+  private val inklessFileMerger: Option[FileMerger] = None // inklessSharedState.map(new FileMerger(_))
```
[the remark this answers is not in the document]
viktorsomogyi: Outdated

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -0,0 +1,115 @@
...
+        Failure(new ReconciliationException(s"Skipping consolidation for $tp because classic-to-diskless migration is still pending"))
+      case classicToDisklessInitOffset if classicToDisklessInitOffset >= 0 =>
+        val topicId = partition.topicId.getOrElse {
+          inklessMetadataView.getTopicId(partition.topic)
+        }
+        val request = new GetLogInfoRequest(topicId, partition.topicPartition.partition)
+        controlPlane.getLogInfo(Collections.singletonList(request)).asScala.headOption match {
+          case Some(response) =>
+            if (response.errors != Errors.NONE) {
+              Failure(response.errors.exception())
```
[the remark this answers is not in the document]
viktorsomogyi: Outdated

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -0,0 +1,115 @@
...
+    }
+    consolidatingPartitionAndOffsets
+  }
+
+  private def reconcileSwitchedConsolidatingDisklessPartition(partition: Partition): Try[Long] = {
+    val tp = partition.topicPartition
+    inklessMetadataView.getClassicToDisklessStartOffset(tp) match {
+      case PartitionRegistration.NO_CLASSIC_TO_DISKLESS_START_OFFSET =>
+        // Born-diskless: no destructive reconciliation.
+        Success(initialFetchOffset(partition.localLogOrException))
```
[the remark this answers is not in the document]
viktorsomogyi: Outdated.

core/src/main/scala/io/aiven/inkless/consolidation/ReconciliationException.scala
```diff
@@ -0,0 +1,5 @@
+package io.aiven.inkless.consolidation
+
+class ReconciliationException(message: String) extends RuntimeException(message) {
+
+}
```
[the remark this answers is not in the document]
viktorsomogyi: Outdated.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -299,6 +303,14 @@ class ReplicaManager(val config: KafkaConfig,
     } else {
       None
     }
+  private val consolidationReconciler: Option[ConsolidationReconciler] =
+    if (config.disklessRemoteStorageConsolidationEnabled) {
+      inklessSharedState.flatMap(st =>
+        consolidationFetcherManager.map(cfm => new ConsolidationReconciler(_inklessMetadataView, st.controlPlane, initialFetchOffset, cfm)))
+    } else {
+      None
+    }
```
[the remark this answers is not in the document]
viktorsomogyi: Outdated.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -0,0 +1,115 @@
...
+      case PartitionRegistration.CLASSIC_TO_DISKLESS_SWITCH_PENDING =>
+        Failure(new ReconciliationException(s"Skipping consolidation for $tp because classic-to-diskless migration is still pending"))
+      case classicToDisklessInitOffset if classicToDisklessInitOffset >= 0 =>
+        val topicId = partition.topicId.getOrElse {
+          inklessMetadataView.getTopicId(partition.topic)
+        }
+        val request = new GetLogInfoRequest(topicId, partition.topicPartition.partition)
+        controlPlane.getLogInfo(Collections.singletonList(request)).asScala.headOption match {
+          case Some(response) =>
```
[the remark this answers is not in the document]
viktorsomogyi: Outdated.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3353,30 +3413,37 @@ class ReplicaManager(val config: KafkaConfig,
...
       // Stopping the fetchers must be done first in order to initialize the fetch
       // position correctly.
-      val (consolidatingDisklessPartitionsToAdd, classicPartitionsToAdd) = partitionsToStartFetching.partition { case (tp, p) =>
+      val (consolidatingDisklessPartitionsToStartFetching, classicPartitionsToStartFetching) = partitionsToStartFetching.partition { case (tp, _) =>
         _inklessMetadataView.isConsolidatingDisklessTopic(tp.topic)
       }
-      replicaFetcherManager.removeFetcherForPartitions(classicPartitionsToAdd.keySet)
-      consolidationFetcherManager.foreach(_.removeFetcherForPartitions(consolidatingDisklessPartitionsToAdd.keySet))
+      replicaFetcherManager.removeFetcherForPartitions(classicPartitionsToStartFetching.keySet)
+      consolidationFetcherManager.foreach(_.removeFetcherForPartitions(consolidatingDisklessPartitionsToStartFetching.keySet))
```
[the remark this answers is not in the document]
viktorsomogyi: Valid comment, added a fix.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -0,0 +1,137 @@
...
+          //
+          // The prune floor is the higher of the seal and the current log start offset:
+          //  - at first switch logStartOffset is still the classic prefix start, so the floor is
+          //    the seal, which blocks pruning the diskless region until consolidation has tiered
+          //    past the boundary;
+          //  - on resume logStartOffset has advanced past the seal as consolidated segments were
+          //    tiered and deleted, so it reflects real pruning progress.
+          val pruneFloor = math.max(seal, log.logStartOffset)
+          partition.maybeAdvanceConsolidationPruneFloor(pruneFloor)
+          ConsolidationStartState.Ready(log.logEndOffset)
```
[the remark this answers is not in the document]
viktorsomogyi: This one is valid, but would require us to track epoch possibly or persist the consolidation frontier. Will defer to a follow-up PR.

viktorsomogyi:
Generated a document with AI that could serve as a companion for the review.

## TL;DR

When a classic topic is switched to a diskless ("consolidating") topic, the leader **seals** its
local log at a committed offset (the *seal*, `classicToDisklessStartOffset`). After that, the
diskless control plane owns the log and consolidation fetchers continue it from the seal onward.

This PR closes three gaps in that handover:

1. **Stale tail records** — a replica may hold uncommitted records *beyond* the committed seal.
   They are now **truncated** before any fetcher starts, so consolidation begins from a clean boundary.
2. **Resuming consolidation correctly** — a new `ConsolidationReconciler` decides where consolidation
   fetching should (re)start across first-switch, restart, failover and reassignment.
3. **Safe pruning** — diskless pruning is gated behind a monotonic **prune floor** so the consolidated
   region is never pruned before tiered (remote) storage has caught up to the seal.

The classic catch-up fetcher now also **hands off** to the consolidation fetcher once it reads past
the seal.

## Background

The seal lifecycle is `-1` (`NO_CLASSIC_TO_DISKLESS_START_OFFSET`, never switched) → `-2`
(`CLASSIC_TO_DISKLESS_SWITCH_PENDING`) → committed (`seal >= 0`). The broker driving the switch
seals at its own `HW == LEO` and is frozen before it can append further.

## Significant changes

### `ConsolidationReconciler.scala` (new) — where consolidation resumes

Decides the start offset and initial prune floor per partition, via a small state machine keyed on
the seal:

| Seal | Decision |
|------|----------|
| `-1` | Born-diskless; `Ready(initialFetchOffset)`, no floor. |
| `-2` (PENDING) | `Retry` — switch not committed yet. |
| `>= 0`, `LEO < seal` | `Retry` — classic prefix not fully replicated locally yet. |
| `>= 0`, `LEO >= seal` | `Ready(LEO)` and set floor to `max(seal, logStartOffset)`. |
| `< -2` | `Failed`. |

The `Ready(LEO)` branch is the core resume logic: starting from the **current LEO** covers first
switch, restart/failover, and reassignment without re-consolidating or skipping local data.
`Retry` states are not self-rescheduling — they rely on a later metadata delta or the catch-up
hand-off. Consolidation fetches use the synthetic endpoint `BrokerEndPoint(-1, "diskless", -1)`.

### `ReplicaManager.scala` — truncation + wiring

- `maybeTruncateNewlySwitchedPartition(...)` runs after `makeLeader`/`makeFollower` and before
  fetchers start. It acts **only when the seal was just committed in this delta**
  (`sealJustCommitted`: previous registration `seal < 0`, new `>= 0`):
  - `LEO > seal` → truncate to seal (drop uncommitted tail);
  - `LEO < seal` and leader → warn only (a leader can't catch up from a peer; defensive);
  - storage error → mark offline.
- Consolidation-fetcher startup is delegated to the new `consolidationReconciler` (replacing the
  old inline peer-`BrokerEndPoint` resolution).
- New `startConsolidationFetchersForCaughtUpClassicPartitions(...)` is the entry point for the
  catch-up hand-off.

### `Partition.scala` — prune floor

`lastAppliedDisklessLogStartOffset: Long` becomes `safeConsolidationPruningFloor: Option[Long]`.
`None` makes "no boundary yet" explicit, so the pruner refuses to prune a switched partition until a
floor is set. The floor is **monotonic** and serves two roles: a *gate* (don't prune until remote
storage reaches it) and a *progress tracker* (advances as pruning proceeds).

### `ConsolidatedDisklessLogPruner.scala` — gated pruning

`PENDING` partitions are skipped. Each request goes through `safePruneOffset`: born-diskless prunes
directly; switched partitions defer to the floor gate (`highestRemoteOffset >= floor`); successful
prunes advance the floor.

### Fetcher hand-off

`ReplicaFetcherThread` gains `shouldEvictFullySwitchedDisklessPartitions` (default `true`); after
evicting a caught-up classic follower it triggers the consolidation fetcher.
`ConsolidationFetcherThread` overrides it to `false` so it keeps fetching the diskless tail instead
of self-evicting at the seal.

## End-to-end (first switch, follower)

1. Controller commits the seal.
2. `makeFollower` → truncation drops any tail past the seal.
3. If `LEO < seal`, a classic catch-up fetcher is scheduled.
4. On reading past the seal the classic fetcher self-evicts and hands off to consolidation.
5. Reconciler sees `LEO >= seal` → `Ready(LEO)` and sets the prune floor.

Restart/failover/reassignment skip steps 2–4: the seal was committed in a prior delta
(`sealJustCommitted == false`, no truncation), and the reconciler resumes from the current LEO.

## Tests

- `ConsolidationReconcilerTest` (new) — every reconcile state (first switch, failover resume,
  consolidated-past-seal-not-tiered, reassignment rehydrate, born-diskless, PENDING/below-seal retry).
- `ConsolidatedDisklessLogPrunerTest` — born-consolidated prunes directly; switched partition omitted
  when floor not reached; PENDING skipped.
- `ReplicaFetcherThreadTest` — hand-off is (not) called on (non-)eviction.
- `ReplicaManagerTest` — diskless endpoint wiring; PENDING→committed truncates (13→10); plus two
  truncation-guard tests added in review: seal already committed in base image is **not**
  re-truncated, and a leader below the committed seal stays online and untruncated.

core/src/main/scala/kafka/server/ReplicaFetcherThread.scala
```diff
@@ -102,6 +102,8 @@ class ReplicaFetcherThread(name: String,
     evictFullySwitchedDisklessPartitions()
   }
 
+  protected def shouldEvictFullySwitchedDisklessPartitions: Boolean = true
```
jeqo, line 112: maybe worth a comment?
```suggestion
  /**
   * Whether the eviction check in processPartitionData should mark fully-switched partitions
   * for removal. The classic ReplicaFetcherThread enables this (so it self-evicts at the seal
   * and hands off to consolidation). The ConsolidationFetcherThread disables it because it
   * intentionally fetches for already-switched partitions.
   */

  protected def shouldEvictFullySwitchedDisklessPartitions: Boolean = true
```

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.logger.StateChangeLogger
+import org.apache.kafka.metadata.PartitionRegistration
+import org.apache.kafka.server.network.BrokerEndPoint
+import org.apache.kafka.storage.internals.log.UnifiedLog
+
+import scala.collection.{Set, mutable}
+import scala.jdk.OptionConverters.RichOptional
+
+private sealed trait ConsolidationStartState
```
jeqo, line 40: Also worth a small doc
```suggestion
/**
 * Outcome of per-partition reconciliation before consolidation can start.
 *  - Ready: the partition is safe to hand to the consolidation fetcher at the given offset.
 *  - Retry: the partition cannot consolidate yet (e.g., pending seal, LEO below seal) — skip
 *    this round; a later metadata delta or classic-fetcher catch-up will re-trigger.
 *  - Failed: an unrecoverable error; mark the partition as failed in the fetcher manager.
 */
private sealed trait ConsolidationStartState
```

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -0,0 +1,132 @@
...
+private class ReconciliationException(message: String) extends RuntimeException(message)
+
+class ConsolidationReconciler(replicaManager: ReplicaManager,
+                              stateChangeLogger: StateChangeLogger,
+                              consolidationMetrics: ConsolidationMetrics,
+                              inklessMetadataView: InklessMetadataView,
+                              initialFetchOffset: UnifiedLog => Long,
+                              consolidationFetcherManager: ConsolidationFetcherManager) {
+
+  def startConsolidationFetchers(consolidatingDisklessPartitionsToStartFetching: mutable.HashMap[TopicPartition, Partition]
```
jeqo: nit: bit less verbose, and maybe describe pre/post conditions as comments?
```suggestion
  def startConsolidationFetchers(consolidatingPartitions: mutable.HashMap[TopicPartition, Partition]
```

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -318,6 +321,16 @@ class ReplicaManager(val config: KafkaConfig,
   this.logIdent = s"[ReplicaManager broker=$localBrokerId] "
   protected val stateChangeLogger = new StateChangeLogger(localBrokerId)
 
+  // Defined after stateChangeLogger so the reconciler is constructed with a non-null logger.
+  private val consolidationReconciler: Option[ConsolidationReconciler] =
+    if (config.disklessRemoteStorageConsolidationEnabled) {
+      consolidationFetcherManager.flatMap(cfm =>
```
jeqo: Should we defensively validate consolidationFetchManager and consolidationMetrics are not none to catch any wiring bug?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3062,6 +3075,85 @@ class ReplicaManager(val config: KafkaConfig,
...
+                                                  newRegistration: PartitionRegistration,
+                                                  delta: TopicsDelta): Unit = {
+    val seal = newRegistration.classicToDisklessStartOffset
+    if (seal < 0) return
+
+    val previousPartition = Option(delta.image().getTopic(topicId)).flatMap { topicImage =>
+      Option(topicImage.partitions().get(tp.partition()))
+    }
+    val sealJustCommitted = previousPartition.exists(_.classicToDisklessStartOffset < 0)
+    if (!sealJustCommitted) return
```
jeqo, line 3104: nit:
```suggestion
    val sealAlreadyCommittedInPreviousImage = previousPartition.exists(_.classicToDisklessStartOffset >= 0)
    if (sealAlreadyCommittedInPreviousImage) return
```

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3062,6 +3075,85 @@ class ReplicaManager(val config: KafkaConfig,
...
+      try {
+        val log = partition.localLogOrException
+        if (log.logEndOffset > seal) {
+          stateChangeLogger.info(s"Truncating switched partition $tp from LEO ${log.logEndOffset} " +
+            s"to classic-to-diskless start offset $seal")
+          partition.truncateTo(seal, isFuture = false)
+        } else if (log.logEndOffset < seal && partition.isLeader) {
+          stateChangeLogger.warn(s"Leader partition $tp has LEO ${log.logEndOffset} below " +
+            s"classic-to-diskless start offset $seal; cannot catch up from another replica")
+        }
```
jeqo, line 3122: this state seems worrying -- should we consider marking the partition offline or do any follow-up?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3062,6 +3075,85 @@ class ReplicaManager(val config: KafkaConfig,
...
+   *  - born-consolidated/born-diskless (seal == -1): no classic prefix to replicate, consolidate
+   *    immediately;
+   *  - switch pending (seal == -2): keep replicating the frozen classic prefix on the classic
+   *    fetcher until the controller commits the seal;
+   *  - switched (seal >= 0): only consolidate once the local LEO has reached the committed seal, so
+   *    that the whole classic prefix has been replicated locally; while below the seal it stays on
+   *    the classic fetcher, which self-evicts and hands off to consolidation at the seal.
+   * Non-consolidating topics are never routed to the consolidation fetcher.
+   */
+  private def isConsolidatingFollowerReadyForConsolidation(tp: TopicPartition, partition: Partition): Boolean = {
```
jeqo: nit:
```suggestion
  private def isReadyForConsolidation(tp: TopicPartition, partition: Partition): Boolean = {
```

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3062,6 +3075,85 @@ class ReplicaManager(val config: KafkaConfig,
...
+    val sealJustCommitted = previousPartition.exists(_.classicToDisklessStartOffset < 0)
+    if (!sealJustCommitted) return
+
+    onlinePartition(tp).foreach { partition =>
+      try {
+        val log = partition.localLogOrException
+        if (log.logEndOffset > seal) {
+          stateChangeLogger.info(s"Truncating switched partition $tp from LEO ${log.logEndOffset} " +
+            s"to classic-to-diskless start offset $seal")
+          partition.truncateTo(seal, isFuture = false)
```
jeqo, line 3115: Not sure if here, but somewhere it has to be documented that `classicToDisklessStartOffset` means _first diskless offset_ and not _last classic offset_, so truncation to this offset is good here.

jeqo (commented):
LGTM overall, solid PR. Just left some minor comments if you agree.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: We had a similar thing when initializing consolidationFetcherManager itself (but then checking the fetch handler) and there we log a warning and return `None`. Do you think both could warrant an exception too instead of the warning log? Feels like a wiring bug shouldn't be passed down with just logging.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Well, we have a documentation in InitDisklessLogRequest.json but I'm not sure if you were thinking something closer 🙂.
I think a short comment can be made here too.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Yea, I can increase this to error and mark it offline.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Coming back to this as I didn't notice the change in the relational operators. We don't need to change them as it represents that the seal is just committed in this image, not in the previous. Hence we need to keep the "<" operator.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Yes, failure should be safer here than just logging.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Right, it is there already -- but shouldn't hurt having it here or maybe in PartitionRegistration

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3062,6 +3077,83 @@ class ReplicaManager(val config: KafkaConfig,
...
+
+    onlinePartition(tp).foreach { partition =>
+      try {
+        val log = partition.localLogOrException
+        if (log.logEndOffset > seal) {
+          stateChangeLogger.info(s"Truncating switched partition $tp from LEO ${log.logEndOffset} " +
+            s"to classic-to-diskless start offset $seal")
+          // Seal is the classicToDisklessStartOffset, which is the first offset after switching.
+          // Since truncateTo removes all log entries <= target offset (seal in this case) the
+          // resulting log will end at seal-1.
```
jeqo: nit: adjusting for clarity
```suggestion
          // Seal is the classicToDisklessStartOffset, the first offset owned by diskless storage.
          // truncateTo(seal) removes all entries with offset >= seal, leaving LEO = seal.
          // The last classic record is at offset seal - 1.
```

jeqo (commented):
Thanks for addressing the suggestions! LGTM, just a minor comment suggestion -- should be good to approve from my side.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3062,6 +3077,83 @@ class ReplicaManager(val config: KafkaConfig,
...
+                                                  newRegistration: PartitionRegistration,
+                                                  delta: TopicsDelta): Unit = {
+    val seal = newRegistration.classicToDisklessStartOffset
+    if (seal < 0) return
+
+    val previousPartition = Option(delta.image().getTopic(topicId)).flatMap { topicImage =>
+      Option(topicImage.partitions().get(tp.partition()))
+    }
+    val sealJustCommitted = previousPartition.exists(_.classicToDisklessStartOffset < 0)
+    if (!sealJustCommitted) return
```
giuseppelillo, line 3104: This is unfortunately not enough to catch 100% of cases. There was a similar check for gating the init diskless log for switching, but then I've removed it for a more safe approach: https://github.com/aiven/inkless/pull/603
This is not exactly the same but it suffers from similar scenario, like this one:
```
Initial state:
  partition is classic
  broker A local log: LEO = 120
  cluster HW / safe seal point =  100
  records [100, 120) are uncommitted stale classic tail on A
Then:
  broker A is down
  controller/leader commits classicToDisklessStartOffset = 100
  metadata now says the partition is switched
Later:
  broker A restarts, on restart, the delta may be effectively: "empty image -> committed switched image"
  it loads metadata image where seal = 100 is already present
  truncation is skipped
```

or this other one:
```
1. Broker receives/applyDelta for PENDING -> seal=100.
2. makeLeader/makeFollower runs.
3. Before maybeTruncateNewlySwitchedPartition actually truncates LEO 120 -> 100, broker crashes.
4. On restart, metadata already contains seal=100.
5. sealJustCommitted is false.
6. truncation is skipped
```

Similar to the other comment by Copilot, we might need to track Leader Epochs in a more precise way and have a safe mechanism for truncation

giuseppelillo: I propose therefore to tackle this into another PR, so I will not block this one.

giuseppelillo (dismissed):
LGTM for now, we need to address a couple of issues about truncation but it's cleaner to do it in another PR.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Yes, I started working on that ([KC-156](https://aiven.atlassian.net/browse/KC-156)). In short, the idea is to record the bumped leader epoch (done in #626 ), store it as the diskless leader epoch and then we could use it for recovery, this way kafka will be able to correctly identify that it should truncate to offset 100.

giuseppelillo (approved):
LGTM

PR #616, opened 2026-05-28
feat(inkless:switch): validate diskless requires remote storage when consolidation enabled

jeqo:
When remote storage consolidation is enabled, reject enabling diskless if remote.storage.enable is explicitly set to false. Allow omission — the controller will auto-persist remote.storage.enable=true.

Also expands isSwitchedFromClassicWithRemoteStorage() to permit CLASSIC→DISKLESS direct switch when consolidation is on and remote storage is being enabled in the same request.

### DIFF
--- a/storage/src/main/java/org/apache/kafka/storage/internals/log/LogConfig.java
+++ b/storage/src/main/java/org/apache/kafka/storage/internals/log/LogConfig.java
@@ -579,10 +579,21 @@ public class LogConfig extends AbstractConfig {
         }
 
         public boolean isSwitchedFromClassicWithRemoteStorage() {
-            return isDisklessAllowFromClassicEnabled
-                && isDisklessEnabled()
-                && wasRemoteStorageExplicitlySet() && wasRemoteStorageEnabled()
-                && requestedRemoteStorageEnabled();
+            // Allows both diskless and remote-storage to be set when:
+            // - The allow-from-classic flag is on, AND
+            // - Diskless is being enabled, AND
+            // - Remote-Storage was already enabled (TIERED→DISKLESS switch), OR
+            //   Remote-Storage is being enabled in the same request
+            //   AND consolidation is on (CLASSIC→DISKLESS direct switch)
+            if (!isDisklessAllowFromClassicEnabled || !isDisklessEnabled() || !requestedRemoteStorageEnabled()) {
+                return false;
+            }
+            // TIERED→DISKLESS: Remote-Storage was already explicitly set and enabled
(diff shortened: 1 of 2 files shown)


storage/src/main/java/org/apache/kafka/storage/internals/log/LogConfig.java
```diff
@@ -679,6 +697,29 @@ private static void validateDisklessAndRemoteStorageMutualExclusion(LogConfigHel
...
+            return;
+        }
+        boolean isRemoteStorageExplicitlyFalse = logConfigHelper.wasRemoteStorageExplicitlySet()
+            || logConfigHelper.isRemoteStorageExplicitlySet();
+        if (!isRemoteStorageExplicitlyFalse) {
+            return;
+        }
+        boolean isDisklessBeingEnabled = logConfigHelper.isCreation()
+            || (logConfigHelper.isDisklessExplicitlySet() && !logConfigHelper.wasDisklessEnabled());
+        boolean isRemoteStorageBeingDisabled = logConfigHelper.isRemoteStorageExplicitlySet() && !logConfigHelper.isRemoteStorageEnabled();
```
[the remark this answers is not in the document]
jeqo: Intentionally kept for local readability — the variable definition is self-contained without needing to know about the guard clause above. If the guard is ever refactored, this conjunct becomes load-bearing.

storage/src/main/java/org/apache/kafka/storage/internals/log/LogConfig.java
```diff
@@ -679,6 +702,29 @@ private static void validateDisklessAndRemoteStorageMutualExclusion(LogConfigHel
         }
     }
 
+    private static void validateDisklessRequiresRemoteStorage(LogConfigHelper logConfigHelper) {
+        // Diskless topics must have remote storage enabled.
+        // Only reject when remote.storage.enable is explicitly set to false.
+        // If remote.storage.enable was never set (implicit default), allow —
+        // the controller will auto-enable remote storage via config record.
```
giuseppelillo, line 709: When does this happen? Could not find anything

jeqo: it's coming as a follow-up: https://github.com/aiven/inkless/pull/619

giuseppelillo (approved):
LGTM. After this feature is code-completed we will need to test it in an integration test like `kafka.server.DisklessAndRemoteStorageConfigsTest`.

PR #617, opened 2026-05-28
test(inkless:switch): add consolidated diskless integration tests with transition matrix

jeqo:
Add integration tests in DisklessAndRemoteStorageConfigsTest that exercise the full controller+broker path for topic type transitions under consolidation. 
Cover all 9 transition scenarios including allow-from-classic gate. 
Document the transition matrix as Javadoc on both LogConfigTest and the integration test class.

### DIFF
--- a/core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
+++ b/core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
@@ -223,19 +223,6 @@ public class DisklessAndRemoteStorageConfigsTest {
             }
         }
 
-        private Optional<String> incrementalAlterTopicConfig(Admin admin, String topic, Map<String, String> newConfigs) {
-            var topicResource = new ConfigResource(ConfigResource.Type.TOPIC, topic);
-            var operations = newConfigs.entrySet().stream()
-                .map(entry -> new AlterConfigOp(new ConfigEntry(entry.getKey(), entry.getValue()), AlterConfigOp.OpType.SET))
-                .toList();
-            try {
-                admin.incrementalAlterConfigs(Map.of(topicResource, operations)).all().get(10, TimeUnit.SECONDS);
-                return Optional.empty();
-            } catch (Exception e) {
-                String message = e.getCause() != null ? e.getCause().getMessage() : e.getMessage();
-                return Optional.ofNullable(message);
-            }
-        }
     }
 
     private KafkaClusterTestKit initCluster() throws Exception {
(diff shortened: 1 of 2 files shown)


core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
```diff
@@ -293,6 +293,255 @@ private void createTopicAndAssertEffective(Admin admin,
...
+     * 3  | (none)     → DISKLESS  | diskless.enable=true, remote.storage.enable=false           | REJECTED                   | testCreateDisklessWithExplicitRemoteFalseRejected
+     * 4  | CLASSIC    → DISKLESS  | allow-from-classic=true                                     | VALID (switch)             | testClassicToDisklessSwitch
+     * 5  | CLASSIC    → DISKLESS  | allow-from-classic=false                                    | REJECTED                   | testClassicToDisklessBlockedWithoutAllowFromClassic
+     * 6  | TIERED     → DISKLESS  | allow-from-classic=true                                     | VALID (switch)             | testTieredToDisklessSwitch
+     * 7  | TIERED     → DISKLESS  | allow-from-classic=false                                    | REJECTED                   | testTieredToDisklessBlockedWithoutAllowFromClassic
+     * 8  | DISKLESS   → forbidden | remote.storage.enable=false                                 | REJECTED (mutual exclusion)| testDisklessCannotDisableRemoteStorage
+     * 9  | DISKLESS   → TIERED    | diskless.enable=false                                       | REJECTED (unsupported)     | testDisklessCannotDisableDiskless
```
viktorsomogyi: nit: the test is called `testDisklessCannotBeDisabled`?

core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
```diff
@@ -293,6 +293,255 @@ private void createTopicAndAssertEffective(Admin admin,
...
+     * </pre>
+     */
+    @Nested
+    class ConsolidatedDisklessTopics {
+        @Test
+        void testCreateDisklessAutoEnablesRemoteStorage() throws Exception {
+            var cluster = initConsolidatedCluster();
```
viktorsomogyi: Creating a cluster for every test seems a bit expensive. Do you think we could handle these tests similarly to other existing tests or was there a specific reason you choose this way?

core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
```diff
@@ -293,6 +293,255 @@ private void createTopicAndAssertEffective(Admin admin,
...
+                .setConfigProp(InklessConfig.PREFIX + InklessConfig.STORAGE_PREFIX + S3StorageConfig.AWS_ACCESS_KEY_ID_CONFIG, s3Container.getAccessKey())
+                .setConfigProp(InklessConfig.PREFIX + InklessConfig.STORAGE_PREFIX + S3StorageConfig.AWS_SECRET_ACCESS_KEY_CONFIG, s3Container.getSecretKey())
+                .build();
+            cluster.format();
+            cluster.startup();
+            cluster.waitForReadyBrokers();
+            return cluster;
+        }
+
+        private Optional<String> incrementalAlterTopicConfig(Admin admin, String topic, Map<String, String> newConfigs) {
```
viktorsomogyi: Seems like this method is identical to the one in `IncrementalAlterConfigs` class. Would you mind refactoring them to be able to reuse it? Perhaps it could be a static method in `DisklessAndRemoteStorageConfigTest`.

core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
```diff
@@ -293,6 +293,255 @@ private void createTopicAndAssertEffective(Admin admin,
...
+                .setConfigProp(InklessConfig.PREFIX + InklessConfig.STORAGE_PREFIX + S3StorageConfig.AWS_ACCESS_KEY_ID_CONFIG, s3Container.getAccessKey())
+                .setConfigProp(InklessConfig.PREFIX + InklessConfig.STORAGE_PREFIX + S3StorageConfig.AWS_SECRET_ACCESS_KEY_CONFIG, s3Container.getSecretKey())
+                .build();
+            cluster.format();
+            cluster.startup();
+            cluster.waitForReadyBrokers();
+            return cluster;
+        }
+
+        private KafkaClusterTestKit initConsolidatedCluster() throws Exception {
```
viktorsomogyi: This is very similar to `initConsolidatedClusterWithoutAllowFromClassic`. Do you think it could be refactored in such a way that they build on each other? As I see, the difference is only that `DISKLESS_ALLOW_FROM_CLASSIC_ENABLE_CONFIG` is `true` in this one.

core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
jeqo: Good point, we can safely reuse the cluster I think. Let me try that.

core/src/test/java/kafka/server/DisklessAndRemoteStorageConfigsTest.java
jeqo: Yeap, will refactor this

viktorsomogyi (approved):

PR #624, opened 2026-05-29
refactor(inkless): extract Inkless tests from ReplicaManagerTest into standalone file

jeqo:
- 94+2 tests moved, reducing ReplicaManagerTest from ~11k to ~6.1k lines
- Zero Inkless code left in the original file
- CI glob *Inkless* still matches


.github/workflows/inkless.yml
```diff
@@ -163,8 +163,33 @@ jobs:
           # inkless tests
           timeout ${TIMEOUT_MINUTES}m \
           ./gradlew ${GRADLE_ARGS} :storage:inkless:test :storage:inkless:integrationTest && \
-          ./gradlew ${GRADLE_ARGS} :metadata:test --tests "org.apache.kafka.controller.*" && \
-          ./gradlew ${GRADLE_ARGS} :core:test --tests "*Inkless*" --tests "*Diskless*" --tests "kafka.log.LogConfigTest" --tests "kafka.server.KafkaConfigTest"
+          ./gradlew ${GRADLE_ARGS} :metadata:test \
+            --tests "org.apache.kafka.controller.*" \
+            --tests "org.apache.kafka.controller.metrics.*" \
```
viktorsomogyi: nit: seems redundant as `org.apache.kafka.controller.*` would apply to the `metrics` package possibly too?

jeqo: Yes, good catch. Checking gradle docs, that's how it should work: 

> Note that the wildcard * has no special understanding of the . package separator. It's purely text based. So --tests *.SomeTestClass will match any package, regardless of its 'depth'.

viktorsomogyi (approved):

PR #626, opened 2026-06-02
fix(inkless:switch): bump leader epoch on classic-to-diskless switch

jeqo:
markClassicToDisklessSwitchStarted emits a PartitionChangeRecord without setting the leader field, so PartitionRegistration.merge() sees NO_LEADER_CHANGE and skips the leaderEpoch bump. Brokers only call makeLeader on epoch changes, leaving the partition stuck in PENDING.

This was latent since the method was introduced, but was masked by BrokerMetadataPublisher.sealExistingLeadersOfTopicsSwitchedToDiskless() which triggered the switch via config-delta scan — independent of epoch. PR #604 (ad647a42) removed that path and moved all switch logic into applyLocalLeadersDelta, which requires isNewLeaderEpoch == true. The controller was never updated to provide it.

Fix: set .setLeader(partition.leader) to force an epoch bump. The broker sees the new epoch, enters makeLeader, seals the log, and registers with InitDisklessLogManager. Also warn when the partition has no leader at switch time — the PENDING state persists and completes once a leader is elected.


viktorsomogyi (approved):
Nice catch! And it is actually needed for my leader epoch issue as well 🙂
(LGTM, didn't find any issues, this is a small and focused PR)

PR #631, opened 2026-06-05
feat(inkless): KC-156 diskless leader epoch for consolidation truncation

viktorsomogyi:
Consolidating diskless followers had no way to truncate a stale classic tail past the seal: diskless records are produced with leader epoch 0, so appending them after a classic prefix broke epoch-cache monotonicity and disabled OffsetsForLeaderEpoch divergence truncation.

Introduce a captured, control-plane-owned diskless leader epoch (E_d):

- Persist E_d in KRaft metadata as a new tagged field (tag 102) on PartitionRegistration; the controller freezes the partition's current leader epoch at the initDisklessLog commit, where it is already greater than every classic-prefix epoch and robust to leader changes between the switch phases. merge() preserves it across later change records.
- Expose it on the broker via InklessMetadataView/ReplicaManager.
- Stamp E_d onto materialized batches in ConsolidationFetcherThread so the local log keeps a monotonic epoch lineage (in place, reusing the existing flatten; partitionLeaderEpoch is outside the batch CRC).
- Implement DisklessLeaderEndPoint.fetchEpochEndOffsets: a queried epoch below E_d returns the seal (truncating a stale tail back to it), while E_d / born-diskless returns the diskless LEO.

E_d lives only in KRaft metadata; no control-plane schema change is needed since both the stamping and epoch-query paths run on the broker.

### DIFF
--- a/core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationFetcherThread.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationFetcherThread.scala
@@ -22,6 +22,8 @@ import io.aiven.inkless.consume.ConcatenatedRecords
 import kafka.server.{FailedPartitions, KafkaConfig, ReplicaFetcherThread, ReplicaManager, ReplicaQuota}
 import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.record.{MemoryRecords, Records}
+import org.apache.kafka.common.requests.FetchResponse
+import org.apache.kafka.metadata.PartitionRegistration
 import org.apache.kafka.server.LeaderEndPoint
 import org.apache.kafka.storage.internals.log.LogAppendInfo
 
@@ -51,6 +53,7 @@ class ConsolidationFetcherThread(name: String,
     partitionLeaderEpoch: Int,
     partitionData: FetchData
   ): Option[LogAppendInfo] = {
+    maybeStampDisklessLeaderEpoch(topicPartition, partitionData)
     val result = super.processPartitionData(topicPartition, fetchOffset, partitionLeaderEpoch, partitionData)
 
     consolidationMetrics.foreach { metrics =>
@@ -69,5 +72,27 @@ class ConsolidationFetcherThread(name: String,
     }
 
     result
+  }
+
+  /**
+   * Stamp the frozen diskless leader epoch (E_d) onto materialized batches so the local log keeps a
+   * monotonic epoch lineage (diskless records are produced with epoch 0, which would otherwise break
+   * the LeaderEpochFileCache after a switched partition's higher classic epochs and disable divergence
+   * truncation). Born-diskless / not-yet-switched partitions (E_d == NO_DISKLESS_LEADER_EPOCH) are left
+   * at epoch 0. Done in place to reuse the append path's single flatten; partitionLeaderEpoch is outside
+   * the batch CRC, so no checksum recompute is needed.
+   */
+  private def maybeStampDisklessLeaderEpoch(topicPartition: TopicPartition, partitionData: FetchData): Unit = {
+    val disklessLeaderEpoch = replicaMgr.disklessLeaderEpoch(topicPartition)
+    if (disklessLeaderEpoch == PartitionRegistration.NO_DISKLESS_LEADER_EPOCH) {
+      return
+    }
+    FetchResponse.recordsOrFail(partitionData) match {
+      case records: ConcatenatedRecords =>
+        records.batches().forEach(batch => batch.setPartitionLeaderEpoch(disklessLeaderEpoch))
+      case records: MemoryRecords =>
+        records.batches().forEach(batch => batch.setPartitionLeaderEpoch(disklessLeaderEpoch))
+      case _ =>
+    }
   }
 }
--- a/core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationFetcherThread.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationFetcherThread.scala
@@ -75,7 +75,7 @@ class ConsolidationFetcherThread(name: String,
   }
 
   /**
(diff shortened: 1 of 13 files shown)


metadata/src/main/java/org/apache/kafka/metadata/PartitionRegistration.java
```diff
@@ -159,13 +165,17 @@ public PartitionRegistration build() {
...
+                disklessLeaderEpoch
             );
         }
     }
 
     public static final long NO_CLASSIC_TO_DISKLESS_START_OFFSET = -1L;
     public static final long CLASSIC_TO_DISKLESS_SWITCH_PENDING = -2L;
+    // Sentinel for a partition that has no frozen diskless leader epoch (never switched, born-diskless,
+    // or switch still pending). A real diskless leader epoch is always >= 0.
+    public static final int NO_DISKLESS_LEADER_EPOCH = -1;
```
giuseppelillo: Would it be possible to assign 0 for born-diskless partitions? In this way this is not only useful for classic to diskless switch but for all diskless partitions

metadata/src/main/java/org/apache/kafka/metadata/PartitionRegistration.java
```diff
@@ -159,13 +165,17 @@ public PartitionRegistration build() {
...
+                disklessLeaderEpoch
             );
         }
     }
 
     public static final long NO_CLASSIC_TO_DISKLESS_START_OFFSET = -1L;
     public static final long CLASSIC_TO_DISKLESS_SWITCH_PENDING = -2L;
+    // Sentinel for a partition that has no captured diskless leader epoch (never switched, born-diskless,
+    // or switch still pending). A real diskless leader epoch is always >= 0.
+    public static final int NO_DISKLESS_LEADER_EPOCH = -1;
```
giuseppelillo, line 178: Would it be possible to assign 0 for born-diskless partitions? In this way this is not only useful for classic to diskless switch but for all diskless partitions

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -1617,6 +1617,14 @@ ControllerResult<InitDisklessLogResponseData> initDisklessLog(
                             ps.batchMaxTimestamp()))
                         .toList();
 
+                // Capture the partition's current leader epoch as the diskless leader epoch (E_d). The
+                // classic-to-diskless switch already bumped the leader epoch (Phase A), so this value is
```
giuseppelillo: What is phase A exactly?

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationFetcherThread.scala
```diff
@@ -70,4 +73,26 @@ class ConsolidationFetcherThread(name: String,
...
+      return
+    }
+    FetchResponse.recordsOrFail(partitionData) match {
+      case records: ConcatenatedRecords =>
+        records.batches().forEach(batch => batch.setPartitionLeaderEpoch(disklessLeaderEpoch))
+      case records: MemoryRecords =>
+        records.batches().forEach(batch => batch.setPartitionLeaderEpoch(disklessLeaderEpoch))
+      case _ =>
+    }
+  }
```
giuseppelillo, line 97: How hard would it be to instead use the real leader epoch also for diskless records? So we don't need to do this, and so E_d is only relevant to understand if the records' LE is > E_d for the truncation

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
viktorsomogyi: Ok, this is a dangling reference. I started implementing this before #626 which supposed to be phase A (or at least something similar) but it landed sooner and I refactored the code but forgot to remove this reference.
But phase A: bump the leader epoch in initDiskless
phase B: store the leader epoch and stamp
This is just fyi, I'll remove this reference.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationFetcherThread.scala
viktorsomogyi: So in classic Kafka the flow is that the partition leader stamps the records with the epoch information and the followers populate their own cache based on what's in the record. What's more, it is assigned when offsets are assigned, so around production. In diskless we assign the offsets later, so that already breaks this flow, because by that time someone else could be the "leader".
Also, in diskless every replica is basically a follower of the diskless control plane. So if we wanted to stamp the real leader epoch here, we must query it from the leader first or query the metadata and find who is the current leader. This introduces a race condition, because we can't be sure that the metadata is up to date, so we could stamp something stale. But again, that leader epoch might not have been up to date even during production due to the late assignment in diskless.
All in all, these situations may cause non-monotonic leader epochs. When we switch back to classic though (E_c2), we should be able to use the real leader epoch though because we can again do a leader epoch bump just before switching back (in theory), so then we could reinstate the monotonicity as the order would be E_c < E_d < E_c2 (where E_c is the classic epoch before the switch to diskless).

metadata/src/main/java/org/apache/kafka/metadata/PartitionRegistration.java
viktorsomogyi: I think it would be possible but I'm not sure about what we gain. How can we use it in the future?
Currently using `NO_DISKLESS_LEADER_EPOCH` covers the functionality that we need and we can also differentiate between born-diskless and switched-to-diskless partitions in this case (seal < 0). I'm afraid I don't really see a good use-case for it right now but let me know if you think otherwise or have a future use-case in mind that is not on my radar.

giuseppelillo: Not a real use case, it's just for overall clarity and coherence

giuseppelillo (approved):
LGTM for the current use case, we can tweak it later if necessary for other use cases

PR #633, opened 2026-06-05
refactor(inkless): clarify fetch routing for consolidating partitions

jeqo:
Some minor changes to make fetch messages more readable:
- Added logging on failing responses,
- Reduce variables serving same purpose to reduce cognitive load
- Renamed some variables for clarity

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2108,12 +2108,7 @@ class ReplicaManager(val config: KafkaConfig,
     def respond(response: Seq[(TopicIdPartition, FetchPartitionData)]): Unit =
       responseCallback(response ++ immediateFetchResponses)
 
-    if (classicFetchInfos.isEmpty && disklessFetchInfos.isEmpty && immediateFetchResponses.nonEmpty) {
-      respond(Seq.empty)
-      return
-    }
-
-    if (classicFetchInfos.isEmpty && disklessFetchInfos.isEmpty && immediateFetchResponses.isEmpty) {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2068,32 +2099,29 @@ class ReplicaManager(val config: KafkaConfig,
...
 
-    if (classicFetchInfos.isEmpty && disklessFetchInfosWithoutTopicId.isEmpty && immediateFetchResponses.nonEmpty) {
+    if (classicFetchInfos.isEmpty && disklessFetchInfos.isEmpty && immediateFetchResponses.nonEmpty) {
       respond(Seq.empty)
       return
     }
 
-    if (classicFetchInfos.isEmpty && disklessFetchInfosWithoutTopicId.isEmpty &&
-      immediateFetchResponses.isEmpty && invalidConsolidatingPartitionFetchResponses.nonEmpty) {
+    if (classicFetchInfos.isEmpty && disklessFetchInfos.isEmpty && immediateFetchResponses.isEmpty) {
```
viktorsomogyi: I think you can collapse these into one `if (classicFetchInfos.isEmpty && disklessFetchInfos.isEmpty)` because we `respond(Seq.empty)` regardless of the value of `immediateFetchResponses`.

jeqo: Good catch, even better.

viktorsomogyi (approved):
LGTM. Reran the Java 17 build because there was a flaky ReplicaManagerTest.testReplicaAlterLogDirsMultipleReassignmentDoesNotBlockLogCleaner(). I've seen this failing locally a couple of times too, it's some mocking wiring issue.

PR #634, opened 2026-06-05
feat(inkless): add controller guards for classic-to-diskless switch [POD-2464]

EelisK:
Enforce the following guards when altering the topic config to use diskless.enable=true:
  - Partitions are online
  - No reassignment in progress
  - Enough ISRs

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/ConfigurationControlManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/ConfigurationControlManager.java
@@ -54,6 +54,7 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 import static org.apache.kafka.clients.admin.AlterConfigOp.OpType.APPEND;
 import static org.apache.kafka.common.config.ConfigResource.Type.BROKER;
@@ -201,6 +202,14 @@ public class ConfigurationControlManager {
         Map<ConfigResource, Map<String, Entry<OpType, String>>> configChanges,
         boolean newlyCreatedResource
     ) {
+        return incrementalAlterConfigs(configChanges, newlyCreatedResource, null);
+    }
+
+    ControllerResult<Map<ConfigResource, ApiError>> incrementalAlterConfigs(
+        Map<ConfigResource, Map<String, Entry<OpType, String>>> configChanges,
+        boolean newlyCreatedResource,
+        Function<ConfigResource, ApiError> postConfigValidation
+    ) {
         List<ApiMessageAndVersion> outputRecords =
                 BoundedList.newArrayBacked(MAX_RECORDS_PER_USER_OP);
         Map<ConfigResource, ApiError> outputResults = new HashMap<>();
@@ -209,7 +218,8 @@ public class ConfigurationControlManager {
             ApiError apiError = incrementalAlterConfigResource(resourceEntry.getKey(),
                 resourceEntry.getValue(),
                 newlyCreatedResource,
-                outputRecords);
+                outputRecords,
+                postConfigValidation);
             outputResults.put(resourceEntry.getKey(), apiError);
         }
         outputRecords.addAll(createClearElrRecordsAsNeeded(outputRecords));
@@ -248,7 +258,8 @@ public class ConfigurationControlManager {
         ApiError apiError = incrementalAlterConfigResource(configResource,
             keyToOps,
             newlyCreatedResource,
-            outputRecords);
+            outputRecords,
+            null);
 
         outputRecords.addAll(createClearElrRecordsAsNeeded(outputRecords));
         return ControllerResult.atomicOf(outputRecords, apiError);
@@ -258,7 +269,8 @@ public class ConfigurationControlManager {
         ConfigResource configResource,
         Map<String, Entry<OpType, String>> keysToOps,
         boolean newlyCreatedResource,
-        List<ApiMessageAndVersion> outputRecords
+        List<ApiMessageAndVersion> outputRecords,
+        Function<ConfigResource, ApiError> postConfigValidation
     ) {
         List<ApiMessageAndVersion> newRecords = new ArrayList<>();
         for (Entry<String, Entry<OpType, String>> keysToOpsEntry : keysToOps.entrySet()) {
@@ -328,6 +340,12 @@ public class ConfigurationControlManager {
         ApiError error = validateAlterConfig(configResource, newRecords, List.of(), newlyCreatedResource);
         if (error.isFailure()) {
             return error;
+        }
+        if (postConfigValidation != null) {
+            error = postConfigValidation.apply(configResource);
+            if (error.isFailure()) {
+                return error;
+            }
         }
         outputRecords.addAll(newRecords);
         return ApiError.NONE;
@@ -450,6 +468,14 @@ public class ConfigurationControlManager {
         Map<ConfigResource, Map<String, String>> newConfigs,
         boolean newlyCreatedResource
     ) {
+        return legacyAlterConfigs(newConfigs, newlyCreatedResource, null);
+    }
+
+    ControllerResult<Map<ConfigResource, ApiError>> legacyAlterConfigs(
+        Map<ConfigResource, Map<String, String>> newConfigs,
+        boolean newlyCreatedResource,
+        Function<ConfigResource, ApiError> postConfigValidation
+    ) {
         List<ApiMessageAndVersion> outputRecords =
                 BoundedList.newArrayBacked(MAX_RECORDS_PER_USER_OP);
         Map<ConfigResource, ApiError> outputResults = new HashMap<>();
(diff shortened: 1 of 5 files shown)


metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -2251,6 +2253,13 @@ void maybeTriggerUncleanLeaderElectionForLeaderlessPartitions(
         while (iterator.hasNext() && records.size() < maxElections) {
             TopicIdPartition topicIdPartition = iterator.next();
             TopicControlInfo topic = topics.get(topicIdPartition.topicId());
+            PartitionRegistration partition = topic.parts.get(topicIdPartition.partitionId());
+            if (partition != null && hasClassicToDisklessSwitchPending(partition)) {
+                log.debug("Skipping unclean leader election for partition {}-{} " +
```
giuseppelillo:
```suggestion
                log.warn("Skipping unclean leader election for partition {}-{} " +
```

This sounds like a pretty important thing to log

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -2269,6 +2278,10 @@ void maybeTriggerUncleanLeaderElectionForLeaderlessPartitions(
         }
     }
 
+    private static boolean hasClassicToDisklessSwitchPending(PartitionRegistration partition) {
+        return partition.classicToDisklessStartOffset != PartitionRegistration.NO_CLASSIC_TO_DISKLESS_START_OFFSET;
```
giuseppelillo:
```suggestion
        return partition.classicToDisklessStartOffset == PartitionRegistration. CLASSIC_TO_DISKLESS_SWITCH_PENDING;
```

this also returns true if classicToDisklessStartOffset >=0 , which is the value for already switched partitions, but we only want to block it for partitions that are being switched.

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -1447,7 +1448,8 @@ ControllerResult<AlterPartitionResponseData> alterPartition(
                     getTopicEffectiveMinIsr(topic.name)
                 )
                     .setEligibleLeaderReplicasEnabled(featureControl.isElrFeatureEnabled());
-                if (configurationControl.uncleanLeaderElectionEnabledForTopic(topic.name())) {
+                if (configurationControl.uncleanLeaderElectionEnabledForTopic(topic.name())
+                        && !hasClassicToDisklessSwitchPending(partition)) {
```
giuseppelillo: Add a log at warn level here too in case hasClassicToDisklessSwitchPending is true

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -2499,7 +2512,8 @@ void generateLeaderAndIsrUpdates(String context,
                 getTopicEffectiveMinIsr(topic.name)
             );
             builder.setEligibleLeaderReplicasEnabled(featureControl.isElrFeatureEnabled());
-            if (configurationControl.uncleanLeaderElectionEnabledForTopic(topic.name)) {
+            if (configurationControl.uncleanLeaderElectionEnabledForTopic(topic.name)
+                    && !hasClassicToDisklessSwitchPending(partition)) {
```
giuseppelillo: Add a log at warn level here too in case hasClassicToDisklessSwitchPending is true

metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
```diff
@@ -1952,8 +1952,13 @@ public CompletableFuture<Map<ConfigResource, ApiError>> incrementalAlterConfigs(
             return CompletableFuture.completedFuture(Map.of());
         }
         return appendWriteEvent("incrementalAlterConfigs", context.deadlineNs(), () -> {
+            Map<ConfigResource, ApiError> preconditionErrors =
+                    replicationControl.validateClassicToDisklessSwitchPreconditions(configChanges);
+            configChanges.keySet().removeAll(preconditionErrors.keySet());
+
             ControllerResult<Map<ConfigResource, ApiError>> result =
                 configurationControl.incrementalAlterConfigs(configChanges, false);
+            result.response().putAll(preconditionErrors);
```
giuseppelillo: Even though it logically makes sense to first validate and then apply incrementalAlterConfigs logic, it would break the error messages for some particular cases, like this one:
e.g. `diskless.allow.from.classic.enable=false` and the topic is under-replicated => this code will classify the topic as "enabling diskless”" return the under-replicated precondition error, and skip ConfigurationControlManager, which would have returned “It is invalid to enable diskless on an already existing topic.”

So the correct order here would be:
1. Run config validation first (`configurationControl.incrementalAlterConfigs`)
2. Only for resources whose config change is otherwise valid, run the replication-state preconditions
3. If preconditions fail, do not emit that resource’s ConfigRecord.

giuseppelillo (changes requested):

metadata/src/main/java/org/apache/kafka/controller/ConfigurationControlManager.java
```diff
@@ -503,6 +531,13 @@ private void legacyAlterConfigResource(ConfigResource configResource,
             outputResults.put(configResource, error);
             return;
         }
+        if (preCommitCheck != null) {
+            error = preCommitCheck.apply(configResource);
+            if (error.isFailure()) {
+                outputResults.put(configResource, error);
+                return;
```
viktorsomogyi, line 538: Why is the early return?

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -2269,6 +2279,16 @@ void maybeTriggerUncleanLeaderElectionForLeaderlessPartitions(
...
 
+    private static boolean hasClassicToDisklessSwitchPending(PartitionRegistration partition) {
+        return partition.classicToDisklessStartOffset == PartitionRegistration.CLASSIC_TO_DISKLESS_SWITCH_PENDING;
+    }
+
+    private void warnSkippingUncleanElectionForPendingSwitch(String topicName, int partitionId) {
+        log.warn("Skipping unclean leader election for partition {}-{} " +
+                "because it has a pending classic-to-diskless switch.",
+            topicName, partitionId);
+    }
```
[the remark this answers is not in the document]
EelisK: I think we can keep it as a warning, since the situation should be ephemeral and spam indicates an actual fault

metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
EelisK: Thanks for the explanation 👍 addressed

metadata/src/main/java/org/apache/kafka/controller/ConfigurationControlManager.java
EelisK: Like above! To avoid making config changes when the validation fails. Or have I misunderstood something?

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -2251,6 +2256,11 @@ void maybeTriggerUncleanLeaderElectionForLeaderlessPartitions(
         while (iterator.hasNext() && records.size() < maxElections) {
             TopicIdPartition topicIdPartition = iterator.next();
             TopicControlInfo topic = topics.get(topicIdPartition.topicId());
+            PartitionRegistration partition = topic.parts.get(topicIdPartition.partitionId());
+            if (partition != null && hasClassicToDisklessSwitchPending(partition)) {
+                warnSkippingUncleanElectionForPendingSwitch(topic.name, topicIdPartition.partitionId());
```
viktorsomogyi, line 2267: `electLeader` can be called from a different path too, from `electLeaders`. Should we add a guard there too?

EelisK: Added here https://github.com/aiven/inkless/pull/634/commits/b7f85701c6477a6716f9d8720670a3a834322c8e ✅

metadata/src/main/java/org/apache/kafka/controller/ConfigurationControlManager.java
viktorsomogyi: Not at all, I missed the context, dumb comment on my part 🙂

viktorsomogyi (dismissed):
LGTM on my part but let's wait for Giuseppe's verdict as he requested some changes earlier.

giuseppelillo (changes requested):
Overall LGTM, I like the approach.

However I think there's a latent bug in the Legacy AlterConfigs (which is present also in the already existing `markClassicToDisklessSwitchStartedForLegacyAlterConfigs`): 
Legacy AlterConfigs deletes existing configs that are missing from the request. If a classic topic has explicit `diskless.enable=false` and the broker default is `log.diskless.enable=true`, a legacy request that only changes another config and omits diskless.enable can effectively switch the topic to diskless, but it won't be caught.

Could you add a test about this and verify that is a real bug, and if so fix it?

EelisK:
@giuseppelillo good catch, it was indeed an issue 👍 I added the fix and tests https://github.com/aiven/inkless/pull/634/commits/e92e01cc51ef93e7de44cbdbd4c4d66c209ea4ee

giuseppelillo (approved):

PR #635, opened 2026-06-05
feat(inkless:consolidation): add separate wiring for consolidation fetcher [KC-160]

jeqo:
Consolidation fetcher gets its own configuration surface independent of the shared consumer fetch path:
- diskless.consolidation.fetch.max.bytes (default: 1MB, per-partition)
- diskless.consolidation.fetch.response.max.bytes (default: 10MB, response-level)
- diskless.consolidation.fetch.min.bytes (default: 1)
- diskless.consolidation.fetch.max.wait.ms (default: 500ms)
- diskless.consolidation.num.fetchers (default: 1)
- diskless.consolidation.fetch.find.batches.max.per.partition (default: 0 = unlimited)

Per-partition and response-level max bytes are separate configs (same pattern as replica.fetch.max.bytes vs replica.fetch.response.max.bytes) so they can be tuned independently when multiple partitions share one fetcher thread.

Separate metrics group (ConsolidationFetchMetrics) so consolidation fetch metrics are distinguishable from consumer fetch metrics in JMX.

Reader gains a constructor accepting KafkaMetricsGroup for injection, and InklessFetchMetrics accepts a custom metrics group. The consolidation FetchHandler constructs its own Reader with these overrides wired from KafkaConfig, and is properly closed during broker shutdown.


storage/inkless/src/main/java/io/aiven/inkless/consume/Reader.java
```diff
@@ -137,6 +138,35 @@ public Reader(
...
+        long laggingConsumerThresholdMs,
+        int laggingConsumerRequestRateLimit,
+        int laggingConsumerThreadPoolSize,
+        long hedgeTtfbThresholdMs,
+        long hedgeTotalTimeThresholdMs,
+        int maxBatchesPerPartitionToFind,
+        KafkaMetricsGroup metricsGroup,
+        String threadNamePrefix
     ) {
         this(
```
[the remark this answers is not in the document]
jeqo: noise

server-common/src/main/java/org/apache/kafka/server/config/ServerConfigs.java
```diff
@@ -161,6 +161,51 @@ public class ServerConfigs {
         "diskless.enable=true and remote.storage.enable=true on new topics. Setting both will start consolidating Diskless WAL segments into " +
         "Kafka tiered log storage on the configured topic.";
 
+    public static final String DISKLESS_CONSOLIDATION_FETCH_MAX_BYTES_CONFIG = "diskless.consolidation.fetch.max.bytes";
+    public static final int DISKLESS_CONSOLIDATION_FETCH_MAX_BYTES_DEFAULT = 1024 * 1024; // 1MB
+    public static final String DISKLESS_CONSOLIDATION_FETCH_MAX_BYTES_DOC = "The maximum number of bytes per partition the consolidation " +
+        "fetcher will request per iteration. Each fetch is appended to the local log as a single unit, so this must not exceed " +
+        "segment.bytes or the append fails with RecordBatchTooLargeException. Defaults to 1MB to match replica.fetch.max.bytes, which " +
```
viktorsomogyi: Should we add validation to assert this?

jeqo: Good idea. Actually added this to keep track of an issue found when testing with small segment sizes and large batch sizes.
I'd say we should get rid of this validation: the data _is already acknowledged_ so there is no need to validate sizes here. This validation is also ignored on Diskless. We can follow-up with a proper validation / removal of this validation as a separate PR. 

I simplified the comment to keep track of this separately.

viktorsomogyi (approved):

PR #636, opened 2026-06-05
feat(inkless): add dedicated rate limit for consolidation fetch [KC-145]

jeqo:
Introduce a separate ReplicationQuotaManager for consolidation fetchers, decoupled from the follower replication quota. This allows operators to cap object-storage read bandwidth (bytes/sec) without affecting regular replication.

Key changes:
- Add DISKLESS_CONSOLIDATION_FETCH QuotaType (JMX: DisklessConsolidationFetch)
- Add diskless.consolidation.fetch.rate.limit.bytes.per.second config
- Wire dedicated quota manager via ConsolidationReconciler (marks topics throttled on partition add, ensuring bytes are recorded to the sensor)
- Zip consolidationQuotaManager into fetcher manager creation (no silent fallback to shared follower quota)

Config lives in ServerConfigs (not InklessConfig) because:
- It is a broker-level config accessed through KafkaConfig, same layer as num.replica.fetchers and leader.replication.throttled.rate
- It uses the diskless.consolidation.* namespace alongside the existing diskless.remote.storage.consolidation.enable config
- InklessConfig is for storage module internals (fetch pools, cache, upload) consumed by SharedState/Reader/Writer — not for fetcher-layer wiring

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -60,7 +60,9 @@ import org.apache.kafka.metadata.{LeaderAndIsr, MetadataCache, PartitionRegistra
 import org.apache.kafka.metadata.LeaderConstants.NO_LEADER
 import org.apache.kafka.server.common.{DirectoryEventHandler, RequestLocal, StopPartition, TransactionVersion}
 import org.apache.kafka.server.log.remote.TopicPartitionLog
-import org.apache.kafka.server.config.ReplicationConfigs
+import org.apache.kafka.server.config.{ReplicationConfigs, ReplicationQuotaManagerConfig}
+import org.apache.kafka.common.metrics.Quota
+import org.apache.kafka.server.quota.QuotaType
 import org.apache.kafka.server.log.remote.storage.RemoteLogManager
 import org.apache.kafka.server.metrics.KafkaMetricsGroup
 import org.apache.kafka.server.network.BrokerEndPoint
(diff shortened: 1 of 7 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -319,15 +321,38 @@ class ReplicaManager(val config: KafkaConfig,
...
     }
+  private val consolidationQuotaManager: Option[ReplicationQuotaManager] =
+    if (config.disklessRemoteStorageConsolidationEnabled) {
+      val quotaConfig = new ReplicationQuotaManagerConfig(
+        config.quotaConfig.numReplicationQuotaSamples,
+        config.quotaConfig.replicationQuotaWindowSizeSeconds
+      )
+      val manager = new ReplicationQuotaManager(quotaConfig, metrics, QuotaType.DISKLESS_CONSOLIDATION_FETCH, time)
+      val rateLimitBytesPerSec = config.disklessConsolidationFetchRateLimitBytesPerSecond
+      manager.updateQuota(new Quota(rateLimitBytesPerSec, true))
```
viktorsomogyi, line 332: Do you think we'd eventually need to add consolidation quotas as dynamic configs?

jeqo: It's debatable how much/soon we would need this, but I'd say for completeness we should have it. Can be done as a follow-up as this PR is large enough and would require moving manager into `QuotaManagers`. wdyt?

viktorsomogyi: Sure, I'm fine with that.

jeqo: Working on it already, will open a new PR after this is merged 👍🏽

viktorsomogyi (approved):

PR #638, opened 2026-06-10
feat(inkless:consolidation): supplement local log with diskless data on fetch [KC-168]

jeqo:
## Problem

At the local/diskless boundary of a consolidating partition, consumers fetching with large minBytes would park in the delayed-fetch purgatory and wait the full maxWaitMs before getting a response — even when diskless data was already available past the local logEndOffset.

## Solution

When the local log read doesn't satisfy minBytes, supplement with a synchronous diskless fetch starting at logEndOffset and merge the results via `ConcatenatedRecords.concat()`.

For requests spanning only consolidating partitions, the supplement is applied inline and the response is produced immediately without entering the purgatory. For mixed requests (consolidating + pure-diskless), the supplement and the diskless fetch run concurrently in `DelayedFetch.onComplete` so latency equals the slower of the two, not their sum.

### Inline supplement in `fetchMessages`
- Fires when consolidating partitions are below minBytes and no pure-diskless partitions are present
- Guards: `!isFromFollower`, `maxWaitMs > 0`, `!hasPreferredReadReplica`, `!errorReadingData`
- `inklessSharedState` checked at collection time so supplements are never gathered on a non-inkless broker
- `bytesReadable` only counts error-free supplement bytes — a supplement response with an error cannot trick the broker into responding prematurely
- Exception handling: `InterruptedException` re-interrupts the thread; `TimeoutException` logged specifically for operator visibility; other failures fall back to local-only

### Concurrent supplement in `DelayedFetch.onComplete`
- Supplement and pure-diskless fetches run in parallel via `CompletableFuture.thenCombine`
- `AtomicBoolean` ensures the response callback fires exactly once
- Skips partitions with local read errors or `fetchOffset >= logEndOffset`
- Skips entirely for follower fetches
- On diskless future failure, per-partition error entries (`Errors.forException`) are returned so clients see the partition in the response rather than it being silently dropped

### `ConcatenatedRecords.concat(prefix, tail)` static factory
- Encapsulates type dispatch over `MemoryRecords`/`ConcatenatedRecords`
- Unwraps inner backing records directly — no extra materialization copy
- Validates inputs with `Objects.requireNonNull`

### `buildConsolidationSupplementFetchInfos`
- Computes per-partition remaining byte budget (original maxBytes minus local bytes read)
- Drops partitions with zero budget remaining

### `mergeConsolidationSupplement`
- Materializes local `FileRecords` → `MemoryRecords` (standard `readInto`/`flip` idiom)
- Promotes diskless HW/LSO; uses `Math.min(local, supplement)` for logStartOffset
- Preserves `localData.abortedTransactions` (warns when present since diskless has no transaction index)
- Preserves `localData.preferredReadReplica` (diskless returns empty for this field)
- Falls back to local-only data on unknown Records types or merge failures with error log

## Tests

- `testFetchConsolidatingSupplementRespondsWithoutDelayWhenDisklessDataAvailable`: RED/GREEN — fails without the supplement (parks for maxWaitMs), passes with it (immediate response, purgatory stays empty)
- `testFetchConsolidatingParksUntilMaxWaitWhenNoDisklessSupplementData`: bounds the contract — no diskless data means the request correctly waits
- `testFetchConsolidatingSupplementWithErrorDoesNotInflateBytesReadable`: supplement with error status must not satisfy minBytes — request still parks
- `testFetchConsolidatingSupplementPreservesInterruptedStatus`: thread interrupted status is preserved after InterruptedException in supplement fetch
- `testFetchMixedConsolidatingPureDisklessAndClassicPartitions`: all three partition types in one request, asserts merged consolidating data and two diskless handler calls
- `testCompletionWithConsolidationSupplement`: focused unit test for the concurrent supplement path in `DelayedFetch.onComplete`
- `testConsolidationSupplementSkippedWhenLocalReadHasError`: supplement merge is not applied when local read returns an error
- `testConsolidationSupplementNotIssuedWhenLocalReadHasError`: supplement fetch is not issued for error-result partitions
- `testDisklessFetchFailureReturnsPerPartitionErrors`: diskless future failure produces per-partition error entries, not silent omission
- Unit tests for `buildConsolidationSupplementFetchInfos`: budget deduction, zero-budget drop, missing partitions filtered
- Unit tests for `mergeConsolidationSupplement`: MemoryRecords merge, FileRecords materialization, unknown-type fallback

### DIFF
--- a/core/src/main/scala/kafka/server/DelayedFetch.scala
+++ b/core/src/main/scala/kafka/server/DelayedFetch.scala
@@ -47,6 +47,7 @@ class DelayedFetch(
   params: FetchParams,
   classicFetchPartitionStatus: util.LinkedHashMap[TopicIdPartition, FetchPartitionStatus],
   disklessFetchPartitionStatus: util.LinkedHashMap[TopicIdPartition, FetchPartitionStatus] = new util.LinkedHashMap[TopicIdPartition, FetchPartitionStatus](),
+  consolidatingSupplements: Map[TopicIdPartition, Long] = Map.empty,
   replicaManager: ReplicaManager,
   quota: ReplicaQuota,
   maxWaitMs: Option[Long] = None,
@@ -250,33 +251,13 @@ class DelayedFetch(
       (results, resultMap)
     } else (Seq.empty, new util.HashMap[TopicIdPartition, LogReadResult]())
 
-    // Identify consolidating partitions that still have budget left after the local read and
-    // fire their supplement fetch concurrently with the pure-diskless fetch. Both futures are
-    // independent so running them in parallel keeps latency equal to the slower of the two.
-    // Guards like maxWaitMs>0 and !hasPreferredReadReplica are unnecessary here: by the time
-    // onComplete fires, those conditions were already true (maxWaitMs was positive to park,
-    // preferredReadReplica triggers immediate response and never parks).
-    val consolidatingSupplements = {
-      val supplements = new mutable.HashMap[TopicIdPartition, Long]()
-      if (!params.isFromFollower) {
-        classicFetchPartitionStatus.asScala.foreach { case (tp, status) =>
-          if (replicaManager.inklessMetadataView().isConsolidatingDisklessTopic(tp.topic)) {
-            val readResult = logReadResultMap.get(tp)
-            if (readResult != null && readResult.error == Errors.NONE) {
-              replicaManager.getPartitionOrError(tp.topicPartition).foreach { partition =>
-                partition.log.foreach { log =>
-                  if (status.startOffsetMetadata.messageOffset < log.logEndOffset)
-                    supplements.put(tp, log.logEndOffset)
-                }
-              }
-            }
-          }
-        }
-      }
-      supplements
-    }
-    val supplementFetchInfos = replicaManager.buildConsolidationSupplementFetchInfos(
-      consolidatingSupplements, classicFetchInfos.toSeq, logReadResultMap)
+    // Supplement consolidating partitions whose local read has remaining byte budget.
+    // consolidatingSupplements is passed from fetchMessages (identified at routing time),
+    // so no per-partition metadata lookups are needed here.
+    val supplementFetchInfos =
+      if (consolidatingSupplements.nonEmpty)
+        replicaManager.buildConsolidationSupplementFetchInfos(consolidatingSupplements, classicFetchInfos.toSeq, logReadResultMap)
+      else Seq.empty
 
     val emptySupplementMap: Map[TopicIdPartition, FetchPartitionData] = Map.empty
     val supplementFuture: CompletableFuture[Map[TopicIdPartition, FetchPartitionData]] =
--- a/core/src/main/scala/kafka/server/DelayedFetch.scala
+++ b/core/src/main/scala/kafka/server/DelayedFetch.scala
@@ -21,15 +21,17 @@ import com.yammer.metrics.core.Meter
 import io.aiven.inkless.control_plane.FindBatchRequest
 import kafka.utils.Logging
 
-import java.util.concurrent.TimeUnit
+import java.util.concurrent.{CompletableFuture, TimeUnit}
+import java.util.concurrent.atomic.AtomicBoolean
 import org.apache.kafka.common.TopicIdPartition
 import org.apache.kafka.common.errors._
 import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.record.MemoryRecords
 import org.apache.kafka.common.requests.OffsetsForLeaderEpochResponse.{UNDEFINED_EPOCH, UNDEFINED_EPOCH_OFFSET}
 import org.apache.kafka.server.metrics.KafkaMetricsGroup
 import org.apache.kafka.server.purgatory.DelayedOperation
 import org.apache.kafka.server.storage.log.{FetchIsolation, FetchParams, FetchPartitionData}
-import org.apache.kafka.storage.internals.log.{FetchPartitionStatus, LogOffsetMetadata}
+import org.apache.kafka.storage.internals.log.{FetchPartitionStatus, LogOffsetMetadata, LogReadResult}
 
 import java.util
 import scala.collection._
@@ -225,7 +227,6 @@ class DelayedFetch(
    * Upon completion, read whatever data is available and pass to the complete callback
    */
   override def onComplete(): Unit = {
-    // Complete the classic fetches first
     val classicFetchInfos = classicFetchPartitionStatus.asScala.iterator.map { case (tp, status) =>
       tp -> status.fetchInfo
     }.toBuffer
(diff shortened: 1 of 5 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2239,9 +2328,47 @@ class ReplicaManager(val config: KafkaConfig,
...
+      params.maxWaitMs > 0 && // safeguard: non-blocking polls must not be held by a diskless round-trip
+      bytesReadable < params.minBytes &&
+      !errorReadingData) {
+      val supplementFetchInfos = buildConsolidationSupplementFetchInfos(consolidatingLocalFetchSupplements, fetchInfos, logReadResultMap)
+
+      if (supplementFetchInfos.nonEmpty) {
+        try {
+          val supplementParams = fetchParamsWithNewMaxBytes(params, supplementFetchInfos.size.toFloat / fetchInfos.size.toFloat)
+          consolidationSupplementData = fetchDisklessMessages(supplementParams, supplementFetchInfos)
+            .get(Math.max(config.disklessFetchMaxWaitMs.toLong, params.maxWaitMs), TimeUnit.MILLISECONDS)
```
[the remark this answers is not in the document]
jeqo: Intentional design: the supplement needs at least disklessFetchMaxWaitMs to complete a remote storage round-trip. The params.maxWaitMs > 0 guard (line 2339) already prevents this path for non-blocking polls.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2035,8 +2115,14 @@ class ReplicaManager(val config: KafkaConfig,
...
           getPartitionOrError(tp.topicPartition) match {
             case Right(partition) =>
-              shouldReadFromUnifiedLog = shouldReadFromUnifiedLog ||
-                partition.log.exists(fetchPartitionData.fetchOffset < _.logEndOffset)
+              val logEndOffset = partition.log.map(_.logEndOffset).getOrElse(0L)
+              if (fetchPartitionData.fetchOffset < logEndOffset) {
+                // Local log has data for this offset range — serve from local, track for diskless supplement
+                shouldReadFromUnifiedLog = true
+                if (inklessSharedState.isDefined)
+                  consolidatingLocalFetchSupplements += (tp -> logEndOffset)
```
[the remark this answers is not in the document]
jeqo: not a valid scenario

core/src/main/scala/kafka/server/DelayedFetch.scala
```diff
@@ -235,50 +236,127 @@ class DelayedFetch(
...
+    // Guards like maxWaitMs>0 and !hasPreferredReadReplica are unnecessary here: by the time
+    // onComplete fires, those conditions were already true (maxWaitMs was positive to park,
+    // preferredReadReplica triggers immediate response and never parks).
+    val consolidatingSupplements = {
+      val supplements = new mutable.HashMap[TopicIdPartition, Long]()
+      if (!params.isFromFollower) {
+        classicFetchPartitionStatus.asScala.foreach { case (tp, status) =>
+          if (replicaManager.inklessMetadataView().isConsolidatingDisklessTopic(tp.topic)) {
+            val readResult = logReadResultMap.get(tp)
+            if (readResult != null && readResult.error == Errors.NONE) {
```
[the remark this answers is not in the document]
jeqo: The gate is already structurally enforced upstream — a partition can only reach the DelayedFetch consolidation check if disklessRemoteStorageConsolidationEnabled was true when the request was routed. The check in DelayedFetch is redundant by design.

core/src/main/scala/kafka/server/DelayedFetch.scala
```diff
@@ -235,50 +236,127 @@ class DelayedFetch(
...
+              replicaManager.getPartitionOrError(tp.topicPartition).foreach { partition =>
+                partition.log.foreach { log =>
+                  if (status.startOffsetMetadata.messageOffset < log.logEndOffset)
```
viktorsomogyi: So if I'm interpreting it right, this might not be enough. When a classic read happens on an older segment, that has less than minBytes remaining in the segment it forces completion instead of reading from the next segment. The current supplementation read I think breaks this and continues from log end instead *IF* I'm not missing something.
So suppose that a log is `[0,5000)`, segment 0 is `[0,2000)`, segment 1 is `[2000,4000)`, segment 2 is `[4000,5000)`. Now we have minBytes as 1MB and suppose we're reading at `1995`. Reading `[1995-2000)` may be 500KB so we would still need to read `[2000-2005)` from the next segment. In this case the broker would force completion and thus force the consumer to issue a next request (and "break" minBytes). The next classic fetch then continue from `2000`.
As I read it, in the classic case if there aren't enough data to return, then we force a diskless fetch from `[localLogEndOffset, fetchEnd)` which seems to be incorrect.
Let me know if I'm missing something.

jeqo: Good catch! I think I may unintentionally changed how classic partitions are handled. Also this helped me to catch how the transition from local logs to diskless had issues. Looking into how to fix this.

jeqo: I have added consolidating partitions map to DelayedFetch so classic (non-consolidating, non-diskless), diskless, and consolidating partitions are separated here and are easier to control. PTAL

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2257,9 +2339,55 @@ class ReplicaManager(val config: KafkaConfig,
...
+    // request will park in DelayedFetch regardless, where the supplement runs concurrently with the diskless fetch.
+    // Running it here as well would block for disklessFetchMaxWaitMs and then discard the result.
+    var consolidationSupplementData = Map.empty[TopicIdPartition, FetchPartitionData]
+    if (consolidatingLocalFetchSupplements.nonEmpty &&
+      disklessFetchInfos.isEmpty &&
+      !params.isFromFollower && // safeguard: followers must not receive diskless records merged into local-log data
+      params.maxWaitMs > 0 && // safeguard: non-blocking polls must not be held by a diskless round-trip
+      !hasPreferredReadReplica &&
+      bytesReadable < params.minBytes &&
+      !errorReadingData) {
```
viktorsomogyi, line 2366: I think this path may have the same problem because we're not checking here if the local data has been exhausted, just that `bytesReadable < params.minBytes` and then switch to diskless.

viktorsomogyi (commented):
I think if we can resolve that remaining issue (if you can confirm it's an issue) then we should be good to go with this.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: This should be solved by `buildConsolidationSupplementFetchInfos` fix should solve that by taking `lastBatch().nextOffset()` as the supplement start offset. https://github.com/aiven/inkless/pull/638/changes/aa6be120edf0d5dd729dc9b8688a70c0d25c80aa#diff-78812e247ffeae6f8c49b1b22506434701b1e1bafe7f92ef8f8708059e292bf0R2023-R2031

`testBuildConsolidationSupplementFetchInfosStartsAtLastBatchNextOffset` should also validate this scenario.

viktorsomogyi: Ok, your change indeed seems to cover the gap that I addressed in the previous review but I discovered a different gap this time (well it consists of 2 parts actually) and I had to dig a bit deeper to find it.

I generated a test case (should be in ReplicaManagerInklessTest) that showcases what I'm thinking of. So essentially when the diskless part doesn't cover the whole local range, then there seems to be a gap still. So let's take that the local range is `[0,1000)` but the segments end at every 500th offset, so 500, 1000 and so on. And the seal offset (classicToDisklessStartOffset) is at 1000 because the topic was just freshly switched. Then reading 480 for instance as in the test and falling below minBytes will result in a supplement read from offset 1000. I think we should somewhere check if we're still below the diskless range before supplementing (`supplementStartOffset >= seal`).

As for the second part, you can see that the read from diskless is mocked to 1000 specifically, but if you take a look at the PG control plane, it never checks the lower bound, so it never returns an error. Therefore the first part never gets exposed. Also, since 500 passes the checks in the find_batches routine, it returns the first record without a second thought. So this is partly on the diskless control plane because it never expects that its logs don't start at 0 AND you can fetch something higher than it but still below its log start.

[This](https://github.com/aiven/inkless/blob/02395b0b0c517c66b40bd6bae374743714594de9/storage/inkless/src/main/resources/db/migration/V8__Find_batches_function_with_limit.sql#L24) is the first check where we should really check log_start_offset instead of 0.

And [this](https://github.com/aiven/inkless/blob/02395b0b0c517c66b40bd6bae374743714594de9/storage/inkless/src/main/resources/db/migration/V8__Find_batches_function_with_limit.sql#L48) is the second one where it slips through and returns the 1st batch.

```
  // REPRODUCES the "sub-seal gap" silent-data-loss bug.
  //
  // Setup of a consolidating partition:
  //   - classic-to-diskless seal  = 1000  (offsets [1000, ...) live in object storage / diskless)
  //   - local classic prefix      = [0, 1000) and spans multiple local log segments
  //   - local logEndOffset (LEO)  = 1000
  //   - localLogStartOffset       = 0     (nothing tiered, so the tiered-range guard is a no-op)
  //
  // A consumer fetches at offset 480, which lands in an *earlier* local segment. LocalLog.read
  // serves from a single segment, so the local read returns only [480, 500) (a segment boundary
  // well below the seal). Because that is smaller than minBytes, the consolidating path triggers
  // a diskless supplement, computed by buildConsolidationSupplementFetchInfos as
  // lastBatch().nextOffset() == 500.
  //
  // But 500 is BELOW the seal (1000): the diskless control plane holds no batches for offsets
  // below the seal, so a FindBatch at 500 returns the first available batch — at the seal, 1000.
  // (The fixed FetchHandler stub returns records starting at 1000 regardless of the requested
  // offset, faithfully modelling that tailSet/"first batch at-or-after" control-plane behavior.)
  //
  // The merge therefore stitches local [480, 500) directly onto diskless [1000, ...), and the
  // committed range [500, 1000) is silently dropped from the consumer-visible response.
  @Test
  def testFetchConsolidatingSupplementBelowSealSilentlySkipsClassicPrefix(): Unit = {
    val seal = 1000L
    val localLeo = 1000L

    // Local single-segment read: offsets [480, 500). 20 records in one batch -> nextOffset == 500.
    val localRecordList = (0 until 20).map(i => new SimpleRecord(i.toLong, s"local-$i".getBytes())).toArray
    val localMemRecords = MemoryRecords.withRecords(
      2.toByte, 480L, Compression.NONE, TimestampType.CREATE_TIME, 123L, 0.toShort, 0, 0, false, localRecordList: _*
    )
    var localFileRecords: FileRecords = null

    // Diskless serves from the seal: offsets [1000, 1005). Models the control plane returning the
    // first batch at-or-after the requested (sub-seal) offset of 500.
    val supplementRecordList = (0 until 5).map(i => new SimpleRecord(i.toLong, s"diskless-$i".getBytes())).toArray
    val supplementRecords = MemoryRecords.withRecords(
      2.toByte, seal, Compression.NONE, TimestampType.CREATE_TIME, 456L, 0.toShort, 0, 0, false, supplementRecordList: _*
    )
    val disklessResponse = Map(disklessTopicPartition ->
      new FetchPartitionData(
        Errors.NONE,
        localLeo, 0L,
        supplementRecords,
        Optional.empty(), OptionalLong.of(localLeo), Optional.empty(), OptionalInt.empty(), false)
    )

    val fetchHandlerCtor = mockFetchHandler(disklessResponse)
    val cp = mock(classOf[ControlPlane])
    val replicaManager = spy(createReplicaManager(
      List(disklessTopicPartition.topic()),
      controlPlane = Some(cp),
      disklessManagedReplicasEnabled = true,
      disklessRemoteStorageConsolidationEnabled = true,
      consolidatingDisklessTopics = Set(disklessTopicPartition.topic()),
    ))
    try {
      when(replicaManager.inklessMetadataView().getClassicToDisklessStartOffset(disklessTopicPartition.topicPartition()))
        .thenReturn(seal)
      stubConsolidatingPartitionWithLocalLeo(replicaManager, localLeo = localLeo)

      localFileRecords = memoryRecordsToFileRecords(localMemRecords)
      doReturn(Seq(disklessTopicPartition ->
        new LogReadResult(
          new FetchDataInfo(new LogOffsetMetadata(480L, 0L, 0), localFileRecords),
          Optional.empty(), localLeo, 0L, localLeo, 0L, 0L, OptionalLong.empty(), Errors.NONE
        ))
      ).when(replicaManager).readFromLog(any(), any(), any(), any())

      val fetchParams = new FetchParams(
        FetchRequest.ORDINARY_CONSUMER_ID, -1L,
        1000L,
        localMemRecords.sizeInBytes + 1, // minBytes > local-only size to trigger the supplement
        1024 * 1024,
        FetchIsolation.HIGH_WATERMARK, Optional.empty()
      )
      val fetchInfos = Seq(
        disklessTopicPartition -> new PartitionData(disklessTopicPartition.topicId(), 480L, 0L, 1024 * 1024, Optional.empty())
      )

      @volatile var responseData: Map[TopicIdPartition, FetchPartitionData] = null
      val responseCallback = (response: Seq[(TopicIdPartition, FetchPartitionData)]) => {
        responseData = response.toMap
      }
      replicaManager.fetchMessages(fetchParams, fetchInfos, QuotaFactory.UNBOUNDED_QUOTA, responseCallback)

      waitForFetchResponse(responseData)
      val result = responseData(disklessTopicPartition)
      assertEquals(Errors.NONE, result.error)

      val offsets = result.records.records().asScala.map(_.offset()).toList
      val gaps = offsets.sliding(2).collect { case Seq(a, b) if b != a + 1 => (a, b) }.toList
      assertTrue(gaps.isEmpty,
        s"Consumer-visible records must be contiguous, but found offset gap(s): $gaps. " +
          s"The supplement started at offset 500 (below the seal $seal); diskless served from the " +
          s"seal, so the committed range [500, $seal) was silently skipped. Offsets=$offsets")
    } finally {
      replicaManager.shutdown(checkpointHW = false)
      fetchHandlerCtor.close()
      if (localFileRecords != null) localFileRecords.close()
    }
  }
  ```

viktorsomogyi (dismissed):
@jeqo although I think I found some further gaps, I approve this so it doesn't block you in the benchmarks and further work. Those gaps if you verified it, can be addressed separately.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Ah, good catch! I wasn't focusing on topics switched, so I was missing this scenario.
This is very relevant actually, so I work on the fix, and have a follow up for the PG validation.

viktorsomogyi:
Thanks for the fix. From my side it's good, but please check those copilot comments if they're relevant or if we wanted to fix them.

jeqo:
@viktorsomogyi thanks! I think all copilot suggestions are resolved now 👍🏽

viktorsomogyi (approved):

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Follow-up for control-plane validation: https://github.com/aiven/inkless/pull/666 👿

PR #640, opened 2026-06-12
fix(inkless:metrics): exclude consolidating partitions from URP metrics

jeqo:
- Exclude consolidating diskless partitions from `underReplicatedPartitionCount`, `atMinIsrPartitionCount`, and `underMinIsrPartitionCount` gauges
- These partitions never materialize followers (ReplicaManager skips `getOrCreatePartition` for them), so they always trip URP-related alerts despite being healthy from a consolidation standpoint
- Consolidation health is already observable via dedicated `ConsolidationLocalLag`/`TotalLag` metrics

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,5 @@ __pycache__
 _data/
 
 .inkless-sync/
+
+core/data/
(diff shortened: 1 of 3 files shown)


jeqo:
~Drafting as I see under-min-isr as well~ Done

viktorsomogyi (approved):

PR #642, opened 2026-06-12
fix(inkless): reject topic creation with config conflicting with diskless regex

gqmelo:
If a create topic request includes diskless.enable=false, but the topic name matches one of the diskless.force.include.topic.regexes, it is unclear what the real intent is.

Since this is mostly likely an user mistake, let's be stricter for now and reject the request to let the user know the topic would likely be created in a different way than they expected.

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/DisklessForceCreateTopicInterceptor.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/DisklessForceCreateTopicInterceptor.java
@@ -19,6 +19,7 @@ package org.apache.kafka.controller;
 
 import org.apache.kafka.clients.admin.AlterConfigOp.OpType;
 import org.apache.kafka.common.config.TopicConfig;
+import org.apache.kafka.common.errors.InvalidRequestException;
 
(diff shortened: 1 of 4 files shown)


tvainika:
Original requirement for #614 was
"Must be configurable by the user, which provides an allow list implemented through a list of regexes: if the topic being created matches at least one of the regexes, then **diskless.enable=true is added to the requested configs (even if the request contained diskless.enable=false)**" see https://aiven.atlassian.net/browse/KC-21

So this change goes against that requirement.

gqmelo:
> Original requirement for #614 was "Must be configurable by the user, which provides an allow list implemented through a list of regexes: if the topic being created matches at least one of the regexes, then **diskless.enable=true is added to the requested configs (even if the request contained diskless.enable=false)**" see https://aiven.atlassian.net/browse/KC-21
> 
> So this change goes against that requirement.

Yes, but I think it was written without being discussed. After some discussion, we opted for changing this behaviour.

tvainika (approved):
Thanks for the clarification. LGTM.

PR #643, opened 2026-06-12
fix(inkless:controller): fix leader skew for managed diskless after rolling restart

jeqo:
Three changes that together ensure balanced leadership for managed diskless topics:

1. Enable preferred leader rebalance: the previous blanket skip for all diskless topics was correct for unmanaged (RF=1) but wrong for managed (RF>1) where tiered storage upload/deletion are leader-only.

2. ISR = all replicas at creation/reassignment/addPartitions: matches unmanaged diskless semantics. Data is in object storage — broker fencing doesn't affect availability.

3. Expand ISR on broker unfence: when a broker returns, re-add it to ISR for diskless managed partitions where it is a replica. This repairs stale ISR from prior shrinks and enables the preferred leader election to redistribute leadership.

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
@@ -1891,14 +1891,25 @@ public class ReplicationControlManager {
                 PartitionRegistration partition = entry.getValue();
                 if (!Replicas.contains(partition.replicas, brokerId)) continue;
                 if (Replicas.contains(partition.isr, brokerId)) continue;
-                // Broker is a replica but not in ISR — expand ISR
-                int[] newIsr = Replicas.copyWith(partition.isr, brokerId);
-                PartitionChangeRecord changeRecord = new PartitionChangeRecord()
-                    .setTopicId(topic.id)
-                    .setPartitionId(partitionId)
-                    .setIsr(Replicas.toList(newIsr));
-                records.add(new ApiMessageAndVersion(changeRecord, (short) 0));
-                expanded++;
+                // Use PartitionChangeBuilder so that ELR is reconciled alongside the ISR
+                // expansion — a raw PartitionChangeRecord setting only `isr` would leave any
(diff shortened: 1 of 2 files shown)


jeqo:
Looking good: 
image

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -1873,6 +1874,37 @@ void handleBrokerUnfenced(int brokerId, long brokerEpoch, List<ApiMessageAndVers
...
+                int partitionId = entry.getKey();
+                PartitionRegistration partition = entry.getValue();
+                if (!Replicas.contains(partition.replicas, brokerId)) continue;
+                if (Replicas.contains(partition.isr, brokerId)) continue;
+                // Broker is a replica but not in ISR — expand ISR
+                int[] newIsr = Replicas.copyWith(partition.isr, brokerId);
+                PartitionChangeRecord changeRecord = new PartitionChangeRecord()
+                    .setTopicId(topic.id)
+                    .setPartitionId(partitionId)
+                    .setIsr(Replicas.toList(newIsr));
```
AnatolyPopov: This sets ISR via a raw PartitionChangeRecord instead of PartitionChangeBuilder, so ELR is never reconciled. If ELR is enabled (and it is default for 4.1+) and ISR previously shrank below min.insync.replicas, the returning broker is in elr. So this adds it back to ISR without removing it from ELR or clearing ELR once ISR ≥ minISR so it violates one of the invariants from ELR KIP that says that ELR can't be ISR.

This might be a bit irrelevant for diskless at all but I wonder if we should maintain that.

jeqo: Good catch! Even though the scenario is unlikely, the invariant violation is real. I'll switch to `PartitionChangeBuilder` to get proper ELR cleanup when ISR expands back.

AnatolyPopov (approved):
LGTM

PR #647, opened 2026-06-15
feat(inkless): ensure unclean leader election is disabled on classic-to-diskless switch [KC-129]

EelisK:


metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -3008,6 +3009,12 @@ ApiError validateClassicToDisklessSwitchPrecondition(
         if (topicId == null) return ApiError.NONE;
         TopicControlInfo topicInfo = topics.get(topicId);
         if (topicInfo == null) return ApiError.NONE;
+        if (configurationControl.uncleanLeaderElectionEnabledForTopic(resource.name())
+            || isSettingConfigToTrue(changes, UNCLEAN_LEADER_ELECTION_ENABLE_CONFIG)) {
+            return new ApiError(INVALID_CONFIG,
+                "Cannot switch topic " + resource.name() + " to diskless: " +
+                "unclean leader election must be disabled.");
+        }
```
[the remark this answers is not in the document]
EelisK: Addressed

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -3035,22 +3042,15 @@ ApiError validateClassicToDisklessSwitchPreconditionForLegacy(
...
-            validatePartitionsForSwitch(topicInfo)
-                .ifPresent(error -> errors.put(resource, error));
+        boolean uncleanExplicit = Boolean.parseBoolean(configs.get(UNCLEAN_LEADER_ELECTION_ENABLE_CONFIG));
+        boolean uncleanImplicit = !configs.containsKey(UNCLEAN_LEADER_ELECTION_ENABLE_CONFIG)
+                && configurationControl.uncleanLeaderElectionEnabledForTopic(resource.name());
+        if (uncleanExplicit || uncleanImplicit) {
+            return new ApiError(INVALID_CONFIG,
+                "Cannot switch topic " + resource.name() + " to diskless: " +
+                "unclean leader election must be disabled.");
         }
```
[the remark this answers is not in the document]
EelisK: Addressed

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -3002,12 +3008,25 @@ ApiError validateClassicToDisklessSwitchPrecondition(
...
         TopicControlInfo topicInfo = topics.get(topicId);
         if (topicInfo == null) return ApiError.NONE;
+
+        if (isSettingConfigToTrue(changes, UNCLEAN_LEADER_ELECTION_ENABLE_CONFIG) && 
+                hasAnyPartitionWithPendingSwitch(topicInfo)) {
+            return new ApiError(INVALID_CONFIG,
+                "Cannot enable unclean leader election for topic " + resource.name() +
+                ": topic has a pending classic-to-diskless switch.");
+        }
+        if (!isSettingConfigToTrue(changes, DISKLESS_ENABLE_CONFIG)) return ApiError.NONE;
```
giuseppelillo: Why was this moved later? Can we move it back up so it returns earlier for these cases?

core/src/test/java/kafka/server/InklessTopicTypeSwitcherClusterTest.java
```diff
@@ -436,6 +436,62 @@ public void testSwitchRejectedWhenPartitionIsOfflineOrUnderReplicated() throws E
...
+            // is enabled (via cluster-level config)
+            final ExecutionException ex = assertThrows(ExecutionException.class, () ->
+                    alterTopicConfigWithIncrementalAlterConfigs(admin, topic, Map.of(
+                            TopicConfig.DISKLESS_ENABLE_CONFIG, "true")));
+            assertInstanceOf(InvalidConfigurationException.class, ex.getCause());
+            assertTrue(ex.getCause().getMessage().contains("unclean leader election"),
+                    "Expected 'unclean leader election' in error, got: " + ex.getCause().getMessage());
+        }
+    }
+
```
giuseppelillo, line 475:
```suggestion
```
nit: remove double new line

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
giuseppelillo: This does not guard against the case where the topic is already diskless but the switch is still pending. The incremental alter configs logic blocks this instead.

metadata/src/test/java/org/apache/kafka/controller/ReplicationControlManagerTest.java
```diff
@@ -7139,6 +7090,180 @@ public void testElectLeadersRejectsUncleanElectionForPendingSwitchPartition() {
                 "Expected pending switch message in: " + partitionResult.errorMessage());
         }
 
+        @Test
+        public void testSwitchRejectedWhenUncleanLeaderElectionAlreadyEnabled() {
```
giuseppelillo, line 7094: Can you also add a test that verifies this: reject a classic→diskless switch when unclean leader election would be effectively enabled (incremental + legacy paths) ?

E.g. mark pending via markClassicToDisklessSwitchStarted + replay, then call validateClassicToDisklessSwitchPrecondition with unclean=SET=true and assert rejection

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -3002,12 +3008,25 @@ ApiError validateClassicToDisklessSwitchPrecondition(
     ) {
```
giuseppelillo, line 3007: I believe this check is missing: an incremental DELETE of an unclean=false override on a pending-switch topic, where the cluster default is true. This is now allowed but it should not be.

giuseppelillo (changes requested):

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
EelisK: Addressed

core/src/test/java/kafka/server/InklessTopicTypeSwitcherClusterTest.java
```diff
@@ -436,6 +436,61 @@ public void testSwitchRejectedWhenPartitionIsOfflineOrUnderReplicated() throws E
...
+            // Create a classic topic
+            admin.createTopics(List.of(
+                    new NewTopic(topic, 1, (short) 3)
+                            .configs(Map.of(TopicConfig.DISKLESS_ENABLE_CONFIG, "false"))
+            )).all().get(30, TimeUnit.SECONDS);
+
+            // Attempt to switch to diskless — should fail because unclean leader election
+            // is enabled (via cluster-level config)
+            final ExecutionException ex = assertThrows(ExecutionException.class, () ->
+                    alterTopicConfigWithIncrementalAlterConfigs(admin, topic, Map.of(
```
giuseppelillo, line 462: Can you also do the same check, in the same test, but using `alterTopicConfigWithLegacyAlterConfigs`?

core/src/test/java/kafka/server/InklessTopicTypeSwitcherClusterTest.java
```diff
@@ -436,6 +436,61 @@ public void testSwitchRejectedWhenPartitionIsOfflineOrUnderReplicated() throws E
...
+                            .configs(Map.of(TopicConfig.DISKLESS_ENABLE_CONFIG, "false"))
+            )).all().get(30, TimeUnit.SECONDS);
+
+            // Enable diskless to mark partition as switch pending
+            alterTopicConfigWithIncrementalAlterConfigs(
+                    admin, topic, Map.of(TopicConfig.DISKLESS_ENABLE_CONFIG, "true"));
+
+            // Should fail trying to enable unclean leader election
+            final ExecutionException ex = assertThrows(ExecutionException.class, () ->
+                    alterTopicConfigWithIncrementalAlterConfigs(admin, topic, Map.of(
```
giuseppelillo, line 493: Same here, can we double this check using `alterTopicConfigWithLegacyAlterConfigs` ?

giuseppelillo (commented):
LGTM, just missing a small check from my point of view

core/src/test/java/kafka/server/InklessTopicTypeSwitcherClusterTest.java
EelisK: Added here https://github.com/aiven/inkless/pull/647/commits/5014c3472db5c1943732a63576fdeb971562ff69

giuseppelillo (approved):

PR #653, opened 2026-06-16
fix(inkless:consolidation): recover a switched consolidated leader after local-log loss

viktorsomogyi:
A leader that comes up below the seal with an empty local log (a full local-storage wipe / disaster recovery where the controller, whose metadata survived, elects a replica that came back empty) has no peer to replicate the classic prefix from -- that prefix lives only in the remote tier. The ConsolidationReconciler now starts consolidation for such a leader (when remote storage is enabled) instead of retrying forever, so the fetcher lands below the diskless WAL start, hits OFFSET_MOVED_TO_TIERED_STORAGE, and rebuilds the leader-epoch cache and producer snapshot from remote.

### DIFF
--- a/core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
@@ -67,17 +67,11 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
       val consolidatingPartitionAndOffsets: mutable.HashMap[TopicPartition, InitialFetchState] =
         initConsolidatingPartitionFetching(consolidatingPartitions)
 
-      // Mark topics throttled BEFORE starting fetchers: addFetcherForPartitions starts the
-      // fetcher threads immediately, and ReplicaFetcherThread only records bytes to the quota
-      // when the partition is already throttled. Marking after would let the first fetch/append
-      // bypass the quota. Unlike classic replication (where throttled replicas are set via topic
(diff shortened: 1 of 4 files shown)


core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -129,12 +143,31 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
...
+            // live leader, hence the Retry below), a leader has no peer source: the classic prefix
+            // [0, seal) lives only in the remote tier. Start consolidation anyway, armed at the
+            // current (post-loss) LEO, so the fetcher's first request lands below the diskless WAL
+            // start and DisklessLeaderEndPoint answers OFFSET_MOVED_TO_TIERED_STORAGE. That drives
+            // the stock tier-state machine to rebuild the leader-epoch cache + producer snapshot
+            // and the whole-log start from remote before resuming the WAL fetch; without this the
+            // partition would wait forever for a classic catch-up that can never happen.
+            val pruneFloor = math.max(seal, log.logStartOffset)
+            partition.ensureConsolidationPruneFloorAtLeast(pruneFloor)
+            ConsolidationStartState.Ready(log.logEndOffset)
```
AnatolyPopov: Nit: these 3 lines are identical to 3 lines in the else block below(lines 184-186), should these be unified?

viktorsomogyi: Done.

AnatolyPopov (approved):

PR #660, opened 2026-06-18
fix(inkless:switch): Fix stale HW on switched leader promotion

giuseppelillo:
Reconcile already-switched leaders after makeLeader so a promoted replica with a stale checkpointed HW advances to the committed seal. Truncate any local classic tail above the seal, and fence leaders whose local LEO is below the seal to avoid serving an incomplete prefix. Keep follower truncation limited to the seal-commit delta. Add a deterministic system test for the stale-HW promotion regression.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -3299,22 +3299,68 @@ class ReplicaManager(val config: KafkaConfig,
   }
 
   /**
-   * Reconciles a partition whose classic-to-diskless seal was *just committed* in this delta
-   * (previous seal < 0, now >= 0); other transitions are left untouched. Records beyond the seal
-   * are uncommitted (the diskless log continues from the seal), so:
-   *  - LEO > seal: truncate down to the seal (a follower/deposed leader that held un-replicated
-   *    records past a seal a different broker committed at a lower HW);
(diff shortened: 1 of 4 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3299,22 +3299,68 @@ class ReplicaManager(val config: KafkaConfig,
...
-                                                  delta: TopicsDelta): Unit = {
+  private def maybeReconcileSwitchedLeader(tp: TopicPartition,
+                                           topicId: Uuid,
+                                           newRegistration: PartitionRegistration): Unit = {
+    val seal = newRegistration.classicToDisklessStartOffset
+    if (seal < 0) return
+
+    onlinePartition(tp).foreach { partition =>
+      try {
+        val log = partition.localLogOrException
```
viktorsomogyi: An unexpected exception here would abort the reconciliation for all following partitions, perhaps adding a `NotLeaderOrFollowerException` to the catch or just using partition.log with a pattern match would make it more defensive.

viktorsomogyi (commented):
just one nit, otherwise good.

core/src/main/scala/kafka/server/ReplicaManager.scala
giuseppelillo: Done, added a pattern match to skip reconciliation for only the partitions that do not have the local log.

viktorsomogyi (approved):

PR #661, opened 2026-06-22
feat(inkless:tools): add operator tooling for topic type switch [KC-97]

EelisK:
This change introduces the necessary scaffolding for topic type switch operator tooling, as well as the first or three commands that will be available via `bin/kafka-topic-switch.sh`. We implement the `state` command, which outputs per-partition switch state, offsets, and producer state. Furthermore, a new admin method `describeTopicPartitions` is added to expose unknown tagged fields, which contains the switch state.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicPartitionsResult.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicPartitionsResult.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.clients.admin;
+
+import org.apache.kafka.common.KafkaFuture;
+import org.apache.kafka.common.message.DescribeTopicPartitionsResponseData;
+
+/**
+ * Raw protocol result for {@link Admin#describeTopicPartitions(java.util.Collection, DescribeTopicsOptions)}.
+ * This exposes the raw protocol response data directly rather than a typed model.
+ * The returned data is mutable and may contain embedded errors that callers must inspect.
+ */
+public class DescribeTopicPartitionsResult {
+
+    private final KafkaFuture<DescribeTopicPartitionsResponseData> future;
+
+    public DescribeTopicPartitionsResult(KafkaFuture<DescribeTopicPartitionsResponseData> future) {
+        this.future = future;
+    }
+
+    /**
+     * Returns the raw {@link DescribeTopicPartitionsResponseData} accumulated across all pagination rounds.
+     * Topic-level and partition-level errors are checked during accumulation and will cause the future
+     * to complete exceptionally. The returned data is mutable.
+     */
+    public KafkaFuture<DescribeTopicPartitionsResponseData> rawResponse() {
+        return future;
+    }
+}
--- a/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
@@ -2454,6 +2454,84 @@ public class KafkaAdminClient extends AdminClient {
     }
 
     @Override
+    public DescribeTopicPartitionsResult describeTopicPartitions(Collection<String> topics, DescribeTopicsOptions options) {
+        if (topics.isEmpty()) {
+            throw new IllegalArgumentException("topics must not be empty");
+        }
+        final KafkaFutureImpl<DescribeTopicPartitionsResponseData> future = new KafkaFutureImpl<>();
+        final DescribeTopicPartitionsResponseData accumulated = new DescribeTopicPartitionsResponseData();
+        final long now = time.milliseconds();
+        final List<String> topicNamesList = new ArrayList<>(topics);
+        topicNamesList.sort(String::compareTo);
+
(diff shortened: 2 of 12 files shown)


tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
```diff
@@ -0,0 +1,217 @@
...
+                throw new RuntimeException("Unknown command " + command);
+        }
+    }
+
+    private static ArgumentParser argumentParser() {
+        ArgumentParser parser = ArgumentParsers
+                .newArgumentParser("kafka-cluster")
+                .defaultHelp(true)
+                .description("The Kafka cluster tool.");
+        Subparsers commandParsers = parser.addSubparsers().dest("command");
```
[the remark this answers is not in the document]
EelisK: Oops

tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
+
+import static net.sourceforge.argparse4j.impl.Arguments.store;
+import static net.sourceforge.argparse4j.impl.Arguments.storeTrue;
+
+public abstract class TopicSwitchCommand {
```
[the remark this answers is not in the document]
EelisK: Having it abstract is consistent with other internal tooling like `TopicCommand`

tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
```diff
@@ -0,0 +1,217 @@
...
+            stream.printf("  leader=%d epoch=%d isr=%s%n",
+                partition.leaderId(), partition.leaderEpoch(), partition.isrNodes());
+            stream.printf("  classicToDisklessStartOffset=%s%n",
+                formatStartOffset(classicToDisklessStartOffset));
+            stream.printf("  disklessLeaderEpoch=%s%n",
+                formatDisklessLeaderEpoch(disklessLeaderEpoch));
+            stream.printf("  logStartOffset=%d highWatermark=%d logEndOffset=%d%n",
+                earliestInfo != null ? earliestInfo.offset() : -1,
+                latestInfo != null ? latestInfo.offset() : -1,
+                latestInfo != null ? latestInfo.offset() : -1);
```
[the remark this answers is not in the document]
EelisK: D'oh!

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -2453,6 +2453,67 @@ private Node leader(PartitionInfo partitionInfo) {
         return partitionInfo.leader();
     }
 
+    @Override
+    public DescribeTopicPartitionsResult describeTopicPartitions(Collection<String> topics, DescribeTopicsOptions options) {
```
giuseppelillo, line 2457: We need some tests for this new method

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -2453,6 +2453,67 @@ private Node leader(PartitionInfo partitionInfo) {
...
+                DescribeTopicPartitionsResponseData data = response.data();
+
+                for (DescribeTopicPartitionsResponseTopic topic : data.topics()) {
+                    DescribeTopicPartitionsResponseTopic existing = accumulated.topics().find(topic.name());
+                    if (existing != null) {
+                        existing.partitions().addAll(topic.partitions());
+                    } else {
+                        accumulated.topics().add(topic.duplicate());
+                    }
+                }
```
giuseppelillo, line 2515:
```suggestion
                for (DescribeTopicPartitionsResponseTopic topic : data.topics()) {
                    Errors error = Errors.forCode(topic.errorCode());
                    if (error != Errors.NONE) {
                        future.completeExceptionally(error.exception());
                        return;
                    }
                    for (DescribeTopicPartitionsResponsePartition partition : topic.partitions()) {
                        Errors partitionError = Errors.forCode(partition.errorCode());
                        if (partitionError != Errors.NONE) {
                            future.completeExceptionally(partitionError.exception());
                            return;
                        }
                    }
                    DescribeTopicPartitionsResponseTopic existing = accumulated.topics().find(topic.name());
                    if (existing != null) {
                        if (!existing.topicId().equals(topic.topicId())) {
                            future.completeExceptionally(new UnknownTopicOrPartitionException(
                                "Topic " + topic.name() + " changed while fetching paginated DescribeTopicPartitions response"));
                            return;
                        }
                        existing.partitions().addAll(topic.partitions());
                    } else {
                        accumulated.topics().add(topic.duplicate());
                    }
                }
```

Need to check for errors here

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicPartitionsResult.java
```diff
@@ -0,0 +1,33 @@
...
+    private final KafkaFuture<DescribeTopicPartitionsResponseData> future;
+
+    public DescribeTopicPartitionsResult(KafkaFuture<DescribeTopicPartitionsResponseData> future) {
+        this.future = future;
+    }
+
+    public KafkaFuture<DescribeTopicPartitionsResponseData> all() {
+        return future;
+    }
+}
```
giuseppelillo, line 43: Please avoid exposing DescribeTopicPartitionsResponseData as a normal public Admin result. Either introduce a typed Inkless-specific result model, or mark/rename this as an unstable raw protocol escape hatch and document that the returned data is mutable and contains embedded errors that callers must inspect.

Also, `DescribeTopicPartitionsResult.all()` is misleading. In Kafka admin APIs, all() usually means "complete successfully only if all requested resources succeeded". I would suggest to rename it to something like `rawResponse` to make it clear that it's receiving the raw data without processing.

giuseppelillo (changes requested):

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
EelisK: Good catch, added some ✅

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicPartitionsResult.java
EelisK: Thanks for the explanation. Documented the model as an escape hatch, I think this is good enough for an operator tool. Also renamed `all` to `rawResponse`

giuseppelillo (approved):

PR #662, opened 2026-06-23
feat(inkless): allow diskless switch only when remote storage is enabled [KC-217]

EelisK:


EelisK:
Closing in favor of https://aiven.atlassian.net/browse/KC-293

PR #664, opened 2026-06-23
Backport to 4.1: KAFKA-19354: KRaft observer should send fetch to best node

giuseppelillo:
Bug: https://issues.apache.org/jira/browse/KAFKA-19354
Original fix: https://github.com/apache/kafka/commit/7339d65b27fd5c1aa5ba239faf193ee0ff55f06d

Observers may get stuck fetching from bootstrap servers even on discovery of a leader from a fetch response.

Observers currently fetch from bootstrap servers once their fetch timeout expires, and even on rediscovery of the leader (e.g. observer receives fetch response indicating location of leader), observers will not  attempt to resume fetching from the leader.

This change simplifies observer polling logic to just rely on maybeSendFetchToBestNode which takes care of the logic of selecting either the leader or bootstrap servers  as the destination based on timeouts and backoffs.

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/FollowerState.java
+++ b/raft/src/main/java/org/apache/kafka/raft/FollowerState.java
@@ -131,12 +131,18 @@ public class FollowerState implements EpochState {
         return fetchTimer.isExpired();
     }
 
+    /**
+     * Reset the fetch timeout after successful fetch from leader.
+     */
     public void resetFetchTimeoutForSuccessfulFetch(long currentTimeMs) {
-        fetchTimer.update(currentTimeMs);
-        fetchTimer.reset(fetchTimeoutMs);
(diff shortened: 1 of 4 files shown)


glenn-mcallister-aiven:
Just doing some due dilligence: this is identical to PR https://github.com/apache/kafka/pull/19854, correct?

EDIT: even just eyeballing the two PRs I can see they are the same.

glenn-mcallister-aiven (approved):
Looks to be a clean backport of upstream. Assuming CI passes, 👍

giuseppelillo:
> Just doing some due dilligence: this is identical to PR https://github.com/apache/kafka/pull/19854, correct?
> 
> 
> 
> EDIT: even just eyeballing the two PRs I can see they are the same. 

Yes, it's also highlighted in the PR description

PR #665, opened 2026-06-25
feat(inkless:switch): implement AlterDisklessSwitch and tooling [KC-97]

EelisK:


clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4882,6 +4885,43 @@ void handleFailure(Throwable throwable) {
...
+
+            @Override
+            void handleFailure(Throwable throwable) {
+                future.completeExceptionally(throwable);
+            }
+        };
+        runnable.call(call, now);
+        return new AlterDisklessSwitchResult(future);
+    }
+
```
giuseppelillo, line 4939: `alterDisklessSwitch` updates only classicToDisklessStartOffset, and for sealOffset >= 0 writes a new disklessLeaderEpoch. For sealOffset == -1 or -2, though, it leaves any existing `disklessProducerStates` and `disklessLeaderEpoch` untouched.

That seems inconsistent with the intended state model: -1 means not switched/aborted, and -2 means pending/re-armed. In both states, producer states and disklessLeaderEpoch should be chosen again by the next initDisklessLog flow, not carried over from the previous completed switch. Otherwise the metadata can contain stale switch state even after an abort or re-arm.

I think this path should explicitly clear dependent switch metadata when sealOffset < 0, probably by making `PartitionRegistration.merge()` enforce the invariant: if an explicit `classicToDisklessStartOffset` tag is present and negative, reset `disklessProducerStates` to empty and `disklessLeaderEpoch` to `NO_DISKLESS_LEADER_EPOCH`.

giuseppelillo: As a follow up, in case of sealOffset >= 0, we should also give the possibility to the client to reset the `disklessProducerStates` or leave them as they were. By default i would leave them as they are.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4882,6 +4885,43 @@ void handleFailure(Throwable throwable) {
         return new UnregisterBrokerResult(future);
     }
 
+    @Override
+    public AlterDisklessSwitchResult alterDisklessSwitch(String topic, int partition, long sealOffset,
+                                                         AlterDisklessSwitchOptions options) {
```
giuseppelillo, line 4904: Should we add some validation here? For example, abort the request if `disklessStartOffset` was never set at all, and if `diskless.enable` is not set to true this should not work at all, as it's a recovery tool and not a tool to switch to diskless.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
```diff
@@ -1819,6 +1821,40 @@ public void testDescribeTopicPartitionsRawResponse() throws ExecutionException,
...
+            MessageUtil.toByteBufferAccessor(responseData, version), version);
+
+        List<RawTaggedField> taggedFields = roundTrippedData.topics().find("test-topic")
+            .partitions().get(0).unknownTaggedFields();
+        assertEquals(1, taggedFields.size());
+        RawTaggedField roundTripped = taggedFields.get(0);
+        assertEquals(classicToDisklessStartOffsetTag, roundTripped.tag());
+        assertEquals(42L, ByteBuffer.wrap(roundTripped.data()).getLong());
+    }
+
```
giuseppelillo, line 1851: This test does not seem to exercise the fix. It only round-trips `DescribeTopicPartitionsResponseData` through the protocol serializer/deserializer, which already preserves unknown tagged fields before this change.

The issue is in `KafkaAdminClient.describeTopicPartitions()`, so the test should exercise the admin client path instead.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
giuseppelillo: Moreover, in case the switch is aborted, `diskless.enable` should also be reset to false.

giuseppelillo (changes requested):
The approach looks good. Some comments to refine validations and values being set.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
EelisK: Setting `diskless.enable=false` here: https://github.com/aiven/inkless/pull/665/changes/dca2f39411321f20f961b2045cdf3c236c4e634a. Additional validation amended here: https://github.com/aiven/inkless/pull/665/changes/eb705f1fea42ef76aec9dbac661aacfba35b8a8e

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
EelisK: Addressed ✅ 
Allow clearing producer states when re-sealing: https://github.com/aiven/inkless/pull/665/changes/36aef740b1202601929ecab8571f7c631225f1d1

tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
```diff
@@ -196,6 +222,67 @@ static void stateCommand(PrintStream stream, Admin adminClient, String topic) th
...
+                    "%s-%d already has a committed seal offset (%d); pass an explicit --offset to re-seal.",
+                    topic, partition, committedSeal));
+            }
+            TopicPartition tp = new TopicPartition(topic, partition);
+            sealOffset = adminClient.listOffsets(Map.of(tp, OffsetSpec.latest()))
+                .all().get().get(tp).offset();
+            stream.printf("Validated %s-%d: sealing at end offset %d.%n", topic, partition, sealOffset);
+        } else {
+            sealOffset = offset.get();
+        }
```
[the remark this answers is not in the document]
giuseppelillo: This is a real issue

core/src/main/scala/kafka/server/ControllerApis.scala
```diff
@@ -688,6 +689,21 @@ class ControllerApis(
...
+    val context = new ControllerRequestContext(request.context.header.data, request.context.principal,
+      OptionalLong.empty())
+    controller.alterDisklessSwitch(context, alterDisklessSwitchRequest.data).handle[Unit] { (result, e) =>
+      val response = if (e != null) {
+        alterDisklessSwitchRequest.getErrorResponse(e)
+      } else {
+        new AlterDisklessSwitchResponse(result)
+      }
+      requestHelper.sendResponseMaybeThrottle(request, _ => response)
+    }
```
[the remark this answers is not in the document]
giuseppelillo: minor issue but still i would fix it

tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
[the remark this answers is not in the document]
giuseppelillo: We might also check if the new sealOffset is behind the log start offset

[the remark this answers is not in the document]
EelisK: Fixed here https://github.com/aiven/inkless/pull/665/changes/f353718e35faca24dfb285fcbaa6ea4d29c50e43

core/src/main/scala/kafka/server/ControllerApis.scala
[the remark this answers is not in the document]
EelisK: ✅ https://github.com/aiven/inkless/pull/665/commits/e921090db885f6b8a296d906b55d01389f593346

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
giuseppelillo: Given the new changes of this PR, this validation now needs to exactly check if `partition.classicToDisklessStartOffset` is equal to `PartitionRegistration.CLASSIC_TO_DISKLESS_SWITCH_PENDING`, as it's the only possible state in which `initDisklessLog` can be safely completed.

giuseppelillo (changes requested):
As discussed, the approach looks good.
Changing diskless.enable to false when aborting the switch will be implemented in a separate command.
Control plane update for changing the diskless start offset to a different value is also a separate command.

Just one comment left and we're good to go.

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
EelisK: ✅ https://github.com/aiven/inkless/pull/665/commits/1197381463a44b49839006aeb1dec8d77044ab3b

giuseppelillo (approved):

PR #667, opened 2026-06-25
fix(inkless): truncate above the seal only if those messages are not consolidated

giuseppelillo:
Check if local log needs to be truncated after a diskless switch based on the Leader Epoch: if consolidation is disabled or messages above the log end offset are written with the leader epoch `E_d`, those are consolidated records that do not need truncation. If leader epoch is not `E_d`, truncate the local log up to the seal.


AnatolyPopov (approved):
LGTM, thanks for fixing this quickly. Will wait before merging for @viktorsomogyi's review just in case I missed anything.

viktorsomogyi (approved):

PR #668, opened 2026-06-26
fix(inkless): Add PostgreSQL socket timeouts for Inkless control plane

giuseppelillo:
Configure bounded Hikari and pgjdbc timeouts for control-plane connections. This prevents stale PostgreSQL endpoints from blocking diskless produce commits indefinitely. Add a failover-style integration test covering stale DNS and recovery.

### DIFF
--- a/docs/inkless/configs.rst
+++ b/docs/inkless/configs.rst
@@ -271,11 +271,35 @@ Under ``inkless.control.plane.``
   * Default: null
   * Importance: high
 
+``connection.pool.timeout.ms``
+  Maximum time in milliseconds to wait for a PostgreSQL connection from the pool.
+
+  * Type: long
+  * Default: 5000 (5 seconds)
+  * Valid Values: [250,...]
+  * Importance: medium
+
 ``max.connections``
   Maximum number of connections to the database
 
   * Type: int
   * Default: 10
+  * Valid Values: [1,...]
+  * Importance: medium
+
+``socket.timeout.ms``
+  Maximum time in milliseconds to wait for PostgreSQL socket reads.
+
+  * Type: int
+  * Default: 5000 (5 seconds)
+  * Valid Values: [1,...]
+  * Importance: medium
+
+``tcp.connect.timeout.ms``
+  Maximum time in milliseconds to establish a PostgreSQL TCP connection.
+
+  * Type: int
+  * Default: 5000 (5 seconds)
   * Valid Values: [1,...]
   * Importance: medium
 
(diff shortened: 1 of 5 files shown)


storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/PostgresControlPlane.java
```diff
@@ -136,12 +137,21 @@ private static HikariConfig dataSourceConfig(final KafkaMetricsGroup metrics, fi
...
+        config.addDataSourceProperty("socketTimeout", timeoutSeconds(connectionConfig.socketTimeoutMs()));
+        config.addDataSourceProperty("loginTimeout", timeoutSeconds(connectionConfig.connectTimeoutMs()));
+        config.addDataSourceProperty("tcpKeepAlive", "true");
 
         // We're doing interactive transactions.
         config.setAutoCommit(false);
         return config;
     }
 
+    private static long timeoutSeconds(final int timeoutMs) {
```
giuseppelillo: Round up

jlprat: This doesn't make that much sense, unless you start passing negative numbers in there. any number + 999 will be bigger than 1 (1000 adjusting for units)

jlprat: Or are you only worried about the 0 case only?

giuseppelillo: yeah we need to guard it against 0 because it will disable it. Changing it

docs/inkless/configs.rst
```diff
@@ -271,6 +271,22 @@ Under ``inkless.control.plane.``
...
 
+``connect.timeout.ms``
+  Maximum time in milliseconds to establish a PostgreSQL TCP connection.
+
+  * Type: int
+  * Default: 5000 (5 seconds)
+  * Valid Values: [1,...]
+  * Importance: medium
+
+``connection.timeout.ms``
```
jlprat: these 2 will be confused quite often I guess. Are we using the same pattern elsewhere?

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/PostgresControlPlane.java
giuseppelillo: Removed the check about 0 because the config definition has a minimum of 1

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/PostgresControlPlane.java
```diff
@@ -136,12 +136,22 @@ private static HikariConfig dataSourceConfig(final KafkaMetricsGroup metrics, fi
...
+        // pgjdbc expects whole seconds, so round positive millisecond values up
+        return (timeoutMs + 999L) / 1000L;
```
jlprat: (0L + 999L) / 1000L would be floored by default, right?
Math.ceilDiv((0L + 999L), 1000L) might work better (https://docs.oracle.com/en/java/javase//19/docs/api/java.base/java/lang/Math.html#ceilDiv(long,long))

jlprat (commented):
I'm not convinced with the timeoutSeconds rounding up logic.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/PostgresControlPlane.java
```diff
@@ -136,12 +136,22 @@ private static HikariConfig dataSourceConfig(final KafkaMetricsGroup metrics, fi
...
+        config.addDataSourceProperty("tcpKeepAlive", "true");
 
         // We're doing interactive transactions.
         config.setAutoCommit(false);
         return config;
     }
 
+    private static long timeoutSeconds(final int timeoutMs) {
+        // pgjdbc expects whole seconds, so round millisecond config values up.
+        return Math.ceilDiv(timeoutMs, 1000L);
```
jlprat: `timeoutMS` needs to be a Long, right now it's an `int.

docs/inkless/configs.rst
giuseppelillo: changed to be more discernible

jlprat (approved):
LGTM

PR #671, opened 2026-06-28
feat(inkless): coalesce contiguous batches into one row in commit_file_v2

jeqo:
Reduce Postgres `batches`-table row growth for low-throughput producers by coalescing contiguous batch runs into a single row at commit time (`commit_file_v2`).

Each produced physical batch becomes one `batches` row (`commit_file_v1` inserts one row per request). For low-throughput / small-batch producers this makes the table grow with produce-request rate — independent of data volume or partition count — and it bleeds. Reducing partitions does not help; the only general lever is writing fewer rows per unit of ingested data.

`commit_file_v2` walks the commit requests exactly as v1 (same per-request producer-state validation and response) but, instead of one INSERT per accepted request, collapses a maximal run of **byte-contiguous, same
`(topic_id, partition, magic, timestamp_type)`, accepted** requests into a single `batches` row.

Adding PG control plane config `inkless.control.plane.batch.coalescing.enabled ` to gate this and make it safe to ship when all brokers support it. 

Measured effect:
- **Idempotency/EOS unchanged**: `producer_state` is still written per physical batch; the per-request response (assigned offset, log start, timestamp, error) and high-watermark progression are byte-identical to v1.
- **Run breaks** on partition / magic / timestamp-type change, or a rejected request — a rejected batch consumes file bytes but doesn't advance the HW, so it breaks byte-adjacency and its dead bytes are excluded. No explicit flush needed.
- **Read path** (`FetchCompleter`) relocates the K physical sub-batches inside a coalesced row's byte range from the row base offset — implementing the existing "concatenate multiple batches into one BatchInfo" TODO. Single-batch rows (all v1 / in-memory / cache / pre-upgrade rows) take the unchanged path.

Commit file function diff:

```diff
@@ -15,7 +15,23 @@
     l_assigned_offset offset_nullable_t;
     l_new_high_watermark offset_nullable_t;
     l_last_sequence_in_producer_epoch BIGINT;
+    l_log_append_timestamp BIGINT;
+
+    -- Open-run accumulator. A run is a maximal sequence of accepted, byte-contiguous
+    -- requests sharing (topic_id, partition, magic, timestamp_type).
+    l_run_open BOOLEAN := FALSE;
+    l_run_topic_id UUID;
+    l_run_partition INT;
+    l_run_magic SMALLINT;
+    l_run_timestamp_type SMALLINT;
+    l_run_base_offset BIGINT;
+    l_run_last_offset BIGINT;
+    l_run_byte_offset BIGINT;
+    l_run_byte_size BIGINT;
+    l_run_max_timestamp BIGINT;
 BEGIN
+    l_log_append_timestamp := (EXTRACT(EPOCH FROM arg_now AT TIME ZONE 'UTC') * 1000)::BIGINT;
+
     INSERT INTO files (object_key, format, reason, state, uploader_broker_id, committed_at, size)
     VALUES (arg_object_key, arg_format, 'produce', 'uploaded', arg_uploader_broker_id, arg_now, arg_file_size)
     RETURNING file_id
@@ -151,30 +167,49 @@

         l_log.high_watermark = l_new_high_watermark;

-        INSERT INTO batches (
-            magic,
-            topic_id, partition,
-            base_offset,
-            last_offset,
-            file_id,
-            byte_offset, byte_size,
-            timestamp_type, log_append_timestamp, batch_max_timestamp
-        )
-        VALUES (
-            l_request.magic,
-            l_request.topic_id, l_request.partition,
-            l_assigned_offset,
-            l_new_high_watermark - 1,
-            l_new_file_id,
-            l_request.byte_offset, l_request.byte_size,
-            l_request.timestamp_type,
-            (EXTRACT(EPOCH FROM arg_now AT TIME ZONE 'UTC') * 1000)::BIGINT,
-            l_request.batch_max_timestamp
-        );
+        -- This request is accepted. Either extend the open run, or flush it and open a new one.
+        -- The byte-adjacency check (see header) guarantees no rejected batch intervened, so
+        -- offsets within the run are contiguous.
+        IF l_run_open
+            AND l_run_topic_id = l_request.topic_id
+            AND l_run_partition = l_request.partition
+            AND l_run_magic = l_request.magic
+            AND l_run_timestamp_type = l_request.timestamp_type
+            AND l_request.byte_offset = l_run_byte_offset + l_run_byte_size
+        THEN
+            l_run_last_offset := l_new_high_watermark - 1;
+            l_run_byte_size := l_run_byte_size + l_request.byte_size;
+            l_run_max_timestamp := GREATEST(l_run_max_timestamp, l_request.batch_max_timestamp);
+        ELSE
+            IF l_run_open THEN
+                PERFORM flush_commit_run_v2(
+                    l_new_file_id, l_run_magic, l_run_topic_id, l_run_partition,
+                    l_run_base_offset, l_run_last_offset, l_run_byte_offset, l_run_byte_size,
+                    l_run_timestamp_type, l_log_append_timestamp, l_run_max_timestamp);
+            END IF;
+            l_run_open := TRUE;
+            l_run_topic_id := l_request.topic_id;
+            l_run_partition := l_request.partition;
+            l_run_magic := l_request.magic;
+            l_run_timestamp_type := l_request.timestamp_type;
+            l_run_base_offset := l_assigned_offset;
+            l_run_last_offset := l_new_high_watermark - 1;
+            l_run_byte_offset := l_request.byte_offset;
+            l_run_byte_size := l_request.byte_size;
+            l_run_max_timestamp := l_request.batch_max_timestamp;
+        END IF;

         RETURN NEXT (l_request.topic_id, l_request.partition, l_log.log_start_offset, l_assigned_offset, l_request.batch_max_timestamp, 'none')::commit_batch_response_v1;
     END LOOP;

+    -- Flush the final open run.
+    IF l_run_open THEN
+        PERFORM flush_commit_run_v2(
+            l_new_file_id, l_run_magic, l_run_topic_id, l_run_partition,
+            l_run_base_offset, l_run_last_offset, l_run_byte_offset, l_run_byte_size,
+            l_run_timestamp_type, l_log_append_timestamp, l_run_max_timestamp);
+    END IF;
+
     -- Transfer from the temporary to real table.
     UPDATE logs
     SET high_watermark = logs_tmp.high_watermark,
```

Tests:
- `CommitFileJobCoalescingTest` — v1↔v2 response parity, row reduction, run-breaks (partition change, mid-run duplicate), cross-producer coalescing (one row, per-producer `producer_state`), offset-based delete/prune over coalesced rows, and time-based retention (kept/deleted whole by the row's max timestamp; safe over-retain when any sub-batch is live).
- `FetchCompleterTest` — multi-batch offset relocation, split-across-extents, LOG_APPEND_TIME, and a mismatch guard.
- The full shared `PostgresControlPlaneTest` suite runs against v2 (production default), proving behavioral parity.

Benchmark test also proves the improvements on commit and find batches time:

```
== WRITE PATH: coalescing (commit_file_v2) vs baseline (commit_file_v1) ==
Row reduction ratio equals batches-per-partition-per-window; sparse-wide is the 1-row-per-partition floor where coalescing cannot help.

shape                                    |    v1 rows |    v2 rows |   ratio |   v1 ms/file |   v2 ms/file
--------------------------------------------------------------------------------------------------------
high-fan-in (4 parts x 16 batches)       |      12800 |        800 |  16.00x |        8.970 |        6.373
frequent-tiny (8 parts x 8 batches)      |      12800 |       1600 |   8.00x |        6.307 |        4.666
sparse-wide (64 parts x 1 batch)         |       6400 |       6400 |   1.00x |        7.299 |        6.275
moderate (16 parts x 4 batches)          |       9600 |       2400 |   4.00x |        6.974 |        4.368

Note: rows are per-topic and exact; commit latency is wall-clock per commit against a local container and is indicative, not a microbenchmark.


== READ PATH: find_batches metadata query (whole-log scan per partition) ==
Coalescing returns K x fewer batch_info entries, shrinking the scan + payload. Object-storage byte I/O is NOT affected (ranges are already merged), so this is the metadata path only.

shape                                    |   v1 batches |   v2 batches |   ratio |   v1 us/find |   v2 us/find
--------------------------------------------------------------------------------------------------------------
high-fan-in (4 parts x 16 batches)       |        12800 |          800 |  16.00x |    39929.738 |     3529.252
frequent-tiny (8 parts x 8 batches)      |        12800 |         1600 |   8.00x |    11741.324 |     2676.897
sparse-wide (64 parts x 1 batch)         |         6400 |         6400 |   1.00x |     1379.748 |     1252.087
moderate (16 parts x 4 batches)          |         9600 |         2400 |   4.00x |     4712.953 |     1684.530

Note: batch counts are exact; find latency is wall-clock per find_batches call (whole-log, offset 0) and is indicative. It measures only the control-plane metadata query, not object-storage fetch.

```


storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/CommitFileJob.java
```diff
@@ -53,8 +55,15 @@ class CommitFileJob implements Callable<List<CommitBatchResponse>> {
...
     private final List<CommitBatchRequest> requests;
+    private final boolean coalesce;
     private final Consumer<Long> durationCallback;
 
+    /**
+     * Production entry point: always commits via {@code commit_file_v2}, which coalesces contiguous
+     * same-partition batch runs into a single {@code batches} row (see V15 migration). {@code commit_file_v1}
+     * is retained only for rollback safety and for the coalescing benchmark to measure against; new code
+     * must not depend on it and it should be dropped once no supported release calls it.
+     */
```
viktorsomogyi: Is this generally safe during a rolling upgrade? If one broker knows coalesced file commit and does it, then an old broker may try to read it and possibly get incorrect results.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/CommitFileJob.java
```diff
@@ -53,8 +55,15 @@ class CommitFileJob implements Callable<List<CommitBatchResponse>> {
     private final int uploaderBrokerId;
     private final long fileSize;
     private final List<CommitBatchRequest> requests;
+    private final boolean coalesce;
     private final Consumer<Long> durationCallback;
 
+    /**
+     * Production entry point: always commits via {@code commit_file_v2}, which coalesces contiguous
+     * same-partition batch runs into a single {@code batches} row (see V15 migration). {@code commit_file_v1}
```
viktorsomogyi: nit: V14

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/CommitFileJob.java
jeqo: ah, good catch! I missed this, and can be problematic. I was thinking on whether splitting this in separate commits, but it would be on the operator to validate how to ship it -- which doesn't seem fair.
Looks like a config flag would be needed here. I was trying to avoid to add more of these, but for now it seems to be the safest option here.

viktorsomogyi: If we could bind this to the metadata version then we could get away without extra configs. It's not a clean fit though because it conflates control plane schema version with metadata version, so the config might still be a better pick from this perspective, although we have to handle it on upper levels. Ideally perhaps it would be beneficial to keep a separate "control plane schema version" or something like that somewhere. With these changes it's beneficial and brokers could keep using the old one up until all of them are upgraded.

jeqo: Good idea. Having this flexibility to negotiate PG versions based on metadata can be useful if this type of changes become a pattern. Let's use the config flag for now, and keep a pin on the metadata-based approach if that's ok. 
Config change is added in the last commit

viktorsomogyi: Thanks, agreed, looks good.

viktorsomogyi (approved):

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/CommitFileJob.java
viktorsomogyi: created a ticket in the backlog for now: https://aiven.atlassian.net/browse/KC-294

PR #672, opened 2026-06-29
fix(inkless:consolidation): start consolidation when remote storage is enabled on a diskless topic

viktorsomogyi:
The controller applies a classic-to-consolidated switch as separate metadata deltas: remote.storage.enable=true arrives as a config-only delta with no leader or partition change, so the leader-delta path never starts the consolidation fetcher. TopicConfigHandler now kicks off consolidation for the topic's online partitions when remote storage flips from false to true on a diskless topic; the reconciler skips any partitions that are not ready yet and it is a no-op for non-consolidating topics.

Note: this hook resolves partitions from logManager.logsByTopic(topic), so it only fires for a topic that still has a local log when remote storage flips on -- i.e. a topic switched from classic, whose classic prefix is still local. A born-diskless topic left untiered before consolidation was enabled has no local log, so this path does not cover that untiered-diskless -> consolidated upgrade.

### DIFF
--- a/core/src/main/scala/kafka/server/ConfigHandler.scala
+++ b/core/src/main/scala/kafka/server/ConfigHandler.scala
@@ -83,6 +83,17 @@ class TopicConfigHandler(private val replicaManager: ReplicaManager,
         rlm.onLeadershipChange((leaderPartitions.toSet: Set[TopicPartitionLog]).asJava, (followerPartitions.toSet: Set[TopicPartitionLog]).asJava, topicIds))
     }
 
+    // Start consolidation on a classic -> consolidated switch. remote.storage.enable=true arrives as a
+    // config-only delta after the seal, with no partition/leader change, so applyLocalLeadersDelta never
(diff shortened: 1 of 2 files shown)


viktorsomogyi:
Reopened #652 as the fix got merged onto a feature branch and disappeared in a later rebase of the base branch.

giuseppelillo (approved):

jeqo:
@viktorsomogyi Missed leaving a comment here before it got merged, but I was wondering if we could make this mostly obsolete by having the switch auto-enable remote.storage atomically — similar to how #619 does it on creation.

viktorsomogyi:
@jeqo possibly, if we can atomically enable remote.storage when switching to diskless, it could be possible to get rid of this piece of code and it would be more symmetric with the creation. Do you think we should pursue that path?

jeqo:
@viktorsomogyi I think so, it should simplify some conditions and edge cases. Will look into it

viktorsomogyi:
@jeqo created the ticket for this: https://aiven.atlassian.net/browse/KC-293

PR #674, opened 2026-06-29
test(inkless:consolidation): add born-consolidated pipeline + WAL-prune system test

viktorsomogyi:
Add ConsolidationPipelineTest for a born-consolidated topic, asserting the full WAL -> local log -> remote pipeline: consolidation JMX gauges are exported, the MinIO tiered-storage prefix grows, the Postgres control plane prunes the WAL, and every acked record is still consumable.

Consolidation is enabled per-test via KafkaService(consolidation=True) (also flips the isolated controller), so it runs in the unified inkless ducktape suite without a run-wide --globals flag. JmxTool runs on brokers only since the controller never exposes the consolidation MBeans.

Add ConsolidationVerifier helper (JMX gauges, MinIO via mc, control plane via psql) and install mc + postgresql-client in the ducker image.

### DIFF
--- a/tests/kafkatest/services/inkless/consolidation_verifier.py
+++ b/tests/kafkatest/services/inkless/consolidation_verifier.py
@@ -204,13 +204,17 @@ class ConsolidationVerifier(object):
         return keys
 
     def tiered_object_count(self):
-        # List only the tiered-storage prefix: the bucket persists across runs, so
-        # scanning the whole bucket here (called repeatedly in wait_until) would
-        # get slower and flakier over time.
+        # Count under the tiered-storage prefix only (scanning the whole bucket
+        # in wait_until would slow and flap). Not topic-scoped, and the prefix
+        # accumulates across tests in one `ducker-ak test` run (the whole inkless
+        # suite is one invocation). Use baseline + delta (`> baseline`);
+        # absolute `== N` flaps with test order.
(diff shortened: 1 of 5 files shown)


tests/kafkatest/services/inkless/consolidation_verifier.py
```diff
@@ -0,0 +1,268 @@
...
+            key = obj.get("key")
+            if key:
+                keys.append(key)
+        return keys
+
+    def tiered_object_count(self):
+        # List only the tiered-storage prefix: the bucket persists across runs, so
+        # scanning the whole bucket here (called repeatedly in wait_until) would
+        # get slower and flakier over time.
+        return len(self.object_keys(self.TIERED_PREFIX))
```
jeqo, line 212: nit: should we pass the topic name/id here to validate them specifically and guard it against reruns? or add a comment to warn on the results when re-running it?

jeqo (commented):
LGTM, just a minor comment before approving

tests/kafkatest/services/inkless/consolidation_verifier.py
viktorsomogyi: So this is called twice during the test run: first to establish a baseline at the beginning and then to count the delta at verification. This makes it safe against reruns, just the baseline increases on every run. I'll amend the comment to reflect this.

jeqo (approved):

PR #673, opened 2026-06-29
fix(inkless:consolidation): recover a switched consolidated leader after local-log loss

viktorsomogyi:
A leader that comes up below the seal with an empty local log (a full local-storage wipe / disaster recovery where the controller, whose metadata survived, elects a replica that came back empty) has no peer to replicate the classic prefix from -- that prefix lives only in the remote tier. The ConsolidationReconciler now starts consolidation for such a leader (when remote storage is enabled) instead of retrying forever, so the fetcher lands below the diskless WAL start, hits OFFSET_MOVED_TO_TIERED_STORAGE, and rebuilds the leader-epoch cache and producer snapshot from remote.

PR #677, opened 2026-06-30
fix(inkless:consolidation): don't fence a consolidating leader below the seal

viktorsomogyi:
maybeReconcileSwitchedLeader fenced a switched leader offline whenever its LEO fell below the committed seal. It runs during applyLocalLeadersDelta, before the ConsolidationReconciler, so the wiped-leader recovery from #673 never ran. The reconciler only sees online partitions, and a wiped consolidating leader (LEO 0 < seal) was offline before it could rebuild. Every fetch and offset request then failed with KAFKA_STORAGE_ERROR.

For a consolidating diskless topic with remote storage enabled, [0, seal) is in the remote tier and can be rebuilt, so leave the partition online and let the reconciler arm consolidation at the current LEO. Non-consolidating topics still fence, since their classic prefix has no remote copy and LEO < seal is unrecoverable corruption.

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -3363,9 +3363,10 @@ class ReplicaManager(val config: KafkaConfig,
                     s"seal $seal at LEO ${log.logEndOffset} with remote storage enabled; leaving online " +
                     s"for consolidation to rebuild the classic prefix from the remote tier.")
                 } else {
-                  stateChangeLogger.error(s"Leader partition $tp has LEO ${log.logEndOffset} below " +
-                    s"classic-to-diskless start offset $seal; cannot catch up from another replica. " +
-                    s"Marking the partition offline as its local log is corrupt below the committed seal.")
+                  stateChangeLogger.error(s"Leader partition $tp has LEO ${log.logEndOffset} below the " +
+                    s"classic-to-diskless seal $seal and the classic prefix [0, $seal) is locally incomplete " +
+                    s"and not recoverable from remote (no consolidation / remote storage on this broker); " +
+                    s"marking the partition offline. Cannot catch up from another replica.")
                   markPartitionOffline(tp)
                 }
               }
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -3321,8 +3321,11 @@ class ReplicaManager(val config: KafkaConfig,
    *  - LEO > seal: truncate down to the seal unless the local suffix [seal, LEO) is already
    *    materialized consolidated diskless data.
    *  - LEO == seal and HW < seal: advance HW to the seal so consumers can cross into diskless.
-   *  - LEO < seal: fence offline. The classic prefix is incomplete, so serving it would hide a
-   *    hole in acknowledged data.
+   *  - LEO < seal: fence offline, unless this is a consolidating diskless topic with remote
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3350,11 +3354,22 @@ class ReplicaManager(val config: KafkaConfig,
                 stateChangeLogger.info(s"Stale high watermark detected: advanced high watermark to seal offset $seal for " +
                   s"switched leader partition $tp")
               } else if (log.logEndOffset < seal) {
-                // This is unreachable in normal operation
-                stateChangeLogger.error(s"Leader partition $tp has LEO ${log.logEndOffset} below " +
-                  s"classic-to-diskless start offset $seal; cannot catch up from another replica. " +
-                  s"Marking the partition offline as its local log is corrupt below the committed seal.")
-                markPartitionOffline(tp)
+                if (isConsolidatingPartition(partition) && log.remoteLogEnabled()) {
```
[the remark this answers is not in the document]
viktorsomogyi: Won't do. For a classic→consolidated switch, the controller writes diskless.enable=true + remote.storage.enable=true together with the PENDING seal, and commits the real seal in a later operation. So by the time the seal-commit delta arrives, the remote.storage.enable ConfigRecord has long since been applied to the cache, and isConsolidatingDisklessTopic is true.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3350,11 +3353,21 @@ class ReplicaManager(val config: KafkaConfig,
...
```
jeqo, line 3356: personal note: this can be simplified when #678 lands -- not a blocker

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -3350,11 +3353,21 @@ class ReplicaManager(val config: KafkaConfig,
...
+                  // ConsolidationReconciler arm consolidation at the current LEO: the first fetch
+                  // lands below the diskless WAL start, answers OFFSET_MOVED_TO_TIERED_STORAGE, and
+                  // the tier-state machine rebuilds the log from remote.
+                  stateChangeLogger.warn(s"Switched leader partition $tp is below the classic-to-diskless " +
+                    s"seal $seal at LEO ${log.logEndOffset} with remote storage enabled; leaving online " +
+                    s"for consolidation to rebuild the classic prefix from the remote tier.")
+                } else {
+                  stateChangeLogger.error(s"Leader partition $tp has LEO ${log.logEndOffset} below " +
+                    s"classic-to-diskless start offset $seal; cannot catch up from another replica. " +
+                    s"Marking the partition offline as its local log is corrupt below the committed seal.")
```
jeqo: nit: a bit more explicit comment
```suggestion
                  stateChangeLogger.error(s"Leader partition $tp has LEO ${log.logEndOffset} below the " +
                    s"classic-to-diskless seal $seal and the classic prefix [0, $seal) is locally incomplete " +
                    s"and not recoverable from remote (no consolidation / remote storage on this broker); " +
                    s"marking the partition offline. Cannot catch up from another replica.")

```

jeqo (commented):
LGMT, just a small logging improvement

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: Applied it.

jeqo (approved):

PR #678, opened 2026-06-30
feat(inkless:switch): auto-enable remote.storage atomically on diskless switch

jeqo:
Make the classic-to-diskless topic switch enable `remote.storage.enable=true` **atomically** — in the same controller metadata delta as `diskless.enable=true` — mirroring how topic *creation* already auto-enables it. This closes the two-delta window that PR #672 compensated for, and lets the broker treat "diskless ⟹ remote-storage" as an invariant it can rely on.

## What changes

The controller injects `remote.storage.enable=true` into the AlterConfigs request **before** validation (both incremental and legacy paths), so:
- an invalid switch (e.g. a `cleanup.policy=compact` topic) is **rejected up front** by the existing remote-storage delete-policy guard, rather than half-applied; and
- the resulting `remote.storage.enable` ConfigRecord **co-commits** with the `diskless.enable` ConfigRecord and the per-partition switch-pending (`PENDING`) records in one atomic batch.

The injection is gated on the classic-to-diskless **switch flag** (`diskless.allow.from.classic.enable`), not on consolidation, so *every* switch enables remote storage — even before consolidation is turned on. A switched topic is therefore never "untiered diskless"; it simply consolidates once consolidation is enabled. To keep the phased rollout ordered, consolidation now **requires** the switch flag (`diskless.remote.storage.consolidation.enable` requires `diskless.allow.from.classic.enable=true`).

The injection **normalizes the operation**, not just its absence:
- a `DELETE` or a null-valued `SET` of `remote.storage.enable` (which would strip the override and leave the topic untiered) is overwritten with `SET true`;
- an explicit non-null value is left as-is — `true` is a no-op, `false` is kept so validation rejects the contradictory switch;
- on the legacy full-map path an *omitted* key is an implicit delete, so `remote.storage.enable` is re-pinned to `true` even for an already-tiered topic (unchanged value → no extra record; it only keeps the key out of the implicit-delete set).

Also guards against deletion of `diskless.enable` via the legacy AlterConfigs API (both implicit omission and explicit null), which would otherwise trigger a switch via the broker default without persisting `diskless.enable=true`.

With the switch atomic, the consolidation reconciler gate simplifies to `isDisklessTopic`, and PR #672's `TopicConfigHandler` hook (which started consolidation on a *separate* remote.storage flip) becomes dead code and is removed with its tests.

## Scope: unreleased, so unsupported inherited states are not handled in code

The switch and consolidation are gated behind flags and **not yet released**. States this code cannot produce, but that older metadata might carry, are explicitly **unsupported** rather than handled at runtime:
- **untiered diskless** (a switched topic with remote storage off): the consolidation reconciler marks the partition `Failed` (surfacing `FailedPartitionsCount`; controller-side `DisklessWithoutRemoteStorageCount` also alerts). Recovery is manual — set `remote.storage.enable=true` **and** force a leader-epoch change (rolling restart, reassignment, or preferred-leader election); a config-only flip does not re-run the reconciler.
- **an existing classic topic left unpinned under a diskless broker default**: pin `diskless.enable` explicitly (`false` to stay classic, `true` to switch).

## Commits (reviewable independently)

- **`feat(inkless:switch): auto-enable remote.storage atomically on diskless switch`** — the core change: request augmentation (with `DELETE`/null normalization) + wiring + the switch-flag gate + the `consolidation requires allow-from-classic` config check + legacy `diskless.enable` deletion guard + reconciler gate collapse to `isDisklessTopic` + #672 hook removal. Controller, `LogConfig`, and reconciler tests, including atomic co-commit, compacted-topic fail-fast, and the `DELETE`/null injection cases.
- **`fix(inkless:consolidation): fail consolidation when remote storage disabled`** — a diskless topic with remote storage durably off (only reachable from a pre-atomic-switch image) is an unsupported broken invariant, not transient lag. Mark the partition `Failed` (not `Retry`): arming consolidation would materialize the WAL locally but never tier/prune it, growing the local log unbounded. The partition stays online for reads/writes; the failed flag clears on the next leader-epoch change.
- **`test(inkless:consolidation): drop untiered-diskless enable systest`** — removes the systest that exercised the now-removed `TopicConfigHandler` remote-flip hook (its Phase C can no longer be triggered). Seal-crossing read-after-prune durability remains covered by `SwitchedReadFromRemoteAfterPruneTest`.

[KC-293]

[KC-293]: https://aiven.atlassian.net/browse/KC-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/test/scala/unit/kafka/server/DynamicConfigChangeTest.scala
jeqo, line 1: the hook tested here has been removed

checkstyle/suppressions.xml
```diff
@@ -324,6 +324,9 @@
     <!-- ReplicationControlManager has multiple long methods; suppression needed for diskless managed replicas logic -->
     <suppress checks="MethodLength"
               files="ReplicationControlManager.java"/>
+    <!-- ReplicationControlManager crossed the class NCSS limit as diskless switch logic was added; the class already carries several complexity suppressions above -->
+    <suppress checks="JavaNCSS"
+              files="ReplicationControlManager.java"/>
```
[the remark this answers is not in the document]
jeqo: pragmatic, and this requires a refactoring out of scope here.

metadata/src/main/java/org/apache/kafka/controller/ConfigurationControlManager.java
```diff
@@ -388,6 +388,13 @@ private ApiError validateAlterConfig(
                 return DISALLOWED_BROKER_MIN_ISR_TRANSITION_ERROR;
             } else if (isDisallowedClusterMinIsrTransition(configRecord)) {
                 return DISALLOWED_CLUSTER_MIN_ISR_REMOVAL_ERROR;
+            } else if (Objects.equals(configRecord.name(), TopicConfig.DISKLESS_ENABLE_CONFIG)) {
```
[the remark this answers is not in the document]
jeqo: the guard at line 391 covers the real threat (implicit omission). Explicit null via legacy is not a client-reachable path, and even if it were, it's a deletion, not a switch trigger.

metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
```diff
@@ -3183,6 +3183,87 @@ private ApiError validatePartitionsForSwitch(TopicControlInfo topicInfo) {
...
+     *   <li><b>Fail-fast</b> — config validation ({@code LogConfig.validate}) runs on the augmented
+     *       request, so an invalid switch (e.g. a {@code cleanup.policy=compact} topic, rejected by the
+     *       remote-storage delete-policy guard) is rejected up front instead of half-applied.</li>
+     * </ul>
+     * No-op when consolidation is off, the topic is not switching to diskless,
+     * or remote storage is already requested/enabled.
+     * The returned map shares the input's inner maps by reference for untouched entries and copies only
+     * the inner maps it augments; callers must not mutate the shared inner maps.
+     */
+    Map<ConfigResource, Map<String, Entry<OpType, String>>> maybeAddRemoteStorageEnableForSwitch(
```
jeqo, line 3217: core change

viktorsomogyi:
We may need to remove this test because it exercises the path that you'll remove now.
https://github.com/aiven/inkless/blob/main/tests/kafkatest/tests/inkless/consolidation_enable_on_untiered_diskless_test.py

jeqo:
@viktorsomogyi good call, I'll drop it

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -129,6 +131,26 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
...
+        // Skip with Retry — not Fail — so the partition recovers automatically once operators re-enable
+        // remote storage. The controller-side KafkaController:DisklessWithoutRemoteStorageCount metric
+        // alerts them; this warn surfaces the same anomaly on the affected broker.
+        // Two log lines are emitted per retry: the WARN below is an operator-facing alert (leads with the
+        // anomaly, includes the recovery action); the INFO at the Retry call-site is the generic reconcile
+        // outcome trace. Different prefixes keep them visually distinct when scanning logs.
+        stateChangeLogger.warn(s"Diskless topic $tp has remote storage disabled: consolidation suspended " +
+          s"(violates diskless.enable implies remote.storage.enable). " +
+          s"Re-enable remote.storage.enable to resume (see DisklessWithoutRemoteStorageCount).")
+        ConsolidationStartState.Retry(s"Reconcile result for $tp: retry — remote storage disabled")
```
viktorsomogyi: When the remote storage config is re-enabled then it will be a ConfigRecord change in the best case which will not trigger this path as there will be no PartitionRecordChanges (as the reconciliation is ultimately called via applyLocal*Delta). Probably it would be better to refactor this somehow so a config change triggers this (although I'm not sure what the right way would be).

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -129,6 +131,26 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
...
+        // only happen to a topic that reached this state by another path — e.g. switched or created
+        // born-diskless while consolidation was off, then consolidation enabled later, or carried across
+        // a KRaft upgrade. Consolidation would materialize the WAL locally but never tier or prune it
+        // (pruning needs highestRemoteOffset >= floor), growing the local log unbounded.
+        // Skip with Retry — not Fail — so the partition recovers automatically once operators re-enable
+        // remote storage. The controller-side KafkaController:DisklessWithoutRemoteStorageCount metric
+        // alerts them; this warn surfaces the same anomaly on the affected broker.
+        // Two log lines are emitted per retry: the WARN below is an operator-facing alert (leads with the
+        // anomaly, includes the recovery action); the INFO at the Retry call-site is the generic reconcile
+        // outcome trace. Different prefixes keep them visually distinct when scanning logs.
```
EelisK: nit: I think it's good to state this assumption somewhere (`diskless.enable is remote.storage.enable and thus always consolidating`), but this level of detail in an inline comment is difficult for a human (and probably AI) to maintain.

EelisK: We could have a discussion with the team about introducing some shared agent configs to this repo, so that claude and other tools can follow shared good practices?

EelisK (commented):
Other than Viktor's comment, I don't see any other functional issues. I have concerns with the codebase being increasingly AI-written, while having no shared guardrails in place, but that's a broader discussion to have with the team.

jeqo:
@viktorsomogyi sorry called for another review by mistake. 
I think the scenario for enabling remote-storage on a switched to diskless topic should be avoided, i.e. Controller should always enable remote-storage for diskless, including when only switching is enabled but consolidation is not. I'm looking into this now, so moving it to draft.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
jeqo: Good catch. This is a smell -- I've moved the PR to branch to do some refactoring, and will consider this on the fixes

viktorsomogyi:
@jeqo yes, perhaps we just need a way to disallow disabling remote storage until we have that functionality. I think we could address config validation strengthening in a follow-up PR if you prefer that.

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -550,6 +550,10 @@ class KafkaConfig private(doLog: Boolean, val props: util.Map[_, _])
         s"${ServerConfigs.DISKLESS_REMOTE_STORAGE_CONSOLIDATION_ENABLE_CONFIG} requires ${ServerConfigs.DISKLESS_MANAGED_REPLICAS_ENABLE_CONFIG}=true")
       require(remoteLogManagerConfig.isRemoteStorageSystemEnabled,
         s"${ServerConfigs.DISKLESS_REMOTE_STORAGE_CONSOLIDATION_ENABLE_CONFIG} requires ${RemoteLogManagerConfig.REMOTE_LOG_STORAGE_SYSTEM_ENABLE_PROP}=true")
+      // Consolidation builds on the classic-to-diskless switch: a switch auto-enables remote storage,
+      // so consolidation can assume diskless.enable implies remote.storage.enable.
+      require(disklessAllowFromClassicEnabled,
+        s"${ServerConfigs.DISKLESS_REMOTE_STORAGE_CONSOLIDATION_ENABLE_CONFIG} requires ${ServerConfigs.DISKLESS_ALLOW_FROM_CLASSIC_ENABLE_CONFIG}=true")
```
jeqo: this should enforce that we are moving by stages from supporting topic switch to ts-consolidation

core/src/main/scala/kafka/server/ControllerServer.scala
```diff
@@ -242,6 +242,7 @@ class ControllerServer(
           setDisklessStorageSystemEnabled(config.disklessStorageSystemEnabled).
           setDisklessManagedReplicasEnabled(config.disklessManagedReplicasEnabled).
           setDisklessRemoteStorageConsolidationEnabled(config.disklessRemoteStorageConsolidationEnabled).
+          setDisklessAllowFromClassicEnabled(config.disklessAllowFromClassicEnabled).
```
jeqo, line 245: needed to make the diskless->remote.storage invariant validation earlier on topic-switching so the "untiered-diskless" topic scenario is unsupported.
The workaround for those scenario where failed partitions on ConsolidationReconciler are listed would be to fix the configs and rebalance partitions (called out in the commit msg now)

checkstyle/suppressions.xml
```diff
@@ -86,9 +86,10 @@
               files="MemoryRecordsBuilder.java"/>
     <suppress checks="ParameterNumber"
               files="ClientUtils.java"/>
-    <!-- ReplicationControlManager is at the edge of crossing this limit; adding it to include params for diskless logic -->
+    <!-- ReplicationControlManager is at the edge of crossing this limit; adding it to include params for diskless logic.
+         The test context constructor mirrors the same param list, so it crosses the limit too. -->
     <suppress checks="ParameterNumber"
-              files="ReplicationControlManager.java"/>
+              files="(ReplicationControlManager|ReplicationControlManagerTest).java"/>
```
jeqo, line 92: we can have a followup to break this into Inkless tests

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -129,6 +131,22 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
         ConsolidationStartState.Ready(initialFetchOffset(partition.localLogOrException))
       case PartitionRegistration.CLASSIC_TO_DISKLESS_SWITCH_PENDING =>
         ConsolidationStartState.Retry(s"Skipping consolidation for $tp because classic-to-diskless migration is still pending")
+      case seal if seal >= 0 && !inklessMetadataView.isRemoteStorageEnabled(tp.topic) =>
+        // Unsupported state: a switched topic (seal >= 0) with remote storage off.
+        // The switch auto-enables remote.storage.enable atomically with diskless.enable,
+        // so current code cannot produce this; it can only come from metadata written before that change
+        // (e.g. a pre-atomic-switch KRaft image).
```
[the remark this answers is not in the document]
jeqo: resolved now.

metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
```diff
@@ -1952,15 +1961,20 @@ public CompletableFuture<Map<ConfigResource, ApiError>> incrementalAlterConfigs(
             return CompletableFuture.completedFuture(Map.of());
         }
         return appendWriteEvent("incrementalAlterConfigs", context.deadlineNs(), () -> {
+            // Auto-enable remote.storage.enable=true on a classic-to-diskless switch, before validation,
+            // so it is validated (fail-fast on an invalid switch, e.g. a compacted topic) and co-commits
+            // atomically with the diskless.enable record and the per-partition switch-pending records.
```
EelisK:
```suggestion
```

metadata/src/main/java/org/apache/kafka/controller/QuorumController.java
```diff
@@ -2007,15 +2021,19 @@ public CompletableFuture<Map<ConfigResource, ApiError>> legacyAlterConfigs(
             return CompletableFuture.completedFuture(Map.of());
         }
         return appendWriteEvent("legacyAlterConfigs", context.deadlineNs(), () -> {
+            // See incrementalAlterConfigs: auto-enable remote.storage.enable=true on a classic-to-diskless
+            // switch before validating, so the switch is validated and committed atomically.
```
EelisK:
```suggestion
```

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -129,6 +131,22 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
...
+        // It is a durable broken invariant, not transient, so mark the partition Failed
+        // (flags it on FailedPartitionsCount) rather than Retry.
+        // Consolidating it would grow the local log unbounded (WAL materialized, never pruned).
+        // The partition stays online for reads/writes; the failed flag clears on the next leader-epoch change 
+        // (restart/reassignment), which re-runs this reconcile.
```
EelisK: Is this true?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -550,6 +550,10 @@ class KafkaConfig private(doLog: Boolean, val props: util.Map[_, _])
         s"${ServerConfigs.DISKLESS_REMOTE_STORAGE_CONSOLIDATION_ENABLE_CONFIG} requires ${ServerConfigs.DISKLESS_MANAGED_REPLICAS_ENABLE_CONFIG}=true")
       require(remoteLogManagerConfig.isRemoteStorageSystemEnabled,
         s"${ServerConfigs.DISKLESS_REMOTE_STORAGE_CONSOLIDATION_ENABLE_CONFIG} requires ${RemoteLogManagerConfig.REMOTE_LOG_STORAGE_SYSTEM_ENABLE_PROP}=true")
+      // Consolidation builds on the classic-to-diskless switch: a switch auto-enables remote storage,
+      // so consolidation can assume diskless.enable implies remote.storage.enable.
```
EelisK: nit: this assumption is mentioned in many comments, but we could probably just document it in one place instead?


```suggestion
```

core/src/test/scala/io/aiven/inkless/consolidation/ConsolidationReconcilerTest.scala
```diff
@@ -114,6 +117,27 @@ class ConsolidationReconcilerTest {
...
   }
 
+  @Test
+  def testConsolidationFailsWhenRemoteStorageDisabledOnSwitchedTopic(): Unit = {
+    // Unsupported state: a switched topic (seal >= 0) with remote storage OFF violates
+    // diskless.enable _is_ remote.storage.enable. The atomic switch always writes both flags together,
+    // so this is only possible for metadata written by an older, pre-atomic-switch path. Arming
+    // consolidation would materialize the WAL locally but never tier/prune it, growing the log unbounded.
+    // It is a durable broken invariant, not transient, so the reconciler marks the partition Failed
+    // (the failed flag clears on the next leader-epoch change); the controller-side metric alerts operators.
```
EelisK: nit: repetition 🔍 
```suggestion
```

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
```diff
@@ -81,17 +81,19 @@ class ConsolidationReconciler(replicaManager: ReplicaManager,
...
+      // The sole caller (ReplicaFetcherThread self-eviction, via ReplicaManager) selects partitions by
+      // seal state (committed seal + local LEO caught up), not by whether the topic is diskless,
+      // so this gate is where that precondition is established.
+      //
+      // Gating on diskless alone is sufficient (no separate remote-storage check needed):
+      // the reconciler only exists when broker consolidation is enabled (ReplicaManager builds it only then),
+      // and under the invariant `diskless.enable _is_ remote.storage.enable` (enforced atomically on the switch),
+      // a diskless topic is always consolidating.
+      // A topic that durably violates that invariant (remote storage off in the metadata cache) is not
+      // filtered out here — reconcileSwitchedConsolidatingDisklessPartition marks it Failed.
```
EelisK:
```suggestion
```

EelisK (dismissed):
LGTM 👍 I think we could document the behavior and clean up the comments a bit, but since it's a cosmetic change, I'll leave it up to you

core/src/main/scala/kafka/server/KafkaConfig.scala
jeqo: Good call. I've move it into ConsolidationReconciler javadoc and removed it from the suggested places.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationReconciler.scala
jeqo: Yes, the reconciler for this partition will stay off until a leader/follower change happen and it's triggered again. I've clarified this comment as it goes a bit hard to grasp.

EelisK (approved):

PR #679, opened 2026-07-01
feat(inkless:consolidation): route consolidation fetcher through coldpath to avoid cache pollution [KC-171]

jeqo:
Route the consolidation Reader through the cold fetch path to avoid polluting the shared object cache with consolidation reads.

## Behavior

- **Cache read-through, no write-back**: consolidation reads check the ObjectCache on hit (piggybacking for free on blobs a producer/consumer already cached), but on a miss they fetch cold via `backgroundStorage` — bypassing the cache entirely rather than populating it. This prevents consolidation from evicting hot consumer data.
- **Cold path without a dedicated pool**: when `laggingFetchDataExecutor` is null but a cold-path fetcher is present, the Reader reuses `fetchDataExecutor`. For consolidation this is always the case — the consolidation Reader never takes the hot path, so its data pool is otherwise idle and is reused for cold fetches. Thread-pool monitoring and shutdown apply only to Reader-owned pools (`ownsLaggingExecutor`).
- **Consumer lagging path is independent of consolidation**: `laggingFetchStorage` is allocated whenever `fetch.lagging.consumer.thread.pool.size > 0`, regardless of whether consolidation is enabled. The two cold paths (consolidation via `backgroundStorage`, consumer lagging via `laggingFetchStorage`) coexist — during the consolidation catch-up window a lagging consumer can still bypass the cache.

## Config

- `diskless.consolidation.fetch.lagging.request.rate.limit` (default 0 = unlimited): a safety valve to bound object-storage request rate from consolidation cold-path fetches.

Consolidation cold-path concurrency is tuned via the existing `diskless.consolidation.fetch.data.thread.pool.size` (the data pool is reused for the cold path, so no separate pool config is needed).

## Notes

- Reader/FetchPlanner carry `isConsolidationFetch` as an instance property rather than a per-call parameter (the flag never varies within a caller), simplifying `FetchHandler.handle()`.
- Lagging consumer thread name changed from `fetch-lagging-consumer` to `fetch-lagging` — update any monitoring/alerting that matches on the old thread name.


storage/inkless/src/main/java/io/aiven/inkless/consume/Reader.java
```diff
@@ -178,16 +224,23 @@ public Reader(
             maxBatchesPerPartitionToFind,
             Executors.newFixedThreadPool(fetchMetadataThreadPoolSize, new InklessThreadFactory(threadNamePrefix + "fetch-metadata-", false)),
             Executors.newFixedThreadPool(fetchDataThreadPoolSize, new InklessThreadFactory(threadNamePrefix + "fetch-data-", false)),
```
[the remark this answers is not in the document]
jeqo: This is not a leak, and the "validate before allocating" restructure would break from the module's established constructor idiom.
And it's pre-existing. This isn't something the PR introduced. On main the public constructor already allocates both fixed pools as this(...) arguments (Reader.java:179-180 on main) and already throws at the fetcher argument after them (the orElseThrow at Reader.java:182). This PR only changed the exception type (IllegalStateException → IllegalArgumentException) and the else-branch (: null → : orElse(null)); the allocate-then-validate argument order is untouched, pre-existing behavior — so even if it were worth restructuring, it's out of scope for this change.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -307,13 +308,24 @@ class ReplicaManager(val config: KafkaConfig,
...
+          // Cold path: use backgroundStorage to bypass cache for unconsolidated blob fetches.
+          // Lagging pool size is 0: the consolidation Reader always takes the cold path, so the cold
+          // path reuses the (otherwise idle) consolidation data pool rather than allocating a second one.
+          // Tune concurrency via diskless.consolidation.fetch.data.thread.pool.size.
+          Optional.of[ObjectFetcher](state.backgroundStorage()),
+          // Reuse the consumer lagging threshold (default -1 = cache TTL) as a recency cutoff.
+          // isConsolidationFetch forces the cold *path* regardless; this only selects range alignment:
+          // data younger than the cutoff is fixed-block aligned so the cache peek can reuse a
+          // consumer-cached block, older data is bounding-range aligned for a cheaper cold fetch.
+          state.config().fetchLaggingConsumerThresholdMs(),
```
jeqo, line 320: was trying to get rid of this config as threshold is not much useful for consolidation; but it's currently wired not only to cache key'ing but the alignment that affects the object fetching -- and would require a further refactoring that increased the scope of this PR too much, so leaving for later.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -262,6 +268,33 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
             return withHedge(primary, objectFetcher, request, fetchDataExecutor, firstByteReceived,
                 CompletableFuture.completedFuture(null));
         } else {
+            // Consolidation recent data: check cache without writing on miss.
+            // Hot data may already be cached by a regular consumer fetch — reuse it for free.
+            // On miss, fall through to the cold fetch without populating the cache.
+            if (isConsolidationFetch) {
```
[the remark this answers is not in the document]
jeqo: A peek-miss is just a getIfPresent null lookup (CaffeineCache.java:107-113) — join() only runs on a hit — so gating saves nanoseconds on a background path while giving up the rare-but-valid reuse and re-coupling the peek to the lagging alignment decision; leaving as-is.

storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
jeqo, line 1: key entry point for review

storage/inkless/src/main/java/io/aiven/inkless/consume/InklessFetchMetrics.java
```diff
@@ -322,8 +324,11 @@ public void recordRecentDataRequest() {
     }
 
     /**
-     * Records a request that used the cold path (lagging consumer, bypasses cache).
-     * This is recorded for ALL cold path requests, regardless of rate limiting.
+     * Records a request that used the cold path (bypasses the cache).
```
AnatolyPopov, line 331: But we are counting cache hits here as well right? Or am I misunderstanding what is written above in FetchPlanner?

storage/inkless/src/main/java/io/aiven/inkless/consume/Reader.java
```diff
@@ -229,7 +283,7 @@ private static ExecutorService createBoundedThreadPool(String threadNamePrefix,
             0L,                          // keepAliveTime: unused for fixed pools (core threads don't time out)
             TimeUnit.MILLISECONDS,
             new ArrayBlockingQueue<>(queueCapacity),  // Bounded queue prevents OOM
-            new InklessThreadFactory(threadNamePrefix + "fetch-lagging-consumer-", false),
+            new InklessThreadFactory(threadNamePrefix + "fetch-lagging-", false),
```
AnatolyPopov, line 286: So the metrics will be also renamed, right? Are we using them anywhere?

jeqo: Will need to be remapped but are not in dashboards yet, so I think it's safe to rename for clarity now

storage/inkless/src/main/java/io/aiven/inkless/consume/InklessFetchMetrics.java
jeqo: Good catch. Yes, cache hits were being counted. I've fixed it now: a cache-hit peek now records under RecentDataRequestRate (hot reuse), and only true cold fetches record under LaggingConsumerRequestRate, so the hot/cold split holds on the consolidation MBean too.
Consumer metrics group is unaffected (its cache hits already went to RecentDataRequestRate). 
Pushed as a separate commit.

AnatolyPopov (approved):
LGTM

PR #685, opened 2026-07-06
feat(inkless:consolidation): delayed fetcher to honor minBytes/maxWaitMs

jeqo:
The consolidation fetcher previously called Reader.fetch() synchronously on every doWork() iteration regardless of whether data was available. Idle partitions therefore queried the control plane on every loop iteration: it returned empty and the loop ran again. Existing configs diskless.consolidation.fetch.{min.bytes,max.wait.ms} were passed into FetchParams but Reader.fetch() ignored them.

Wrap the fetcher path in DisklessLeaderEndPoint.fetch with a DelayedConsolidationFetch that parks in a new
delayedConsolidationFetchPurgatory until the initial probe sees minBytes of accumulated diskless bytes or maxWaitMs expires. tryComplete runs one cheap findDisklessBatches probe (gated by a CAS so tryCompleteElseWatch's before/after double-check probes only once); onComplete runs the full FetchHandler.handle once. This purgatory has no wake-up sites, so parked ops complete via the maxWaitMs timeout, keeping control-plane pressure bounded regardless of loop rate.

Set the purgatory purgeInterval to 0 (matches delayedRemoteFetchPurgatory) so completed ops release their captured response references immediately for GC. Each completed op pins a Map[TopicIdPartition, FetchPartitionData] of fetched records; without aggressive purging, watch lists accumulate up to fetchPurgatoryPurgeIntervalRequests completed ops and exhaust the heap.

Raise the consolidation fetch defaults now that min.bytes/max.wait.ms are honored: max.bytes 1MB to 10MB (per partition), response.max.bytes 10MB to 64MB (total), min.bytes 1 to 8MB, max.wait.ms 500 to 1000.

See commits for details.

### DIFF
--- a/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
+++ b/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
@@ -21,20 +21,20 @@ package io.aiven.inkless.consolidation
 import io.aiven.inkless.consume.{FetchHandler, FetchOffsetHandler}
 import kafka.server.{KafkaConfig, ReplicaManager, ReplicaQuota}
 import kafka.utils.Logging
-import org.apache.kafka.common.Uuid
 import org.apache.kafka.common.errors.{KafkaStorageException, UnknownTopicOrPartitionException}
-import org.apache.kafka.common.message.{FetchResponseData, OffsetForLeaderEpochRequestData}
 import org.apache.kafka.common.message.ListOffsetsRequestData.ListOffsetsPartition
 import org.apache.kafka.common.message.OffsetForLeaderEpochResponseData.EpochEndOffset
+import org.apache.kafka.common.message.{FetchResponseData, OffsetForLeaderEpochRequestData}
 import org.apache.kafka.common.protocol.Errors
 import org.apache.kafka.common.record.FileRecords.TimestampAndOffset
-import org.apache.kafka.common.requests.{FetchRequest, FetchResponse, ListOffsetsRequest, OffsetsForLeaderEpochResponse}
-import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.record.MemoryRecords
+import org.apache.kafka.common.requests.{FetchRequest, FetchResponse, ListOffsetsRequest, OffsetsForLeaderEpochResponse}
+import org.apache.kafka.common.{TopicIdPartition, TopicPartition, Uuid}
 import org.apache.kafka.metadata.{LeaderAndIsr, PartitionRegistration}
 import org.apache.kafka.server.common.{MetadataVersion, OffsetAndEpoch}
 import org.apache.kafka.server.network.BrokerEndPoint
-import org.apache.kafka.server.storage.log.{FetchIsolation, FetchParams}
+import org.apache.kafka.server.purgatory.TopicPartitionOperationKey
+import org.apache.kafka.server.storage.log.{FetchIsolation, FetchParams, FetchPartitionData}
 import org.apache.kafka.server.{LeaderEndPoint, PartitionFetchState, ReplicaFetch, ResultWithPartitions}
 import org.apache.kafka.storage.internals.log.OffsetResultHolder.FileRecordsOrError
 import org.apache.kafka.storage.internals.log.UnifiedLog
@@ -42,6 +42,7 @@ import org.apache.kafka.storage.internals.log.UnifiedLog
 import java.util
 import java.util.Optional
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ConcurrentHashMap
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 import scala.util.Try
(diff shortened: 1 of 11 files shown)


viktorsomogyi (dismissed):
All good from my side, but @AnatolyPopov mentioned yesterday that he'd like to take a look, so I'm not merging yet to give him a chance.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -315,15 +410,29 @@ class DisklessLeaderEndPoint(
...
+            val localLog = replicaManager.localLogOrException(topicPartition)
+            val logStartOffset = localLog.logStartOffset
+            val logConfig = localLog.config
+            // Reserve headroom for the follower append; clampRecordsToSegment enforces it per batch in
+            // fetch() (find_batches limits at coordinate granularity and a coalesced coordinate can
+            // overshoot by more than one batch). See clampRecordsToSegment.
+            val maxOvershootBytes = math.min(logConfig.maxMessageSize, logConfig.segmentSize - 1)
+            val segmentSafeFetchSize = math.max(1, logConfig.segmentSize - maxOvershootBytes)
+            val partitionFetchSize = math.min(fetchSize, segmentSafeFetchSize)
+            if (logConfig.maxMessageSize >= logConfig.segmentSize && unsafeSegmentConfigWarnings.add(topicPartition)) {
```
AnatolyPopov, line 430: Do I understand correctly that we can have a situation when maxMessageBytes has been lowered over the time and here we can get an old bigger message that will not fit?

jeqo: Agree, this is a valid, pre-existing edge case. A batch committed under a higher max.message.bytes (or a coalesced diskless unit) can exceed the current segment.bytes, and the follower append then rejects it with RecordBatchTooLargeException. Since the batch was already validated and committed at produce time, refusing to replicate it is the wrong outcome.

A few follow-ups worth considering: config-time validation rejecting max.message.bytes > segment.bytes on diskless topics (prevents new oversized batches, but not ones already committed); a per-partition metric for the largest batch/message size; and a stronger fix — an upstream change letting the follower append roll a dedicated oversized segment (what maybeRoll already does for a single batch on the leader path). That last one touches shared upstream code, so it's a deliberate follow-up, not this PR.

For now, adding observability (ERROR log + a per-partition metric) and converting the failure into a retriable per-partition error, so the partition parks and retries with backoff instead of hard-failing. That way bumping segment.bytes self-heals the fetcher without a become-follower/restart.

jeqo: Created [KC-345] to track follow-up

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -161,6 +173,89 @@ class DisklessLeaderEndPoint(
     }.toMap.asJava
   }
 
+  /**
+   * Trim a fetched block so the follower's single append fits `segment.bytes`, restoring the classic
+   * invariant `returned <= maxBytes + one physical batch (<= max.message.bytes) <= segment.bytes`.
```
AnatolyPopov: I think we are not following this invariant here because the last overshoot message can be bigger than max.message.bytes and can exceed the segment.bytes

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -269,6 +269,21 @@ class ReplicaManager(val config: KafkaConfig,
...
+  // Hosts long-poll DelayedConsolidationFetch operations created by the consolidation fetcher (DisklessLeaderEndPoint).
+  // Kept separate from delayedFetchPurgatory because the latter is strictly typed to DelayedFetch.
+  // Operations perform one initial metadata probe, then complete via maxWaitMs timeout if parked;
+  // wake-up sites are intentionally not mirrored here to keep control-plane pressure bounded.
+  //
+  // purgeInterval = 0 (matches delayedRemoteFetchPurgatory) so completed ops release their captured response
+  // references immediately for GC.
+  // Each completed op pins a Map[TopicIdPartition, FetchPartitionData] holding fetched records,
+  // bounded per partition by diskless.consolidation.fetch.max.bytes and in aggregate by
+  // diskless.consolidation.fetch.response.max.bytes.
```
AnatolyPopov: But is the aggregate boundary respected? I think the first coordinate of every partition will be fetched unconditionally and that can overflow the boundary.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
jeqo: Good catch. I've reworded the comment to state this is best-effort; not a guaranteed invariant.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Agree. the aggregate bound is best-effort, not hard cap: find_batches always admits each partition first coordinate, so it can overshoot by the sum of those. Should be benign (append is per-partition, so no append failure) -- heap-sizing only. Reworded the comment to say so.

giuseppelillo (approved):

PR #686, opened 2026-07-07
feat(inkless:switch): implement `repair` tool for diskless switch

EelisK:
This PR introduces repair tooling for diskless partitions.
This includes:
- Control plane operation `repairDisklessLog`
- Admin API `REPAIR_DISKLESS_LOG (502)`

The tool reconciles the diskless control plane with with KRaft,
by writing the seal offsets to the control plane.


storage/inkless/src/main/resources/db/migration/V18__Repair_diskless_log.sql
```diff
@@ -0,0 +1,47 @@
...
+        SELECT *
+        FROM unnest(arg_requests)
+    LOOP
+        IF l_request.diskless_start_offset < l_request.log_start_offset THEN
+            RAISE EXCEPTION 'diskless_start_offset (%) must be >= log_start_offset (%) for topic_id=% partition=%',
+                l_request.diskless_start_offset, l_request.log_start_offset,
+                l_request.topic_id, l_request.partition;
+        END IF;
+
+        INSERT INTO logs (topic_id, partition, topic_name, log_start_offset, high_watermark, byte_size, diskless_start_offset)
```
giuseppelillo: Let's keep this API strictly for repairs, meaning that diskless log init should always come from the regular switching flow. If a switch is stuck, we should attempt to unblock it on the broker side (the one that decides the disklessStartOffset), and not force the initialization on the Control Plane.
This API here will be needed only in case if a wrong offset has been chosen during the switch flow.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/RepairDisklessLogRequest.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package io.aiven.inkless.control_plane;
+
+import org.apache.kafka.common.Uuid;
+
+public record RepairDisklessLogRequest(
+    Uuid topicId,
+    String topicName,
+    int partition,
+    long logStartOffset,
```
giuseppelillo:
```suggestion
```
This is not needed if the API is only called for updates

giuseppelillo (changes requested):

storage/inkless/src/main/resources/db/migration/V18__Repair_diskless_log.sql
EelisK: Makes sense 👍

clients/src/main/java/org/apache/kafka/clients/admin/RepairDisklessLogOptions.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.apache.kafka.common.annotation.InterfaceStability;
+
+/**
+ * Options for {@link Admin#repairDisklessLog(String, int, int, RepairDisklessLogOptions)}.
+ *
+ * The API of this class is evolving, see {@link Admin} for details.
+ */
+@InterfaceStability.Unstable
+public class RepairDisklessLogOptions extends AbstractOptions<RepairDisklessLogOptions> {
```
giuseppelillo, line 28: Is this really needed?

tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
```diff
@@ -95,7 +95,11 @@ static void execute(String... args) throws Exception {
                 }
                 break;
             case "repair":
-                throw new RuntimeException("Command \"" + command + "\" not implemented");
+                try (Admin adminClient = Admin.create(properties)) {
+                    repairCommand(System.out, adminClient, topic,
+                        Optional.ofNullable(namespace.getInt("partition")));
+                }
+                break;
```
giuseppelillo, line 102: This command now requires to be sent to a broker, therefore we need to make it accept only `--bootstrap-server` and not `--bootstrap-controller`

giuseppelillo (commented):
LGTM, just a couple of comments

clients/src/main/java/org/apache/kafka/clients/admin/RepairDisklessLogOptions.java
EelisK: It seems to be the common pattern to include options for admin APIs even if they are empty. A couple examples:
https://github.com/aiven/inkless/blob/cf88633abe5e9ff32abdca98ebea09a71506a3ba/clients/src/main/java/org/apache/kafka/clients/admin/ElectLeadersOptions.java#L24-L30
https://github.com/aiven/inkless/blob/cf88633abe5e9ff32abdca98ebea09a71506a3ba/clients/src/main/java/org/apache/kafka/clients/admin/DeleteRecordsOptions.java#L22-L27

tools/src/main/java/org/apache/kafka/tools/TopicSwitchCommand.java
EelisK: Good catch 👍  addressed in https://github.com/aiven/inkless/pull/686/commits/d83c5b67827115f90d0777773ac5cc2579af9577

giuseppelillo (approved):

PR #689, opened 2026-07-08
feat: introduce vendor-neutral AI guardrails [KC-323]

EelisK:
This PR introduces vendor-neutral AI guardrails to the repo. The goal is to have shared standards for coding
agents working on this project, and to produce consistent quality code.

### DIFF
--- a/.ai-agents/README.md
+++ b/.ai-agents/README.md
@@ -1,0 +1,90 @@
+# AI Agents Configuration
+
+Shared configuration directory for AI coding assistants working in this
+repository.
+
+This directory provides a unified structure for configuring multiple AI coding
+tools (OpenCode, Cursor, VS Code Copilot, Claude Code) with consistent agents,
+skills, and commands.
+
+## How It Works
+
+The main entry point for all AI tools is [`AGENTS.md`](../AGENTS.md) in the
+repository root. Each tool-specific directory (`.opencode/`, `.cursor/`,
+`.claude/`) symlinks its `agents/`, `commands/`, and `skills/` back to this
+shared `.ai-agents/` directory, ensuring consistent configuration across tools.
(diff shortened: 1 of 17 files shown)


AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+The following requirement levels are used throughout this document to indicate
+the importance of a given rule or guideline:
+
+| Keyword            | How to treat it          |
+| ------------------ | ------------------------ |
+| MUST/REQUIRED      | mandatory                |
+| SHOULD/RECOMMENDED | deviate only with reason |
+| MAY/OPTIONAL       | use judgment             |
+
+## Repository Structure
```
jeqo, line 14: This is a good idea. I'd suggest update and pointing to `inkless-sync/config/protected-patterns.txt` that already has a similar purpose. I think we can adopt a sort of "CODEOWNERS" format to document this properly, but better to have it outside the md to maintain it, othewise can be easy to miss. Can do as a follow-up

.ai-agents/README.md
jeqo, line 1: this is referenced by path, we don't need a symlink right?

jeqo (dismissed):
LGTM, thanks @EelisK. A couple of comments, non-blockers

AGENTS.md
EelisK: Good catch, I'll update this PR 👍

.ai-agents/README.md
EelisK: We are using symlinks only for `agents/`, `commands/`, and `skills/` because tools auto-discover those at fixed paths (eg. `.claude/skills`). Everything else is referenced by paths since there's no similar constraints, eg. Claude sees the reference for `AGENTS.md` in `CLAUDE.md` and opencode sees it in `opencode.json`

jeqo (approved):

PR #697, opened 2026-07-14
fix(inkless:switch): clear under-replicated partitions after classic-to-diskless switch

giuseppelillo:
A follower that dropped out of ISR and recovered after a classic-to-diskless switch stayed under-replicated forever, keeping `UnderReplicatedPartitions` stuck.

- **Leader fetch handler:** Validate a switched follower's seal fetch through the classic follower checks, then record its fetch state so ISR can re-expand without reading diskless data locally. Divergence and stale broker epochs produce partition-level responses.
- **makeFollower:** Give a switched, at-seal, out-of-ISR follower a catch-up fetcher.
- **ReplicaFetcherThread:** Keep fetching until controller metadata shows the replica back in ISR.
- Add unit coverage for seal fetch validation, fetcher scheduling and eviction, plus a URP-recovery system test with complete live-broker JMX observations.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2508,6 +2508,7 @@ class ReplicaManager(val config: KafkaConfig,
         if (!partitionLookupFailed) {
           val disklessSwitchCompleted = !shouldReadFromUnifiedLog && classicToDisklessStartOffset >= 0
           if (params.isFromFollower && disklessSwitchCompleted) {
+            var fetchError = Errors.NONE
             var divergingEpoch = Optional.empty[FetchResponseData.EpochEndOffset]
             // A recovered follower for a switched partition may already be caught up to the
             // seal offset but still be outside ISR. Record the seal-offset fetch so the normal
@@ -2518,25 +2519,36 @@ class ReplicaManager(val config: KafkaConfig,
                 val requestEpochMatchesLeader =
                   fetchPartitionData.currentLeaderEpoch.toScala.forall(_.intValue() == partition.getLeaderEpoch)
                 if (requestEpochMatchesLeader) {
-                  partition.getReplica(params.replicaId).foreach { _ =>
-                    // Use the classic follower-read validation without returning any records.
-                    val fetchAtSeal = new PartitionData(
-                      fetchPartitionData.topicId,
-                      classicToDisklessStartOffset,
-                      fetchPartitionData.logStartOffset,
-                      0,
-                      fetchPartitionData.currentLeaderEpoch,
-                      fetchPartitionData.lastFetchedEpoch
-                    )
-                    val readInfo = partition.fetchRecords(
-                      fetchParams = params,
-                      fetchPartitionData = fetchAtSeal,
-                      fetchTimeMs = time.milliseconds,
-                      maxBytes = 0,
-                      minOneMessage = false,
-                      updateFetchState = true
-                    )
-                    divergingEpoch = readInfo.divergingEpoch
+                  try {
+                    partition.getReplica(params.replicaId).foreach { _ =>
+                      // Use the classic follower-read validation without returning any records.
+                      val fetchAtSeal = new PartitionData(
+                        fetchPartitionData.topicId,
+                        classicToDisklessStartOffset,
+                        fetchPartitionData.logStartOffset,
+                        0,
+                        fetchPartitionData.currentLeaderEpoch,
+                        fetchPartitionData.lastFetchedEpoch
+                      )
+                      val readInfo = partition.fetchRecords(
+                        fetchParams = params,
+                        fetchPartitionData = fetchAtSeal,
+                        fetchTimeMs = time.milliseconds,
+                        maxBytes = 0,
+                        minOneMessage = false,
+                        updateFetchState = true
+                      )
+                      divergingEpoch = readInfo.divergingEpoch
+                    }
+                  } catch {
+                    case e@(_: UnknownTopicOrPartitionException |
+                            _: NotLeaderOrFollowerException |
+                            _: UnknownLeaderEpochException |
+                            _: FencedLeaderEpochException |
+                            _: ReplicaNotAvailableException |
+                            _: KafkaStorageException |
+                            _: InconsistentTopicIdException) =>
+                      fetchError = Errors.forException(e)
                   }
                 }
               }
@@ -2550,9 +2562,9 @@ class ReplicaManager(val config: KafkaConfig,
             // local data intact and remains able to serve consumer reads from the local log.
             immediateFetchResponses += tp ->
               new FetchPartitionData(
-                Errors.NONE,
-                classicToDisklessStartOffset,
-                0L,
+                fetchError,
+                if (fetchError == Errors.NONE) classicToDisklessStartOffset else UnifiedLog.UNKNOWN_OFFSET,
+                if (fetchError == Errors.NONE) 0L else UnifiedLog.UNKNOWN_OFFSET,
                 MemoryRecords.EMPTY,
                 divergingEpoch,
                 OptionalLong.empty(),
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2540,17 +2540,21 @@ class ReplicaManager(val config: KafkaConfig,
                   } catch {
                     case NonFatal(e) =>
                       fetchError = Errors.forException(e)
+                      if (fetchError == Errors.UNKNOWN_SERVER_ERROR) {
+                        error(s"Error validating at-seal fetch from " +
+                          s"${FetchRequest.describeReplicaId(params.replicaId)} on partition $tp " +
+                          s"at seal $classicToDisklessStartOffset", e)
+                      }
                   }
                 case Left(error) => fetchError = error
               }
             }
             // The partition has fully switched to diskless and the follower is asking for an offset at or beyond it.
-            // Followers must never replicate diskless records into their local log. Return
-            // an empty response with HW clamped to the seal offset so the fetcher loop sees the
-            // partition as caught up and goes idle, rather than treating it as out of range.
-            // Deliberately pass logStartOffset=0 (a no-op for the follower since
-            // maybeIncrementLogStartOffset only ever advances) so the follower keeps its classic
-            // local data intact and remains able to serve consumer reads from the local log.
+            // Followers must never replicate diskless records into their local log.
+            // Empty records and HW at the seal offset make the follower treat the local log as caught up.
+            // ReplicaFetcherThread evicts once this replica is in ISR, or immediately if consolidating.
+            // logStartOffset=0 is a no-op for the follower (maybeIncrementLogStartOffset only ever advances),
+            // so classic local data stays in place and can still serve consumer reads.
             immediateFetchResponses += tp ->
               new FetchPartitionData(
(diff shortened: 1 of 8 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2508,6 +2509,41 @@ class ReplicaManager(val config: KafkaConfig,
...
+                      fetchPartitionData = fetchAtSeal,
+                      fetchTimeMs = time.milliseconds,
+                      maxBytes = 0,
+                      minOneMessage = false,
+                      updateFetchState = true
+                    )
+                    divergingEpoch = readInfo.divergingEpoch
+                  } catch {
+                    case NonFatal(e) =>
+                      fetchError = Errors.forException(e)
```
jeqo, line 2542: May be worth logging non-typed errors: 
```suggestion
                    case NonFatal(e) =>
                      fetchError = Errors.forException(e)
                    if (fetchError == Errors.UNKNOWN_SERVER_ERROR) {
                      error(s"Error validating at-seal fetch from " +
                        s"${FetchRequest.describeReplicaId(params.replicaId)} on partition $tp " +
                        s"at seal $classicToDisklessStartOffset", e)
                    }
```

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -4102,11 +4138,14 @@ class ReplicaManager(val config: KafkaConfig,
               val isNewLeaderEpoch = partition.makeFollower(info.partition, isNew, offsetCheckpoints, Some(info.topicId), partitionAssignedDirectoryId)
               partition.seal()
               changedPartitions.add(partition)
-              if (seal >= 0 && partition.localLogOrException.highWatermark < seal) {
+              val isOutOfIsr = !info.partition.isr.contains(config.brokerId)
+              if (seal >= 0 && (partition.localLogOrException.highWatermark < seal || isOutOfIsr)) {
```
jeqo: Aligning with how shutting down brokers are handled on classic topics:

```suggestion
              val isOutOfIsr = !info.partition.isr.contains(config.brokerId)
              // A shutting-down broker is not ISR-eligible on the leader (isReplicaIsrEligible), so a
              // catch-up fetch started here could never complete; the classic branch below stops
              // fetching in the same situation.
              if (seal >= 0 && !isInControlledShutdown &&
                  (partition.localLogOrException.highWatermark < seal || isOutOfIsr)) {
```

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -4102,11 +4138,14 @@ class ReplicaManager(val config: KafkaConfig,
...
                 // Schedule a catch-up fetch when the local HW is below the seal -- either
                 // because we restarted with a stale HW (unclean shutdown) or because we
                 // were just added as a replica and have an empty local log. The
                 // ReplicaFetcher self-evicts once the follower has read past the seal.
+                // Also schedule one fetch when this replica is already caught up but out
+                // of ISR, so the leader observes its fetch state and can expand ISR.
```
jeqo: let's update the claims as two are no longer true — "the ReplicaFetcher self-evicts once the follower has read past the seal" (eviction now also requires ISR membership, or a consolidating partition), and "schedule *one* fetch" for the out-of-ISR case (it is a backed-off loop until the ISR expansion lands)

```suggestion
                // Schedule a catch-up fetch when the local HW is below the seal -- either
                // because we restarted with a stale HW (unclean shutdown) or because we
                // were just added as a replica and have an empty local log.
                // Also schedule one when this replica is caught up but out of ISR, so the leader
                // observes its fetch state and can expand ISR.
                // Eviction waits for ISR membership, not just for the seal -- see ReplicaFetcherThread.processPartitionData.
```

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2517,11 +2553,11 @@ class ReplicaManager(val config: KafkaConfig,
             // local data intact and remains able to serve consumer reads from the local log.
```
jeqo: Updating that "goes idle" is no longer what happens,  the follower either evicts the partition (in ISR, or consolidating) or backs off and retries until the ISR expansion lands — idling is the outcome that is not possible now.
```suggestion
            // The partition has fully switched to diskless and the follower is asking for an offset at or beyond it.
            // Followers must never replicate diskless records into their local log. 
            // Return an empty response with HW clamped to the seal offset so the fetcher sees the partition
            // as caught up rather than out of range; the fetcher evicts it once in ISR, and backs off
            // and retries until then.
            // Deliberately pass logStartOffset=0 (a no-op for the follower since
            // maybeIncrementLogStartOffset only ever advances) so the follower keeps its classic
            // local data intact and remains able to serve consumer reads from the local log.
```

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
```diff
@@ -7055,44 +7074,383 @@ class ReplicaManagerInklessTest {
...
+      assertEquals(MemoryRecords.EMPTY, invalidEpochData.records)
+      assertEquals(UnifiedLog.UNKNOWN_OFFSET,
+        invalidEpochPartition.getReplica(followerId).get.stateSnapshot.logEndOffset)
+
+      val validData = responseData(validTopicPartition)
+      assertEquals(Errors.NONE, validData.error)
+      assertEquals(sealOffset, validData.highWatermark)
+      assertEquals(MemoryRecords.EMPTY, validData.records)
+      assertEquals(sealOffset, validPartition.getReplica(followerId).get.stateSnapshot.logEndOffset)
+      verify(alterPartitionManager, times(1)).submit(any(), any())
```
jeqo: This one proves only that some `AlterPartition` went out, not that it contained the recovered follower.

```suggestion
      verify(alterPartitionManager, times(1)).submit(
        ArgumentMatchers.eq(validTopicPartition), any())
```

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
```diff
@@ -7055,44 +7074,383 @@ class ReplicaManagerInklessTest {
...
+
+      assertNotNull(responseData)
+      val data = responseData(disklessTopicPartition)
+      assertEquals(Errors.NOT_LEADER_OR_FOLLOWER, data.error)
+      assertEquals(UnifiedLog.UNKNOWN_OFFSET, data.highWatermark)
+      assertEquals(UnifiedLog.UNKNOWN_OFFSET, data.logStartOffset)
+      assertEquals(MemoryRecords.EMPTY, data.records)
+      assertEquals(UnifiedLog.UNKNOWN_OFFSET,
+        partition.getReplica(followerId).get.stateSnapshot.logEndOffset)
+      verify(alterPartitionManager, never()).submit(any(), any())
```
jeqo, line 7194:
```suggestion
      verify(alterPartitionManager).submit(any(), argThat[LeaderAndIsr](_.isr.contains(followerId)))
```

jeqo (commented):
Looks good, just a few more comments and a finding:

`ReplicationControlManager#expandIsrForDisklessManagedPartitions` re-adds an unfenced broker to the ISR of every diskless.enable=true partition without consulting classicToDisklessStartOffset. I wrote that in #643 for born-diskless managed-RF, where any live replica is current because all the data is in object storage but missed the switched case, where the classic prefix below the seal exists only in the replicas' local logs. So today a returning replica can enter a switched partition's ISR with no evidence its prefix is complete, and is then electable. 

I can open a follow-up PR to skip partitions whose seal isn't `NO_CLASSIC_TO_DISKLESS_START_OFFSET` to let them earn ISR through the path this PR builds, if this feels like stretching the scope of this PR.

core/src/main/scala/kafka/server/ReplicaManager.scala
viktorsomogyi: I think "clamped" isn't the right wording here as we don't do min/max limiting. We simply set it to the classicToDisklessStartOffset. (As I noticed, Opus uses "clamping" many times and often incorrectly, so much so that I had to explicitly set a rule to avoid using this term :) ).
Will adopt the comment though, just with somewhat different wording.

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
viktorsomogyi: actually, why do we need this? The test supposed to test whether there is a partition error for a stale epoch, and in that case we should verify `never()` here. I can drop the mock for this method though earlier in the test to improve what the test is trying to do.

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
```diff
@@ -7055,44 +7074,384 @@ class ReplicaManagerInklessTest {
...
+      val data = responseData(disklessTopicPartition)
+      assertEquals(Errors.NONE, data.error)
+      assertEquals(MemoryRecords.EMPTY, data.records)
+      assertEquals(sealOffset, data.highWatermark)
+
+      // ...but because the request epoch matched the leader's, the leader recorded the follower's
+      // fetch state at the seal...
+      assertEquals(sealOffset, partition.getReplica(followerId).get.stateSnapshot.logEndOffset)
+      // ...which drives the normal ISR-expansion path for the now caught-up follower.
+      verify(alterPartitionManager).submit(any(), any())
```
jeqo: nit: this one proves that alter partition fired but not that the ISR inside contained the follower. 
```suggestion
      verify(alterPartitionManager).submit(any(), argThat[LeaderAndIsr](_.isr.contains(followerId)))
```

Needs import `org.apache.kafka.metadata.LeaderAndIsr`

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
jeqo: You're right. My comment is on the wrong line, sorry. :7194 is the stale-epoch test and `never()` is correct there; ignore the suggestion, I'll move it.

While here, there's a related gap in `...WithNewerLeaderEpochSkipsFetchState`, `...WithFencedLeaderEpochSkipsFetchState`, `...FromRemovedReplicaReturnsUnknownLeaderEpoch` and `...ReturnsStorageErrorWhenLogDirOffline` with `prepareAlterPartitionSubmit()` in place, an unexpected submit there would go unnoticed.

Dropping the mock works too, but the NPE lands in the NonFatal catch and surfaces as `UNKNOWN_SERVER_ERROR` which may make the troubleshooting harder.

viktorsomogyi: Yea, I'll add the mock back. In the methods you mentioned there is a `assertRejectedSealFetch` which asserts it. What else should I add?
```
  private def assertRejectedSealFetch(data: FetchPartitionData, expectedError: Errors): Unit = {
    assertEquals(expectedError, data.error)
    assertEquals(MemoryRecords.EMPTY, data.records)
    assertEquals(UnifiedLog.UNKNOWN_OFFSET, data.highWatermark)
    assertEquals(UnifiedLog.UNKNOWN_OFFSET, data.logStartOffset)
    verify(alterPartitionManager, never()).submit(any(), any())
  }
```

jeqo: Ah, missed that one. Looks good as-is.

jeqo (approved):

PR #698, opened 2026-07-14
test(inkless): extract shared TopicConfigTestUtils from duplicated test helpers

viktorsomogyi:
Extracts common test helper methods from Inkless tests and adds some hardening to these tests as there were recent flaky failures were caused by metadata propagation issues.

### DIFF
--- a/core/src/test/java/kafka/server/TopicConfigTestUtils.java
+++ b/core/src/test/java/kafka/server/TopicConfigTestUtils.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kafka.server;
+
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.common.config.ConfigResource;
+import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
+import org.apache.kafka.common.message.DescribeTopicPartitionsResponseData;
+import org.apache.kafka.test.NoRetryException;
+import org.apache.kafka.test.TestUtils;
+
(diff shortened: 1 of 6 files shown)


core/src/test/java/kafka/server/TopicConfigTestUtils.java
```diff
@@ -0,0 +1,150 @@
...
+ */
+public final class TopicConfigTestUtils {
+
+    private TopicConfigTestUtils() {
+    }
+
+    /**
+     * Read all config entries for a topic via {@code describeConfigs}. Retries up to 3 times
+     * (1s delay) on {@link UnknownTopicOrPartitionException} to ride out metadata propagation
+     * delays after topic creation or config alteration.
```
EelisK: Doesn't this method only wait for topic creation? Config alteration shouldn't result in `UnknownTopicOrPartitionException` 🤔 
```suggestion
     * delays after topic creation.
```

core/src/test/java/kafka/server/TopicMetadataProbe.java
EelisK, line 1: nit: I think it's good to think twice before introducing utility classes because often they end up as a catch-all with functionality that is difficult to discover 😄 

In this case it's probably okay because it's part of the test suite, but if you are open to it, I would consider renaming it to something like `TopicMetadataAwaiter` to make it clear what purpose this class serves.

EelisK (dismissed):
Solid cleanup 🧽 left a couple cosmetic suggestions but feel free to ignore

core/src/test/java/kafka/server/TopicConfigTestUtils.java
viktorsomogyi: Ah, indeed. Removed it.

core/src/test/java/kafka/server/TopicMetadataProbe.java
viktorsomogyi: Yea, unfortunately it's not the perfect solution. What do you think about just the `TopicMetadata` name? I wanted to avoid `TopicMetadataAwaiter` because we have 2 methods that don't really do waiting in the same sense as the others. Also, we could shorten the method names:
```
public final class TopicMetadata {
    static Map<String,String> configs(Admin admin, String topic)                               // was getTopicConfig
    static String readValue(Admin admin, String topic, String key)                               // was getTopicConfigValue
    static void awaitValue(Admin admin, String topic, String key, String expected)    // was waitForTopicConfigValue
    static void awaitVisible(Admin admin, String topic)                                                  // was waitForTopicToExist
}
```

EelisK: This interface looks clean 👍  we could even add an appropriate verb at the end to make it clear it's not a data class. Something like `TopicMetadataProbe`?

viktorsomogyi: Ok, sounds good.

viktorsomogyi: Done.

EelisK (approved):
LGTM 👍

PR #703, opened 2026-07-16
test(inkless): give each broker an exclusive TS fetch chunk cache dir

jeqo:
DiskChunkCache wipes its base directory on every configure() and promotes any IO error (including concurrent-delete races) to a fatal ConfigException. Both InklessConsolidatedDisklessReassignmentTest and InklessConsolidatedDisklessTopicsTest pointed every combined broker at a single fetch.chunk.cache.path set cluster-wide, so the brokers' startup wipes raced on that one directory and intermittently aborted KafkaClusterTestKit.startup with "Failed to reset cache directory".

Set fetch.chunk.cache.path per broker via per-server properties so each broker owns its cache directory.


EelisK (approved):
Good catch 👍

PR #704, opened 2026-07-16
feat(inkless:retention): track earliest batch timestamp per log

jeqo:
Adds `logs.earliest_batch_timestamp` and maintains it in the control-plane write paths, so a future retention check can decide "time retention deletes nothing" from one column instead of the current O(depth) reverse-aggregate scan over a partition's whole `batches` history. This PR only adds and maintains the column (plus tests and a benchmark); no reader consumes it yet, so retention behavior is unchanged. The reader-side short-circuit is a follow-up.

## The column

`logs.earliest_batch_timestamp` is the effective timestamp (per `batch_timestamp`: `batch_max_timestamp` for CreateTime, `log_append_timestamp` for LogAppendTime) of the batch at `log_start_offset` -- the oldest retained batch. `NULL` means "unknown, must scan"; pre-existing and freshly created logs start `NULL` (no migration-time backfill) and are populated lazily as writes touch each log.

## Maintenance

Maintained server-side under the existing `logs FOR UPDATE`, sourced from the `batches` row (so it matches coalesced rows), via a single indexed lookup of the oldest batch:

- `delete_records_v1`: recompute when a delete advances `log_start` (or `NULL` when the log empties); no advance leaves it untouched.
- `commit_file_v1` and `commit_file_v2`: populate only when the partition has no value yet (`IS NULL`) -- covers empty->non-empty and lazily backfills pre-existing `NULL` logs. Appends add newer batches at the head, so a populated log is untouched and steady-state appends stay free. Both are maintained because batch coalescing defaults off (`batch.coalescing.enabled=false`), making `commit_file_v1` the default path.
- `prune_batches_below_highest_tiered_offset_v1`: like delete, cross-tier prune removes the oldest batches and advances `log_start`, so it recomputes when it deleted something (or `NULL` when empty).

## Changes

- Migration `V20`: adds the nullable `timestamp_nullable_t` domain and the `earliest_batch_timestamp` column, and `CREATE OR REPLACE`s the four functions above in place (signatures unchanged; no new/dropped versions).
- Regenerated jOOQ sources (new column + domain, schema-version stamp bump).
- Updated existing control-plane tests for the new `LogsRecord` arity (with correct expected timestamps).
- `LogsEarliestBatchTimestampTest`: invariant across commit/delete/prune (empty->non-empty, CreateTime vs LogAppendTime, advance recompute, delete-all/prune-all -> `NULL`, no-op delete/prune, backfill of a pre-existing `NULL` log).
- `EarliestBatchTimestampMaintenanceBenchmarkTest` (`@Tag("benchmark")`, excluded from CI).

## Compatibility / rolling upgrade

Backward compatible; old and new brokers can coexist indefinitely.

- Additive nullable column, no default -> metadata-only `ADD COLUMN` (no rewrite/backfill), ignored by old broker code.
- Unchanged function signatures and response contracts (`CREATE OR REPLACE` bodies only), so no cross-version function mismatch.
- Maintenance is server-side, so any broker's writes keep the column correct without knowing it exists.
- No reader yet -> no behavioral divergence.

Only the usual migration-ordering discipline applies (schema stays ahead of not-yet-upgraded brokers). Broker-binary rollback is safe; the migration need not be rolled back for a downgrade.

## Benchmark

The benchmark is schema-agnostic (only commit/delete Job APIs, never the new column) and is committed first, so its parent is the pre-column baseline: the A/B is reproducible by checkout -- run `./gradlew :storage:inkless:benchmarkTest --tests "*EarliestBatchTimestampMaintenanceBenchmarkTest"` on the benchmark commit ("before") and the feature commit ("after"), then diff `ms/call`.

Representative run (16 partitions/call, 500 calls/shape, wall-clock vs a local Postgres container; indicative):

| Shape                               | before (main) | after (branch) |
| ----------------------------------- | ------------- | -------------- |
| steady-commit (append, guard skips) |  5.07 ms/call |   4.66 ms/call |
| commit-into-empty (recompute fires) |  4.07 ms/call |   3.78 ms/call |
| delete-advance (recompute fires)    |  2.87 ms/call |   2.41 ms/call |

The gap is within run-to-run variance (delete-advance spanned 1.9-2.4 across branch runs; "after" came out equal-or-lower), so there is no measurable regression: the steady-state append hot path is untouched and the recompute paths stay flat rather than scaling with log depth.

## Scope / follow-ups

- Out of scope: the retention reader that consumes the column to short-circuit the scan (separate PR, so its own benchmark stays a clean A/B).
- Independent of the in-flight retention lock-decoupling work; can land alongside it.

## Testing

`./gradlew :storage:inkless:test --tests "io.aiven.inkless.control_plane.postgres.*"` is green; the benchmark runs via the opt-in `benchmarkTest` task.


ivanyu (approved):
LGTM. Do you think it makes sense to mirror this to the in-memory implementation?

jeqo:
@ivanyu good idea, can do on the follow-up https://github.com/aiven/inkless/pull/705

PR #705, opened 2026-07-16
fix(inkless:retention): decouple enforce_retention_v2 from the log lock

jeqo:
`enforce_retention_v2` took `logs FOR UPDATE` before its O(depth) retention boundary scan and looped every request in a single transaction. A concurrent `commit_file` on a locked partition therefore waited for the whole enforcement pass, and in production the commit-latency tail tracked the enforce latency (commit max was roughly equal to enforce max).

## Change

Enforcement is restructured so the lock is held only for the short delete, not the scan:

- The retention boundary is computed in a read-only step with no `FOR UPDATE`. `byte_size` is read inside that scan statement (a `selected_log` CTE) so the reverse-aggregate stays consistent with the `batches` rows under READ COMMITTED. The lock is taken only afterwards, for the delete.
- `EnforceRetentionJob` runs one transaction per request instead of wrapping all requests in one. A plpgsql function cannot release a row lock before it returns, so per-request transactions are what actually free the lock between partitions rather than accumulating it until the end. `now` is computed once per cycle and shared. Each transaction holds at most one partition lock, so enforce can no longer self-deadlock.
- Deleted batches/bytes are recounted under the lock so the reported counts stay accurate when a concurrent delete advances `log_start` between the unlocked scan and the locked delete.
- The `max_batches_per_request` cap probes for the `(cap+1)`-th deletable batch with `LIMIT/OFFSET` instead of `COUNT(*)` over all deletable rows, keeping the locked section O(delete-size) rather than O(depth) when a small cap meets a large deletable set.
- The boundary scan is skipped entirely when the oldest retained batch provably survives every enabled policy, using `logs.byte_size` and `logs.earliest_batch_timestamp`. A `NULL` `earliest_batch_timestamp` means "unknown", so the short-circuit does not fire and the scan runs.

## Safety

Computing the boundary unlocked is safe: retention deletes at the tail while commit appends at the head, and `delete_records_v1` re-clamps to the current `log_start` under its own lock, so a concurrent commit or delete cannot cause over-deletion. Partial progress across partitions is fine; retention is idempotent and re-runs next cycle.

## Results

Indicative local numbers (testcontainers, single machine) from the committed `@Tag("benchmark")` harness, `maxBatchesPerRequest=1` so delete work is tiny and the measured tail is lock coupling. `max/enforce` is the concurrent commit's max latency as a fraction of the enforce latency; `commits` is how many commits completed during one enforce pass.

Single partition, commit to the same partition being enforced:

| rows | before (max, commits) | after (max, commits) |
| --- | --- | --- |
| 200k | 147 ms, 1 | 13 ms, 52 |
| 400k | 312 ms, 1 | 44 ms, 121 |
| 800k | 731 ms, 1 | 18 ms, 285 |

Multi-partition (8 partitions enforced in one cycle), commit to the first partition:

| rows/partition | before (max, commits) | after (max, commits) |
| --- | --- | --- |
| 50k | 271 ms, 1 | 29 ms, 135 |
| 100k | 776 ms, 1 | 15 ms, 223 |

Commit p50/p95 during enforcement drop to roughly 1-5 ms, and commit max no longer scales with partition depth.

## Testing

- Full control-plane contract (`PostgresControlPlaneTest`, including the `Retention` size/time suites) and the in-memory contract remain green; the SQL semantics are unchanged.
- New PG concurrency tests in `EnforceRetentionJobTest`: log deleted between scan and delete returns `unknown_topic_or_partition`; `log_start` advanced between scan and delete returns the recounted deletion; and a `NULL` `earliest_batch_timestamp` falls back to the scan instead of short-circuiting.

## Stacking / dependencies

Stacked on the `find_batches` bounded-scan change and on the `logs.earliest_batch_timestamp` change (the short-circuit reads that column, so this must apply after it).

## Out of scope (follow-ups)

- Bound the boundary scan itself to O(result) with a reverse walk (the scan is now unlocked, so its cost is background-job latency, not commit blocking).
- `FOR UPDATE SKIP LOCKED` with a fairness backstop, if a residual enforce-vs-commit tail remains after this change.
- Leader-scheduled enforcement once managed replicas provide a per-partition owner.


ivanyu (approved):
Thank you, very nice!

PR #710, opened 2026-07-21
fix(inkless:consolidation): route DeleteRecords to the real leader and report cross-tier low watermark [KC-332]

viktorsomogyi:
For a switched consolidating diskless topic the metadata transformer advertises an AZ-selected replica (a follower) as the client-facing leader, so a DeleteRecords request looped on NOT_LEADER_OR_FOLLOWER and timed out. DisklessDeleteRecordsForwarder forwards the leader-only leg to the partition's real KRaft leader; KafkaApis fans the request out into local and forwarded legs and combines the results (KafkaApisBuilder / BrokerServer wire the forwarder).

Completion no longer waits on the ISR low watermark for consolidating partitions (their follower log starts are frozen at the switch). Instead ReplicaManager.deleteRecords advances the control-plane cross-tier earliest and reports it as the low watermark, so DeleteRecords and a subsequent ListOffsets(EARLIEST) agree on every broker. Adds the shared crossTierEarliestOffset / isConsolidatingDisklessPartition accessors.

### DIFF
--- a/core/src/main/scala/kafka/server/DisklessDeleteRecordsForwarder.scala
+++ b/core/src/main/scala/kafka/server/DisklessDeleteRecordsForwarder.scala
@@ -1,0 +1,243 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// Copyright (c) 2024 Aiven, Helsinki, Finland. https://aiven.io/
+
+package kafka.server
+
+import kafka.server.metadata.InklessMetadataView
+import kafka.utils.Logging
+import org.apache.kafka.clients.{ClientResponse, NetworkClient, RequestCompletionHandler}
+import org.apache.kafka.common.message.DeleteRecordsRequestData
+import org.apache.kafka.common.message.DeleteRecordsRequestData.{DeleteRecordsPartition, DeleteRecordsTopic}
+import org.apache.kafka.common.message.DeleteRecordsResponseData.DeleteRecordsPartitionResult
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.requests.{DeleteRecordsRequest, DeleteRecordsResponse}
+import org.apache.kafka.common.utils.Time
+import org.apache.kafka.common.{Node, TopicPartition}
+import org.apache.kafka.metadata.{MetadataCache, PartitionRegistration}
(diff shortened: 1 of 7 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1674,7 +1707,133 @@ class ReplicaManager(val config: KafkaConfig,
...
+    }
+
+    // A consolidating partition's converted delete offset X: the concrete requested offset, or -- for
+    // HIGH_WATERMARK, which on a consolidating topic always reaches into the WAL and thus runs the
+    // diskless leg -- the low watermark that leg returned (= log_start_offset = HWM).
+    val requests = new java.util.ArrayList[AdvanceCrossTierLogStartOffsetRequest]()
+    val partitionsInOrder = mutable.ArrayBuffer.empty[TopicPartition]
+    response.foreach { case (topicPartition, result) =>
+      if (result.errorCode == Errors.NONE.code &&
+        _inklessMetadataView.isConsolidatingDisklessTopic(topicPartition.topic)) {
```
ivanyu, line 1746: it feels without `else` we're silently skipping errors?

viktorsomogyi: In the caller (`finalizeCrossTierAndRespond`) we will do a `getOrElse(topicPartition, result)` where we default to the original result if the partition was skipped here because of an error or if it wasn't a consolidating diskless topic. So the original error will be propagated to the client.

core/src/main/scala/kafka/server/DisklessDeleteRecordsForwarder.scala
```diff
@@ -0,0 +1,243 @@
...
+    val now = time.milliseconds()
+    var pending = queue.poll()
+    while (pending != null) {
+      requests.add(buildRequest(pending, now))
+      pending = queue.poll()
+    }
+    requests
+  }
+
+  private def buildRequest(pending: PendingForward, now: Long): RequestAndCompletionHandler = {
```
ivanyu, line 177: We don't preserve the principal when we forward these requests. This may result in `TOPIC_AUTHORIZATION_FAILED` when the request is performed from the broker. AFAIU Kafka has the envelop mechanism for preserving the original principal

core/src/main/scala/kafka/server/ReplicaManager.scala
ivanyu: Gotcha, maybe a comment on this?

core/src/main/scala/kafka/server/DisklessDeleteRecordsForwarder.scala
viktorsomogyi: Well, we use this class in KafkaApis.handleDeleteRecordsRequest, which does an authorization, so it should be gated. Regarding the `TOPIC_AUTHORIZATION_FAILED` error I'm not sure how can we get this if we forward with the brokers own ID? Should we assume that broker principals may not have permission to delete records?

ivanyu: > Should we assume that broker principals may not have permission to delete records?

I assumed this :thinking: I mean, it's possible to configure brokers to have only `CLUSTER_ACTION`, for example, which won't allow `DeleteRecords`.

ivanyu: We can address this as a followup, though

viktorsomogyi: Yes, that's probably a bigger change, we can keep this PR to the minimal.

ivanyu (approved):

PR #711, opened 2026-07-21
fix(inkless:consolidation): stop over-reclaiming the classic prefix on rebuilt leaders [KC-332]

viktorsomogyi:
## Summary

Builds on the two merged consolidation fixes (#709 ListOffsets(EARLIEST) routing and #710 DeleteRecords forwarding). Those made the control plane the authoritative cross-tier earliest for reads and the DeleteRecords write path. This PR carries that same authority into remote-retention reclaim and the become-leader log-start report, so
a consolidating topic never deletes remote data that is still within the delete/retention boundary, and never reports a stale earliest.

## Background
After a classic topic switches to diskless, the classic prefix `[0, seal)` lives only in the remote tier. A broker that rebuilds the partition resumes its local log at the seal, so the whole-log `UnifiedLog.logStartOffset` is pinned at the seal by the tier-state rebuild and only ever increments. Two things then read that stale seal instead of the
true cross-tier earliest `X` that DeleteRecords/retention left in the control plane:
- the RemoteLogManager reclaim floor deleted the whole remote prefix `[X, seal)` and reported the seal as the earliest on becoming leader;
- a fetch in `[X, seal)` was rejected as out of range and the rebuild restarted at the seal. The reclaim leg is irreversible, so this is a data-loss path, not just a stale read.

## What changed
- **Whole-log start for the consolidation fetch/rebuild** (`DisklessLeaderEndPoint`): a consolidating partition now advertises `min(crossTierEarliest, localLog.logStartOffset)` as the whole-log start instead of the raw local seal, so a fetch in `[X, seal)` rebuilds from the remote tier and the rebuild restarts the log at `X`.
- **Reclaim floor and become-leader report** (`RemoteLogManager`, wired in `BrokerServer`): both now consult a control-plane override for consolidating partitions. On an empty override they fail safe to the remote earliest (`findLogStartOffset`), never the seal; classic/non-inkless topics keep the upstream broker-local behavior.
- **Raw remote-log-start accessor** (`ReplicaManager.crossTierRemoteLogStartOffset` + new `ControlPlane.getCrossTierLogStart`): the reclaim floor reads the raw `remote_log_start_offset`, null-aware, rather than `ListOffsets(EARLIEST)`. `EARLIEST` returns `COALESCE(remote_log_start_offset, log_start_offset)`, which falls back to the pruned WAL frontier when the remote start is unreported. Using that as the floor would delete still-live remote segments and lock the wrong value in through the forward-only advance. Returning empty here makes the RLM fail safe to the true remote earliest instead. `crossTierEarliestOffset` keeps the COALESCE fallback for reads, where the worst case is a transient, self-healing over-rejection.
- **New control-plane query and metric**: `GetCrossTierLogStartJob` (in-memory + Postgres) reads the raw column directly, with a `GetCrossTierLogStart` timer in `PostgresControlPlaneMetrics` and a `metrics.rst` entry.
- **Error handling**: the cross-tier accessors catch `Exception` rather than `Throwable`, so `Error`s propagate instead of silently failing open.

## Testing
- `RemoteLogManagerTest`: reclaim floor honored over the seal; fail-safe on empty override for consolidating vs. upstream behavior for classic; retention x DeleteRecords composition; and hardening tests proving the become-leader report resolves to the remote earliest (never the seal, never skipped), that `findLogStartOffset` returns the remote earliest despite a seal-pinned local start, and a combined bootstrap-plus-reclaim guard.
- `ReplicaManagerInklessTest` / `DisklessLeaderEndPointTest`: the accessors, the consolidating predicate, and the whole-log-start `min()` behavior.
- `AbstractControlPlaneTest`: `getCrossTierLogStart` is null-aware (returns empty for an unreported remote start) unlike `EARLIEST`.
- End-to-end coverage lives on the stacked e2e branch (`DeleteRecordsAcrossTiersTest`), which asserts the earliest settles at exactly the delete boundary on every broker with retention unset.

## Compatibility / risk
No behavior change for classic or non-inkless partitions: the override defaults to empty and the predicate to false, so the reclaim floor stays the broker-local log start. The control-plane advance is forward-only, and a stale cross-tier cache entry can only be too low, which under-reclaims and over-serves rather than losing data.

### DIFF
--- a/storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
+++ b/storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
@@ -154,6 +154,19 @@ public class RemoteLogManager implements Closeable, AsyncOffsetReader {
     private final Time time;
     private final Function<TopicPartition, Optional<UnifiedLog>> fetchLog;
     private final BiConsumer<TopicPartition, Long> updateRemoteLogStartOffset;
+    // Inkless: overrides the log start offset used to drive remote-retention reclaim (and the
+    // become-leader log-start report) for consolidating diskless partitions. For such partitions the
+    // broker-local UnifiedLog.logStartOffset can be pinned at the classic-to-diskless seal on a
+    // freshly-rebuilt leader, which would over-reclaim the remote classic prefix and report the seal as
+    // the cross-tier earliest. This function returns the control-plane cross-tier earliest for those
+    // partitions and OptionalLong.empty() for everything else (classic topics / non-inkless), where the
+    // upstream behavior (log.logStartOffset()) is unchanged.
+    private final Function<TopicPartition, OptionalLong> logStartOffsetOverride;
+    // Inkless: tells whether a partition is a consolidating diskless topic. Used to pick the reclaim
+    // floor's fallback when {@link #logStartOffsetOverride} is empty: consolidating partitions must never
+    // fall back to the broker-local log start (the classic-to-diskless seal on a rebuilt leader), while
+    // classic topics keep the upstream local-log-start behavior. No-op predicate for non-Inkless callers.
+    private final Predicate<TopicPartition> isConsolidatingDisklessPartition;
     private final BrokerTopicStats brokerTopicStats;
     private final Metrics metrics;
 
@@ -208,7 +221,6 @@ public class RemoteLogManager implements Closeable, AsyncOffsetReader {
      * @param brokerTopicStats BrokerTopicStats instance to update the respective metrics.
      * @param metrics  Metrics instance
      */
-    @SuppressWarnings({"this-escape"})
     public RemoteLogManager(RemoteLogManagerConfig rlmConfig,
                             int brokerId,
                             String logDir,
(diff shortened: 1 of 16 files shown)


storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
```diff
@@ -882,7 +919,13 @@ protected void execute(UnifiedLog log) throws InterruptedException {
...
         private void maybeUpdateLogStartOffsetOnBecomingLeader(UnifiedLog log) throws RemoteStorageException {
             if (!isLogStartOffsetUpdated) {
-                long logStartOffset = findLogStartOffset(topicIdPartition, log);
+                // For a consolidating diskless partition, prefer the control-plane cross-tier earliest so a
+                // freshly-elected leader does not report its local seal as the cross-tier earliest (which
+                // would push the broker-agnostic earliest up to the seal). No-op for classic topics.
+                OptionalLong override = logStartOffsetOverride.apply(topicIdPartition.topicPartition());
+                long logStartOffset = override.isPresent()
+                        ? override.getAsLong()
+                        : findLogStartOffset(topicIdPartition, log);
```
giuseppelillo, line 932: `ReplicaManager.crossTierEarliestOffset` returns `ListOffsets(EARLIEST) = COALESCE(remote_log_start_offset, log_start_offset)`. If `remote_log_start_offset` is NULL due to an error in reporting the value, the cross tier earliest offset returned is the pruned WAL frontier, so there's the chance that all data from remote gets deleted in case this happens.

Example:

1. Born-consolidated topic; records [0,1000) produced, [0,700) tiered to remote.
2. The CrossTierLogStartReporter never lands a successful flush for this partition — crash between enqueue and the 1s flush, a persistent partition-specific advanceCrossTierLogStartOffset failure, or an enqueue guard drop during post-switch metadata propagation. remote_log_start_offset stays NULL.
3. The 5-min WAL pruner advances log_start_offset to 700.
4. A leader runs become-leader / expiration. crossTierEarliestOffset returns COALESCE(NULL, 700) = 700 → present.
5. reclaimFloorLogStartOffset returns 700 → deletes remote [0,700), which was still within retention. And become-leader reports 700 → the reporter persists it via the forward-only advance_cross_tier_log_start_v1, so remote_log_start_offset is now permanently 700 and EARLIEST is permanently wrong.

viktorsomogyi: I think this is a good catch, let me fix it.

viktorsomogyi:
@giuseppelillo addressed the bug you discovered, PTAL.

viktorsomogyi:
Moving this back to draft until I resolve the test failures.

viktorsomogyi:
@giuseppelillo the tests should pass now, it's ready for review again.

giuseppelillo (approved):

PR #715, opened 2026-07-22
fix(inkless:storage): drain object-store deletes monotonically under throttling [KC-413]

jeqo:
When object storage throttled deletes, the WAL file cleaner aborted the whole cycle and re-attempted keys it had already deleted, looping without draining and amplifying the throttle. This makes deletion drain monotonically: only the keys the backend confirms deleted are dereferenced, and the rest are retried on the next cycle.

## Problem

`S3Storage.delete(Set)` threw on the first per-key or whole-request error. Since `FileCleaner` deletes from storage *before* updating the control plane, that throw dereferenced nothing for the whole batch — including keys S3 had already deleted in earlier slices. The next cycle re-issued `DeleteObjects` for objects that were already gone, so under sustained throttling the set never drained.

## Change

`ObjectDeleter.delete(Set)` now returns the subset of keys confirmed deleted (already-absent counts as deleted) instead of `void`, so callers can advance on partial progress. `FileCleaner` dereferences only that subset; undeleted keys stay marked and are retried by the next periodic cycle.

- **S3**: single pass per 1000-key batch, returning `DeleteObjectsResponse.deleted()` and logging the rest. No longer throws on partial or whole-request failure (including a 503 the SDK's adaptive retry exhausted).
- **Azure**: deletes blob-by-blob, so it now accumulates the confirmed-deleted subset instead of throwing on the first failure and abandoning the rest.
- **GCS**: stays all-or-nothing (documented inline) — `Storage.delete(Iterable)` reports batch failure as a thrown exception, so no confirmed subset can be extracted.
- **In-memory / config-test backends**: return the input set on success.

No control-plane change needed: `deleteFiles` already accepts any subset of keys.

## Observability

Partial progress is now the normal case, and nothing reported it: `FileCleanerErrorRate` counts thrown exceptions, which partial failure no longer raises, and `FileCleanerFilesRate` counts only what succeeded — so a cycle draining half its worklist looked identical to one draining all of it. New `FileCleanerFilesFailedRate` counts submitted minus confirmed; a sustained non-zero value means keys are being re-attempted every cycle rather than draining, which separates throttling (self-recovering) from a hard per-key failure.

## Testing

- `FileCleanerMockedTest`: partial drain, no progress, and storage failure, plus the failed-files count.
- New `S3StorageDeleteTest` (WireMock): partial return without retry, and a whole-request 503 that must not propagate.
- Azure/GCS covered by existing `BaseStorageTest.testDeletes` against Azurite / fake-gcs.

### DIFF
--- a/docs/inkless/metrics.rst
+++ b/docs/inkless/metrics.rst
@@ -236,14 +236,15 @@ FileCleaner metrics
 io.aiven.inkless.delete:type=FileCleaner
 ----------------------------------------
 
-=====================  =========================================================
-Attribute name         Description                                              
-=====================  =========================================================
-FileCleanerErrorRate   Total number of file cleaning errors                     
-FileCleanerFilesRate   Total number of files cleaned                            
-FileCleanerRate        Total number of file cleaning cycles started             
-FileCleanerTotalTime   Total time spent on a file cleaning cycle in milliseconds
-=====================  =========================================================
+===========================  =================================================================================================================================
+Attribute name               Description                                                                                                                      
+===========================  =================================================================================================================================
+FileCleanerErrorRate         Total number of file cleaning errors                                                                                             
+FileCleanerFilesFailedRate   Total number of files the storage backend did not confirm deleted; they stay marked for deletion and are retried on a later cycle
+FileCleanerFilesRate         Total number of files cleaned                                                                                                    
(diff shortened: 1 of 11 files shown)


tvainika (approved):
LGTM

PR #721, opened 2026-07-25
control_plane: retry transient postgres errors in control plane jobs

Mwea:
## Problem

A production cluster (`cdip-diskless-prod-spoke-*`) threw a burst of `FetchError` / `Read future failed` exceptions. Every trace bottoms out in the same PostgreSQL condition on the **read replica** that serves `find_batches_v2`, in two escalating forms:

1. **Statement cancellation**
   ```
   org.postgresql.util.PSQLException: ERROR: canceling statement due to conflict with recovery
     Detail: User was holding shared buffer pin for too long.
   ```
2. **Connection termination** (escalation of the same conflict)
   ```
   org.postgresql.util.PSQLException: FATAL: terminating connection due to conflict with recovery
     Detail: User was holding shared buffer pin for too long.
   ```
   which then surfaces downstream as `server conn crashed?`, `Cannot commit transaction`, and a suppressed `Connection is closed` / `Cannot get autoCommit`.

Both are **hot-standby recovery conflicts**: a read query holds a shared buffer pin while WAL replay needs the same page. They are transient and the server rolls the transaction back — but `JobUtils` had no retry (only a `// TODO add retry with backoff`), so a single conflict propagated all the way up as `ControlPlaneException("Error finding batches")` → `FindBatchesException` and `FetchHandler` returned `UNKNOWN_SERVER_ERROR` for every partition in the fetch.

> Note: the DB-side mitigations (`hot_standby_feedback = on`, raised `max_standby_streaming_delay`) are already enabled on this cluster. This PR is the application-side resilience for the conflicts that still occur under replay pressure.

## Change

Implement the missing retry in `JobUtils`: retry the whole job with jittered exponential backoff (3 attempts, ~50 ms → ~100 ms with 0.2 jitter, capped at 1 s; via the injected `Time` so it is fully virtual in tests).

**Exceptions handled** — retry is keyed on rollback-guaranteed PostgreSQL `SQLState`s:

| SQLState | Meaning | Why retriable |
|----------|---------|---------------|
| `40001` | `serialization_failure` | PostgreSQL raises this (`ERRCODE_T_R_SERIALIZATION_FAILURE`) for **both** recovery-conflict forms above — the statement cancellation *and* the connection termination. Server guarantees rollback. |
| `40P01` | `deadlock_detected` | Victim transaction is rolled back. |

Key design points:

- **Retry wraps the whole transaction**, not the inner lambda. Because the control-plane `DSLContext`s are Hikari-pool-backed, a retry pulls a **fresh connection** — this is what recovers from the connection-terminated form where the previous connection is dead.
- **Full cause chain + suppressed throwables are scanned.** The retriable `PSQLException` is buried under `ControlPlaneException` → `DataAccessException`; in the connection-killed form it appears as a **suppressed** throwable (the failed rollback), exactly as in the traces.
- **Generic connection-loss states (`08*`, `57P01`) are intentionally NOT retried** — a drop *during commit* leaves a write in-doubt, so blindly replaying could double-apply. The recovery-conflict family that caused this incident is fully covered by the rollback-guaranteed states, so the retry is safe for both read and write jobs and can live in the shared helper.
- **Error contract unchanged**: on exhaustion or a non-retriable failure, the original exception is rethrown and mapped to `ControlPlaneException` / wrapped `RuntimeException` as before.

### Hot-path safety

This retry sits on the fetch hot path shared by all 16 control-plane jobs, so it is designed to stay safe under a conflict storm rather than amplify one:

- **Metrics stay accurate.** `measureDurationMs` fires the duration callback in a `finally`, so timing each attempt directly would record N samples per run and pollute the latency/rate percentiles with failed-attempt durations. Only the **decisive** attempt (success, or the final failure) is reported to the callback → exactly one sample per `run()`.
- **Jittered backoff** (`ExponentialBackoff`, reused from `clients`) prevents many jobs failing against the same struggling standby from retrying in lockstep and re-colliding on the same recovery-conflict window.
- **Interrupt-aware.** `Utils.sleep` restores the interrupt flag but returns normally; the loop checks `Thread.isInterrupted()` after backoff and aborts (keeping the flag set) rather than spinning through remaining attempts and delaying broker shutdown.
- **One terminal log** on exhaustion, so the final user-visible failure is greppable instead of appearing only as intermediate "retrying" warnings.

## Tests

`JobUtilsTest` (pure unit — no DB container, uses `MockTime`): happy path, recovery-conflict-then-succeed (reproduces the incident with the exact wrapping from the traces), deadlock retry, max-attempts exhaustion preserving `ControlPlaneException`, non-retriable SQLState (`23505`) and plain runtime exceptions not retried, suppressed-throwable scanning, self-referential cause chain (terminates, no `StackOverflow`), plus: jittered backoff timing asserted via `MockTime` deltas, a single duration-callback per run (guards the metrics contract), no sleep on non-retriable failures, and interrupt-driven abort.

## Follow-up (not in this PR)

- When retries are exhausted the client still gets `UNKNOWN_SERVER_ERROR`. Surfacing exhausted transient failures as a *retriable* Kafka error would be a separate change in the `FetchHandler` error-mapping path.
- The connection-terminated form relies on pgjdbc surfacing SQLState `40001`; there is a residual driver-level race where a socket EOF beating the FATAL message could downgrade to `08006`. Confirming this end-to-end warrants a hot-standby integration test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
@@ -45,34 +45,14 @@ public class JobUtils {
     static final int BACKOFF_MULTIPLIER = 2;
     static final double BACKOFF_JITTER = 0.2;
 
-    /**
-     * Backoff between retries. Uses jitter (a randomized factor in {@code [1 - jitter, 1 + jitter]})
-     * so that many jobs failing against the same struggling standby at once do not retry in lockstep
-     * (thundering herd) and re-collide on the same recovery-conflict window.
-     */
+    // Jittered so concurrent retries against the same standby do not collide in lockstep.
     private static final ExponentialBackoff BACKOFF =
         new ExponentialBackoff(INITIAL_BACKOFF_MS, BACKOFF_MULTIPLIER, MAX_BACKOFF_MS, BACKOFF_JITTER);
 
     /**
-     * PostgreSQL {@code SQLState}s that are safe to retry because they guarantee the transaction did
-     * <em>not</em> commit (the server rolled it back), so replaying the whole transaction cannot
-     * double-apply a write.
-     *
-     * <ul>
-     *   <li>{@code 40001} — {@code serialization_failure}. PostgreSQL also raises this
-     *       ({@code ERRCODE_T_R_SERIALIZATION_FAILURE}) for hot-standby recovery conflicts, in both
-     *       observed forms: the statement cancellation
-     *       ("canceling statement due to conflict with recovery") <em>and</em> the connection
-     *       termination ("terminating connection due to conflict with recovery"). The latter closes
-     *       the JDBC connection, but because {@code readJooqCtx}/{@code writeJooqCtx} are backed by a
-     *       connection pool, the retry transparently obtains a fresh connection.</li>
-     *   <li>{@code 40P01} — {@code deadlock_detected}. The victim transaction is rolled back.</li>
-     * </ul>
-     *
-     * <p>We deliberately do <b>not</b> retry generic connection-loss states (class {@code 08*},
-     * {@code 57P01}) here: a connection dropped <em>during commit</em> leaves the transaction
-     * in-doubt, so blindly replaying a write job could apply it twice. The recovery-conflict family
-     * that motivated this retry logic is fully covered by the rollback-guaranteed states above.
+     * PostgreSQL failures that guarantee transaction rollback and are safe to retry:
+     * {@code 40001} for serialization or recovery conflicts, and {@code 40P01} for deadlocks.
+     * Connection-loss states are excluded because the commit outcome may be unknown.
      */
     private static final Set<String> RETRIABLE_SQL_STATES = Set.of("40001", "40P01");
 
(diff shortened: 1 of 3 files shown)


storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
```diff
@@ -52,29 +95,82 @@ public static void run(final Runnable runnable, final Time time, final Consumer<
...
+                if (attempt >= MAX_ATTEMPTS || !isRetriable(e)) {
+                    // Decisive failure: record its duration once, then propagate unchanged.
+                    durationCallback.accept(lastAttemptDurationMs[0]);
+                    if (attempt > 1) {
+                        LOGGER.warn("Giving up after {} attempts on transient database error", attempt, e);
+                    }
+                    throw e;
+                }
+                final long backoffMs = BACKOFF.backoff(attempt - 1);
+                LOGGER.warn("Transient database error on attempt {}/{}, retrying in {} ms",
```
viktorsomogyi: If it's a transient error, maybe the warn level is a bit too harsh. I think this ideally should be debug or maybe info level. The final attempt is fair to be warn though.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
```diff
@@ -17,33 +17,76 @@
...
+     *       the JDBC connection, but because {@code readJooqCtx}/{@code writeJooqCtx} are backed by a
+     *       connection pool, the retry transparently obtains a fresh connection.</li>
+     *   <li>{@code 40P01} — {@code deadlock_detected}. The victim transaction is rolled back.</li>
+     * </ul>
+     *
+     * <p>We deliberately do <b>not</b> retry generic connection-loss states (class {@code 08*},
+     * {@code 57P01}) here: a connection dropped <em>during commit</em> leaves the transaction
+     * in-doubt, so blindly replaying a write job could apply it twice. The recovery-conflict family
+     * that motivated this retry logic is fully covered by the rollback-guaranteed states above.
+     */
```
viktorsomogyi, line 56: The comment seems to be a bit too verbose. Perhaps:
```suggestion
    /**
     * PostgreSQL failures that guarantee transaction rollback and are safe to retry:
     * {@code 40001} for serialization or recovery conflicts, and {@code 40P01} for deadlocks.
     * Connection-loss states are excluded because the commit outcome may be unknown.
     */
```

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
```diff
@@ -52,29 +95,82 @@ public static void run(final Runnable runnable, final Time time, final Consumer<
...
+    private static boolean hasRetriableCause(final Throwable throwable, final Set<Throwable> seen) {
+        if (throwable == null || !seen.add(throwable)) {
+            return false;
+        }
+        if (throwable instanceof SQLException) {
+            final String sqlState = ((SQLException) throwable).getSQLState();
+            if (sqlState != null && RETRIABLE_SQL_STATES.contains(sqlState)) {
+                return true;
+            }
+        }
```
viktorsomogyi, line 130: Would it make sense to take a look at the whole chain by iterating `getNextException()`?

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
```diff
@@ -52,29 +95,82 @@ public static void run(final Runnable runnable, final Time time, final Consumer<
...
+        // when we stop retrying.
+        final long[] lastAttemptDurationMs = {0L};
+        final Consumer<Long> captureDuration = d -> lastAttemptDurationMs[0] = d;
+
+        for (int attempt = 1; ; attempt++) {
+            try {
+                final T result = TimeUtils.measureDurationMs(time, callable, captureDuration);
+                durationCallback.accept(lastAttemptDurationMs[0]);
+                return result;
+            } catch (final Exception e) {
```
viktorsomogyi: Can't we just catch `SQLException` here?

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
```diff
@@ -52,29 +95,82 @@ public static void run(final Runnable runnable, final Time time, final Consumer<
...
-                throw (ControlPlaneException) e;
-            } else {
-                throw new RuntimeException(e);
+    /**
+     * Returns {@code true} if any throwable in the cause chain (including suppressed throwables) is a
+     * {@link SQLException} whose {@code SQLState} is in {@link #RETRIABLE_SQL_STATES}. The retriable
+     * cause is typically wrapped several layers deep (e.g. {@code ControlPlaneException} ->
+     * {@code DataAccessException} -> {@code PSQLException}), so the whole tree is scanned.
+     */
+    static boolean isRetriable(final Throwable throwable) {
```
viktorsomogyi, line 112: The parameter could be narrowed down to `SQLException` based on its behavior and the javadoc on the function too.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
```diff
@@ -52,29 +95,82 @@ public static void run(final Runnable runnable, final Time time, final Consumer<
...
+    /**
+     * Execute {@code callable}, retrying with jittered exponential backoff on transient PostgreSQL
+     * failures (see {@link #RETRIABLE_SQL_STATES}). On exhaustion, or for any non-retriable failure,
+     * the original exception is rethrown so callers observe the unchanged error contract.
+     *
+     * <p>Only the <em>decisive</em> attempt (the one that succeeds, or the final failure) is reported
+     * to {@code durationCallback}. Intermediate retried attempts are timed internally but not
+     * forwarded, so failed-attempt durations do not inflate the latency/rate metrics — exactly when
+     * an operator relies on them during a conflict storm.
+     */
```
viktorsomogyi: I suggest to refrain from commenting private methods with extensive javadocs. The implementation can change and the docs get left behind.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
viktorsomogyi: didn't apply at the end

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/JobUtils.java
viktorsomogyi: didn't apply this either

storage/inkless/src/test/java/io/aiven/inkless/control_plane/postgres/JobUtilsTest.java
```diff
@@ -0,0 +1,284 @@
+/*
+ * Inkless
+ * Copyright (C) 2024 - 2025 Aiven OY
```
tvainika, line 3: year of AI :)

tvainika (dismissed):
LGTM

tvainika (approved):

PR #728, opened 2026-07-30
test(inkless:consolidation): adjust tests to address failures

viktorsomogyi:
Consolidation fetch requests segment.bytes minus max.message.bytes so one extra batch still fits the follower segment. Default max.message.bytes is 1048588 (1 MiB plus record overhead), so the old 1 MiB setting left a 1-byte fetch and starved the fetcher.

2 MiB leaves about 1 MiB of fetch budget. segment.ms still rolls on time.


tests/kafkatest/tests/inkless/consolidation_dependency_outage_test.py
```diff
@@ -91,7 +91,7 @@ def test_dependency_outage_during_produce(self, metadata_quorum, dependency):
                         "diskless.enable": "true",
                         "remote.storage.enable": "true",
                         "min.insync.replicas": 2,
-                        "segment.bytes": 1048576,
+                        "segment.bytes": 2097152,
```
jeqo: Non-blocker: Wonder if we can get the same effect by lowering `max.message.bytes` instead, so segments keep rolling at the current cadence and the retention/prune paths stay as exercised as before. `512 KiB` on a `1 MiB` segment leaves a `512 KiB` fetch budget (two fetches per segment), which should be plenty for these datasets.

Batching is safe: `VerifiableProducer` only overrides `request.timeout.ms`, so `batch.size` stays at the 16 KiB default -- well inside the headroom.

```suggestion
                        "segment.bytes": 1024 * 1024,
                        "max.message.bytes": 512 * 1024,
```

jeqo (dismissed):
LGTM, just a suggestion on leveraging max.message.bytes instead of `segment.bytes` -- either one should work though.

tests/kafkatest/tests/inkless/consolidation_dependency_outage_test.py
viktorsomogyi: I think we can make this change, almost finished rerunning the tests.

jeqo (approved):

PR #730, opened 2026-07-30
fix(inkless:release): gate finalize-release on event_name != push

jeqo:
finalize-release skipped on every workflow_call from inkless-release.yml, so no GitHub Release was created (run 30553955976, v0.46). Its guard tested github.event_name == 'workflow_call', but a reusable workflow inherits the caller's event name (workflow_dispatch here), never 'workflow_call'. Gate on event_name != 'push' instead, matching the prepare job's push/call split.


jeqo:
Closing in favor of https://github.com/aiven/inkless/pull/729

PR #737, opened 2026-08-10
fix(inkless:switch): preserve classic prefix on partition increase during switch [KC-387]

viktorsomogyi:
A CreatePartitions committed during a classic-to-diskless switch created the switching partition's control-plane logs row at high_watermark=0. That made diskless LATEST report 0, so the consolidation fetcher truncated the classic
prefix and lost the pre-switch records.

The create path now snapshots each topic's ID and partition count from one metadata image before the mutation. A matching post-mutation ID creates only [priorCount, count). A missing or changed ID creates [0, count), with overlap
resolved by the control-plane conflict handling. Validation-only requests skip this work.

init_diskless_log_v1 now locks existing target rows in topic-partition order, replaces an empty placeholder with a positive seal, and replaces producer state when that update applies. The in-memory control plane mirrors this behavior.

SQL evolution from V11 to V23:
```diff
--- V11__Init_diskless_log.sql
+++ V23__Init_diskless_log_authoritative_seal.sql
@@
 DECLARE
     l_request RECORD;
     l_producer_state RECORD;
-    l_inserted BOOLEAN;
 BEGIN
+    -- Every caller takes existing row locks in the same order; responses still follow arg_requests order.
+    PERFORM 1
+    FROM logs
+    JOIN unnest(arg_requests) AS request
+        ON logs.topic_id = request.topic_id
+        AND logs.partition = request.partition
+    ORDER BY logs.topic_id, logs.partition
+    FOR UPDATE OF logs;
+
     FOR l_request IN
         SELECT *
         FROM unnest(arg_requests)
@@
         INSERT INTO logs (topic_id, partition, topic_name, log_start_offset, high_watermark, byte_size, diskless_start_offset)
         VALUES (l_request.topic_id, l_request.partition, l_request.topic_name,
                 l_request.log_start_offset, l_request.diskless_start_offset, 0, l_request.diskless_start_offset)
-        ON CONFLICT (topic_id, partition) DO NOTHING;
-
-        l_inserted := FOUND;
+        ON CONFLICT (topic_id, partition) DO UPDATE
+            SET log_start_offset      = EXCLUDED.log_start_offset,
+                high_watermark        = EXCLUDED.high_watermark,
+                diskless_start_offset = EXCLUDED.diskless_start_offset
+            WHERE logs.high_watermark = 0
+                AND logs.diskless_start_offset = 0
+                AND logs.byte_size = 0
+                -- A zero-offset re-init would not advance the placeholder, so leaving it alone keeps
+                -- re-initing an existing empty partition a no-op.
+                AND EXCLUDED.high_watermark > 0;

-        IF NOT l_inserted THEN
+        IF NOT FOUND THEN
             RETURN NEXT (l_request.topic_id, l_request.partition, 'already_initialized')::init_diskless_log_response_v1;
             CONTINUE;
         END IF;

+        -- The update path can reach a row an earlier init already wrote producer state for, and
+        -- producer_state is keyed on a BIGSERIAL row_id, so those rows would accumulate rather than be
+        -- replaced. The incoming snapshot is the authority for the partition.
+        DELETE FROM producer_state
+        WHERE topic_id = l_request.topic_id
+            AND partition = l_request.partition;
+
         FOR l_producer_state IN
             SELECT *
             FROM unnest(arg_producer_states)
```
Tests cover both control-plane backends, exact controller ranges, topic recreation, topics appearing during a request, validation-only behavior, and the end-to-end partition-increase switch path.

Co-authored-by: Cursor <cursoragent@cursor.com>


storage/inkless/src/test/java/io/aiven/inkless/control_plane/CreateTopicAndPartitionsRequestTest.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Inkless
+ * Copyright (C) 2024 - 2025 Aiven OY
```
jeqo, line 3: nit: 
```suggestion
 * Copyright (C) 2024 - 2026 Aiven OY
```

core/src/main/scala/kafka/server/ControllerApis.scala
```diff
@@ -991,7 +1006,14 @@ class ControllerApis(
...
             } else {
-              Some(new CreateTopicAndPartitionsRequest(topicIdOrError.result(), req.name(), req.count()))
+              val topicId = topicIdOrError.result()
+              // The cached range is only usable when it belongs to the topic the controller just mutated.
+              // Otherwise create the full range and rely on init_diskless_log_v1 to resolve overlap (KC-387).
+              val firstPartition = priorTopicStates.get(topicName) match {
+                case Some(state) if state.topicId == topicId => math.min(state.numPartitions, req.count())
+                case _ => 0
+              }
+              Some(new CreateTopicAndPartitionsRequest(topicId, topicName, firstPartition, req.count()))
```
jeqo, line 1016: Note: Claude found an interesting side effect of this: As Controller commits the metadata mutation before Control Plane call: If Control Plane fails to write, nothing reconciles this gap, and these diskless partition will not have rows and fail with `UNKNOWN_TOPIC_OR_PARTITION`. Also, `repair_diskless_log_v1` will not be able to help here. 

Before this PR, if a _larger_ increase on the same topic would be attempted, it would be healed by inserting `[0, count)`. With this PR, it would insert `[N, count)` missing the failed partitions in between.

A _same-count_ retry never reached the control plane either way, then nothing was really depending on this so I don't think we need to fix this here. 
Let's have a follow up to reconcile this. Maybe let's also mention this on the commit?

storage/inkless/src/main/java/io/aiven/inkless/control_plane/postgres/TopicsAndPartitionsCreateJob.java
```diff
@@ -71,14 +71,16 @@ private void runOnce() {
...
-                for (int partition = 0; partition < request.numPartitions(); partition++) {
+                for (int partition = request.firstPartition(); partition < request.numPartitions(); partition++) {
                     insertStep = insertStep.values(request.topicId(), partition, request.topicName(), 0L, 0L, 0L);
                 }
             }
             final int rowsInserted = insertStep.onConflictDoNothing().execute();
 
             // This is not expected to happen, but checking just in case.
-            final int maxInserts = requests.stream().mapToInt(CreateTopicAndPartitionsRequest::numPartitions).sum();
+            final int maxInserts = requests.stream()
```
jeqo: With the changes here, a non-empty request can contribute zero values, so we can return early now that `maxInserts` can be zero

storage/inkless/src/main/java/io/aiven/inkless/control_plane/CreateTopicAndPartitionsRequest.java
```diff
@@ -19,7 +19,33 @@
...
 import org.apache.kafka.common.Uuid;
 
+/**
+ * Request to create control-plane {@code logs} rows for a topic's partitions.
+ *
+ * <p>{@code numPartitions} is the topic's partition count once the operation completes; rows are created
+ * for {@code [firstPartition, numPartitions)}. Topic creation starts at 0, while a partition-count
+ * increase starts at the previous count so it does not insert an empty row over a partition that is
+ * concurrently switching from classic to diskless (KC-387, see
+ * {@code V23__Init_diskless_log_authoritative_seal.sql}).
```
jeqo: Let's maybe reword here to account for the count coming from a metadata image that can lag from the controller, so this is best-effort. V23 header already words it that way: "best-effort first layer… this guard is what actually decides the outcome"

core/src/test/scala/unit/kafka/server/ControllerApisTest.scala
```diff
@@ -1317,6 +1440,7 @@ class ControllerApisTest {
 
   @AfterEach
   def tearDown(): Unit = {
+    metadataCache.setImage(MetadataImage.EMPTY)
```
jeqo: nit: Another non-obvious Claude catch: this per-instance val with no `@TestInstance(PER_CLASS)`, so JUnit constructs a fresh `KRaftMetadataCache` per test,  and a fresh one already holds MetadataImage.EMPTY -- so this setImage is inert (same on :1096).

Tested locally and seems to be right.

storage/inkless/src/test/java/io/aiven/inkless/control_plane/AbstractControlPlaneTest.java
```diff
@@ -1957,6 +1957,61 @@ void existingTopicReturnAlreadyInitialized() {
...
+            assertThat(responses).containsExactly(InitDisklessLogResponse.success());
+
+            assertThat(controlPlane.getLogInfo(List.of(new GetLogInfoRequest(NEW_TOPIC_ID, 0))))
+                .containsExactly(GetLogInfoResponse.success(0, seal, seal, 0));
+
+            assertThat(controlPlane.listOffsets(List.of(new ListOffsetsRequest(switchingPartition, LATEST_TIMESTAMP))))
+                .map(ListOffsetsResponse::offset)
+                .containsExactly(seal);
+        }
+
```
jeqo, line 2014: We may be missing a test scenario here where `WHERE logs.high_watermark = 0 AND logs.diskless_start_offset = 0 AND logs.byte_size = 0` stops the init from affecting a live partition. Here's a draft:

```java
@Test
void initMustNotOverwriteAPartitionThatCarriesData() {
    controlPlane.initDisklessLog(List.of(
        new InitDisklessLogRequest(NEW_TOPIC_ID, NEW_TOPIC, 0, 0, 50, List.of())
    ));
    controlPlane.commitFile("a", ObjectFormat.WRITE_AHEAD_MULTI_SEGMENT, BROKER_ID, FILE_SIZE, List.of(
        CommitBatchRequest.of(0, new TopicIdPartition(NEW_TOPIC_ID, 0, NEW_TOPIC),
            1, 10, 1, 10, 1000, TimestampType.CREATE_TIME)
    ));
    final var before = controlPlane.getLogInfo(List.of(new GetLogInfoRequest(NEW_TOPIC_ID, 0)));

    // A divergent, larger seal must not win once the row carries diskless data.
    final var responses = controlPlane.initDisklessLog(List.of(
        new InitDisklessLogRequest(NEW_TOPIC_ID, NEW_TOPIC, 0, 0, 999, List.of())
    ));

    assertThat(responses).containsExactly(InitDisklessLogResponse.alreadyInitialized());
    assertThat(controlPlane.getLogInfo(List.of(new GetLogInfoRequest(NEW_TOPIC_ID, 0))))
        .isEqualTo(before);
}
```

jeqo (commented):
Looks good to me overall. Few minor findings

core/src/main/scala/kafka/server/ControllerApis.scala
viktorsomogyi: This is a good point, will create a follow-up fix. I was thinking in some kind of reconciliation loop: take the controller metadata as the ground truth, then query the control plane for log row and reconcile where needed (with retry and backoff).

core/src/test/scala/unit/kafka/server/ControllerApisTest.scala
viktorsomogyi: Removed the redundant setImage instead of adding `PER_CLASS` as removing seems to be a more narrow change with less possible side effects.

jeqo (approved):

PR #738, opened 2026-08-11
fix(inkless:consume): stop counting a fast cache load as a cache hit

jeqo:
The hot path decided cache hit vs miss from primary.isDone() right after computeIfAbsent, on the premise that a miss cannot be done yet because the load runs on the fetch executor. The premise does not hold: the load is a supplyAsync on that executor and can finish before the calling thread reaches isDone(), which is likelier the smaller the range, the faster the object store, and the busier the broker. Those misses were counted as cache hits, so CacheHitBytes overstated cache effectiveness on exactly the fetches that are cheapest to serve.

Track whether this caller ran the load instead. Running it means a miss whatever isDone() reports by then; only when it did not run does isDone() separate an already-cached value from coalescing onto another caller's in-flight load, which stays a miss.

This is also what was failing `recentDataUsesRecentExecutorWithoutRateLimit` on CI. Since #732 that test waits for the asynchronous byte accounting, so it observes the mislabeled hit reliably instead of racing past it; the de-flake exposed this bug rather than causing it.

The new tests pin the classification deterministically by running the load on a same-thread executor, so a first load always completes before the caller observes the future - the ordering CI reached by chance. One of them asserts that coalescing onto an in-flight load is still a miss, which is what fails if the fix drops the isDone() check.

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
@@ -259,20 +259,21 @@ public class FetchPlanner implements Supplier<List<FetchPlanner.FetchRequestWith
             // doesn't run, so this flag stays false — but with per-key hedge dedup (hedgeGuards),
             // at most one hedge fires per primary regardless of how many callers have stale TTFB flags.
             final AtomicBoolean firstByteReceived = new AtomicBoolean(false);
+            final AtomicBoolean loadRan = new AtomicBoolean(false);
             final CompletableFuture<FileExtent> primary = cache.computeIfAbsent(
                 request.toCacheKey(),
                 k -> {
+                    loadRan.set(true);
                     final FileExtent fileExtent = fetchFileExtent(objectFetcher, request, firstByteReceived);
                     metrics.recordStorageBytesIn(fileExtent.data().length);
                     return fileExtent;
                 },
                 fetchDataExecutor
             );
-            // A true cache hit returns an already-completed future; the loader runs asynchronously on
-            // fetchDataExecutor, so a cache miss (whether this caller ran the loader or coalesced onto an
-            // in-flight load) is never done at this point. Deciding the hit from isDone() here avoids
-            // counting a coalesced in-flight miss as cache-hit bytes.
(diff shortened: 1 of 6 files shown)


storage/inkless/src/main/java/io/aiven/inkless/consume/FetchPlanner.java
```diff
@@ -259,20 +259,21 @@ private CompletableFuture<FileExtent> submitSingleRequest(final ObjectFetchReque
...
             );
-            // A true cache hit returns an already-completed future; the loader runs asynchronously on
-            // fetchDataExecutor, so a cache miss (whether this caller ran the loader or coalesced onto an
-            // in-flight load) is never done at this point. Deciding the hit from isDone() here avoids
-            // counting a coalesced in-flight miss as cache-hit bytes.
-            final boolean alreadyCached = primary.isDone();
+            // This caller running the load means a miss whatever isDone() reports by now: the load may already
+            // have finished on the fetch executor. Only when it did not run does isDone() separate a value that
+            // was already cached from coalescing onto another caller's in-flight load.
+            final boolean alreadyCached = !loadRan.get() && primary.isDone();
```
viktorsomogyi: What if we read `loadRan == false` because it's not evaluated yet, then pause, let the executor finish the load (which flips `loadRan` to `true`), but since it's already evaluated we see that `isDone() == true`? This way we categorize a cache miss as a hit.

jeqo: Good catch! `loadRan` was set inside the load body, which runs on `fetchDataExecutor`, so a caller could read it as `false` before the load started and then see `isDone() == true` after that same load completed.

Fixed in `221c43ba`: `computeIfAbsent` now takes an `onLoadStarted` callback that `CaffeineCache` runs inside Caffeine's mapping function -- on the calling thread, before the call returns, and only on a miss. `isDone()` still separates an already-cached value from a coalesced in-flight load, but neither read can be stale now.

New `CaffeineCacheTest` cases validate that contract: the callback fires before `computeIfAbsent` returns even with a load executor that never runs the load, and does not fire for a coalesced or already-cached lookup.

viktorsomogyi (approved):

PR #743, opened 2026-08-11
feat(inkless:retention): bound the file-cleaner deletion worklist [KC-404]

jeqo:
FileCleaner asked the control plane for every file in state `deleting` and filtered the grace period on the broker, so one cycle cost was proportional to the whole deletion backlog: no `LIMIT`, no fetch size, the full result set in broker heap while holding a jobs-pool connection and pinning a snapshot against vacuum, per broker, every cycle. Bounding it is only safe because the grace-period cutoff moves into the query too: with the age filter server-side every returned row is actionable, so a truncated worklist still drains.

## Commits

1. `bound the file-cleaner deletion worklist` - `getFilesToDelete(markedBefore, limit)` replaces the no-arg form; both control planes enforce both bounds; adds `file.cleaner.max.files.per.cycle`.
2. `count saturated file-cleaning cycles` - adds `FileCleanerCycleSaturatedRate` so a backlog held up by the cap can be told apart from one held up by failing deletes.
3. `index the file-cleaner grace-period predicate` - adds a partial index on `marked_for_deletion_at` so the cap bounds scan work, not just rows returned.
4. `flag schema changes in the changelog generator` - migrations added in a range now show up in the changelog with the blocking DDL they run, so an upgrade lock is not discovered by an operator after the fact.

See the commit messages for the reasoning behind each step.

## Config

- `file.cleaner.max.files.per.cycle` (int, default `20000`, `0` = unbounded). Caps the worklist and doubles as the delete pacing knob: per-broker drain rate is the cap over `file.cleaner.interval.ms`. Set it above the rate at which files are marked, otherwise the backlog grows.

## Metrics

- `FileCleanerCycleSaturatedRate` (new) - cycles whose worklist came back over the cap. The query asks for one row beyond the cap, so an exactly-full cycle is not misreported as saturated.

## Schema

`V25__File_cleaner_marked_for_deletion_index.sql` creates `files_by_marked_for_deletion_deleting_idx ON files (marked_for_deletion_at) WHERE state = 'deleting'` and drops `files_by_state_only_deleting_idx`, which covered `state` only and left the grace-period predicate as a residual filter (so `LIMIT` bounded rows returned but not scan work: a backlog entirely inside the grace period was walked in full to return zero rows).

**Operator action recommended.** The index is small - it holds only `deleting` rows - but building it scans all of `files` to evaluate the predicate, and Flyway runs migrations in a transaction, so `CONCURRENTLY` is unavailable and the build holds `ACCESS EXCLUSIVE`. Commits to `files` block for that scan, which scales with the whole table rather than with the `deleting` subset, and `files` is on the produce commit path. Both statements are idempotent (`IF NOT EXISTS` / `IF EXISTS`), so the lock can be avoided entirely by running the DDL concurrently before upgrading, which makes the migration a no-op:

```sql
SET LOCAL lock_timeout = '5s';
CREATE INDEX CONCURRENTLY IF NOT EXISTS files_by_marked_for_deletion_deleting_idx
    ON files (marked_for_deletion_at) WHERE state = 'deleting';
DROP INDEX CONCURRENTLY IF EXISTS files_by_state_only_deleting_idx;
```

If the concurrent build fails it leaves an INVALID index behind; drop it and retry before upgrading, because `CREATE INDEX IF NOT EXISTS` matches on name only and would otherwise keep an unusable index. The migration itself stays non-concurrent on purpose: Flyway rolls it back as a unit, so a failure retries cleanly on the next broker start, whereas a failed concurrent build inside the migration would leave both an invalid index and a failed history row that blocks every broker start until someone runs `flyway repair`. It sets `lock_timeout = 5s` so a pending `ACCESS EXCLUSIVE` request cannot queue behind a long transaction and stall every commit arriving behind it.

## Behaviour change worth knowing

The worklist is now the oldest eligible files rather than an arbitrary prefix, because the plan walks the new index in `marked_for_deletion_at` order. Keys the object store never confirms stay marked and stay oldest, so in principle `limit` or more permanently-failing keys could crowd out later files. In practice there is no such state to reach: deleting a missing object succeeds, throttling is transient and hits a different subset each cycle, and the only persistent failures (bucket policy, Object Lock) are bucket-wide - at which point nothing is deletable and rotating the worklist would not slow the backlog either way. Documented on `ControlPlane#getFilesToDelete`, with `FileCleanerFilesFailedRate` as the signal.

## Testing

- `FileCleanerMockedTest`: probe limit passed to the control plane, saturation at cap+1, no saturation at exactly the cap, and the delete set trimmed back to the cap in both cases.
- `FindFilesToDeleteJobTest`: eligible file returned, grace-period boundary exclusive, `limit` honoured and `0` unbounded - run against the migrated schema.
- `InMemoryControlPlane` now returns oldest-first to match Postgres, so tests that exercise truncation see the same prefix in both planes.
- Changelog generator run over past release ranges and over `inkless-release-0.46..HEAD`; V24 is reported as `ALTER TABLE`, function-only migrations produce no note, and a header quoting the `CONCURRENTLY` form is not mistaken for one.


storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
```diff
@@ -106,7 +110,14 @@ public void run() {
                 LOGGER.info("No files to delete, sleeping for {}", sleepDuration);
                 time.sleep(sleepMillis);
             } else {
-                LOGGER.info("Running file cleaner: deleting {} of {} marked files", objectKeyPaths.size(), filesToDelete.size());
+                final boolean saturated = maxFilesPerCycle > 0 && filesToDelete.size() >= maxFilesPerCycle;
+                if (saturated) {
+                    metrics.recordFileCleanerCycleSaturated();
+                    LOGGER.info("Running file cleaner: deleting {} files (per-cycle cap reached, more remain)",
+                        objectKeyPaths.size());
```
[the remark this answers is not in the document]
jeqo: Good catch. Solving with a sentinel added to the initial query, so we can diff between these cases.

storage/inkless/src/main/java/io/aiven/inkless/control_plane/ControlPlane.java
```diff
@@ -109,7 +110,20 @@ List<FindBatchResponse> findBatches(
...
-    List<FileToDelete> getFilesToDelete();
+    /**
+     * Files that are dereferenced and whose deletion grace period has passed, i.e. those the caller may
+     * physically delete from object storage.
+     * <p>
+     * Both parameters are bounds the implementation MUST enforce: filtering on the caller side does not
+     * help, because the cost of a fully-drained result is already paid by then.
+     * <p>
+     * No ordering is guaranteed. Since {@code markedBefore} is applied here, every returned entry is
+     * actionable, so an arbitrary truncation still makes progress.
```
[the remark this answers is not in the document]
jeqo: Ordering refactored.

storage/inkless/src/main/resources/db/migration/V25__File_cleaner_marked_for_deletion_index.sql
```diff
@@ -0,0 +1,51 @@
...
+-- than the build. On timeout the migration rolls back and the next broker start retries it. The build
+-- itself is not capped by this -- lock_timeout covers acquisition only.
+SET LOCAL lock_timeout = '5s';
+
+CREATE INDEX IF NOT EXISTS files_by_marked_for_deletion_deleting_idx
+    ON files (marked_for_deletion_at) WHERE state = 'deleting';
+
+-- Superseded: the new index serves `state = 'deleting'` equally well, and the only other query with
+-- that predicate (delete_files_v1) is keyed by object_key.
+DROP INDEX IF EXISTS files_by_state_only_deleting_idx;
```
viktorsomogyi: One possible problem is that all these statements run in a single transaction as I understand it, and when the drop fails, it can invalidate the newly built index and may cause a broker failure. At the next restart, the same problem could be possible. The drop command can fail because it needs exclusive lock, so if autovacuum for instance takes a bit too long, then it'll easily time out. Also the index build on a GenD cluster may not be so fast, which could add to the startup time, although it might be a background task though.
We could refactor it to a separate migration file (separate transaction) or do it after the broker is started (and then we'd avoid unnecessary broker restarts because of an index drop failure).

storage/inkless/src/main/resources/db/migration/V25__File_cleaner_marked_for_deletion_index.sql
```diff
@@ -0,0 +1,51 @@
...
+
+-- Index the file cleaner's grace-period predicate so its LIMIT bounds the scan, not just the result.
+--
+-- files_by_state_only_deleting_idx indexed `state` alone, so `marked_for_deletion_at < ?` was a
+-- residual filter: a backlog of deleting rows still inside the grace period was walked in full to
+-- return zero rows. Ordering the partial index by marked_for_deletion_at lets the scan stop as soon
+-- as the limit is met, and turns the all-within-grace case into a range scan that ends at the first
+-- non-matching entry.
+--
+-- Consequence to be aware of: the returned prefix is now the oldest eligible files rather than
```
viktorsomogyi: I think this "oldest eligible files" comment a little bit contradicts what we have in `FindFilesToDeleteJob` on line 64:
```No ORDER BY: the returned rows are all actionable, so any prefix makes progress.```

File: storage/inkless/src/main/resources/db/migration/V25__File_cleaner_marked_for_deletion_index.sql (hunk 3 above)
Comment (jeqo) [reply storage/inkless/src/main/resources/db/migration/V25__File_cleaner_marked_for_deletion_index.sql]: Good catch. I added the lock timeout for bounding this to a short interruption instead of running until the index is completed (of fail to do so), but having both creation and deletion on the same file, could be problematic. 

I think we need to apply the index creation concurrently as an separate operation on existing services so we can avoid issues.

I've separated the files in 2 flyway migrations so they can fail/rollback independently.

File: storage/inkless/src/main/resources/db/migration/V25__File_cleaner_marked_for_deletion_index.sql (hunk 4 above)
Comment (jeqo) [reply storage/inkless/src/main/resources/db/migration/V25__File_cleaner_marked_for_deletion_index.sql]: Good catch -- left over from previous iteration. Fixing it

File: storage/inkless/src/main/java/io/aiven/inkless/control_plane/InMemoryControlPlane.java [hunk 5]
```diff
@@ -518,11 +519,16 @@ public synchronized OptionalLong getCrossTierLogStart(final TopicIdPartition top
...
+    public List<FileToDelete> getFilesToDelete(final Instant markedBefore, final int limit) {
+        final Stream<FileToDelete> candidates = files.values().stream()
             .filter(f -> f.fileState == FileState.DELETING)
+            .filter(f -> f.markedForDeletionAt.isBefore(markedBefore))
             .map(f -> new FileToDelete(f.objectKey, f.markedForDeletionAt))
-            .toList();
+            // Oldest first, mirroring the Postgres plane, where the order falls out of
+            // files_by_marked_for_deletion_deleting_idx. Keeps the truncated prefix the same in both,
+            // including the head-of-line case where undeletable keys stay oldest.
+            .sorted(Comparator.comparing(FileToDelete::markedForDeletionAt));
```
viktorsomogyi: Can we avoid ordering here to stay faithful to the "no ordering guaranteed" contract of the interface?

storage/inkless/src/test/java/io/aiven/inkless/control_plane/postgres/FindFilesToDeleteJobTest.java
```diff
@@ -89,10 +89,50 @@ void tearDown() {
...
+        final Instant markedBefore = MARKED_FOR_DELETION_AT.plusMillis(1);
+
+        final FindFilesToDeleteJob bounded = new FindFilesToDeleteJob(
+            time, pgContainer.getJooqCtx(), markedBefore, 2, duration -> {});
+        assertThat(bounded.call()).hasSize(2);
+
+        final FindFilesToDeleteJob unbounded = new FindFilesToDeleteJob(
+            time, pgContainer.getJooqCtx(), markedBefore, 0, duration -> {});
+        assertThat(unbounded.call()).hasSize(3);
+    }
```
viktorsomogyi, line 108: Can we have a test that mixes eligible and non-eligible files?

viktorsomogyi (commented):
had 2 minor comments, otherwise LGTM

storage/inkless/src/main/java/io/aiven/inkless/control_plane/InMemoryControlPlane.java
jeqo: Yes, good call. Adding this change

storage/inkless/src/test/java/io/aiven/inkless/control_plane/postgres/FindFilesToDeleteJobTest.java
jeqo: Good catch. I've also refactored the test a bit as the carried state from the `beforeEach` was making it hard to read with indirection. Should be better now.

viktorsomogyi (approved):

PR #746, opened 2026-08-11
feat(inkless:retention): add LastSuccessfulFileCleanupAgeMs [KC-412]

jeqo:
#735 added age-of-last-success gauges so a stuck operation surfaces before a customer reports it, but FileCleaner did not get one. Its counters are all cumulative, so a cleaner that has stopped cycling looks exactly like one with nothing to delete.

Follow the same pattern as LastSuccessfulFileCommitAgeMs: record the clock at the end of every cycle that completes without error and expose the age, returning -1 until the first completion so "never ran" is distinguishable from "just ran".

Cycles with no work count as successful. The gauge answers whether the cleaner is running, not whether it is deleting; delete volume is already covered by FileCleanerFilesRate.

This overlaps with GetFilesToDeleteLastSuccessfulQueryAgeMs, which #735 generated for every control-plane query and which also goes stale when the cleaner stops cycling. The difference is what happens when a cycle starts but never finishes - an S3 delete stalling under throttling keeps the query gauge fresh while no cleanup completes. That case is the reason to scope this to the cycle rather than rely on the query gauge.

### DIFF
--- a/docs/inkless/metrics.rst
+++ b/docs/inkless/metrics.rst
@@ -236,15 +236,16 @@ FileCleaner metrics
 io.aiven.inkless.delete:type=FileCleaner
 ----------------------------------------
 
-===========================  =================================================================================================================================
-Attribute name               Description                                                                                                                      
-===========================  =================================================================================================================================
-FileCleanerErrorRate         Total number of file cleaning errors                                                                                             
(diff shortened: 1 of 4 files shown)


jclarysse (dismissed):
LGTM 👍

jclarysse (approved):
LGTM 👍

PR #750, opened 2026-08-12
fix(inkless:control_plane): make the diskless fetch budget fair across partitions [KC-407]

jeqo:
`find_batches` guarantees every partition in a fetch at least one batch. Upstream grants that allowance **once per request** and returns **zero bytes** past the shared `fetch.max.bytes`, and that zero-byte response is the only input to Kafka's fetch fairness — both readers of this path demote a partition only when it returned bytes. The per-partition floor therefore makes every partition look served, the rotation never advances, and lagging partitions sit at one batch per round while the head drains: catch-up serializes instead of running concurrently.

Restoring the upstream semantics fixes the consumer path and the consolidation fetcher together — same function, same rotation rule — with no consolidation-specific change.

## Commits

- **preserve fetch-session order into the diskless control plane** — `fetchInfos.toMap` discarded the rotated order; above 4 entries `immutable.Map` is a `HashMap` whose iteration order is a pure function of the key set, so it was identical on every fetch.
- **make find_batches minOneMessage request-level** — one predicate in `find_batches_v2` (shipped as migration `V27`).
- **pin the empty-response contract the fetch rotation needs** — test-only guard on the consolidation path.

**Not reorderable:** commit 2 without commit 1 is worse than the bug (request order stays a fixed `HashMap` order, so the tail gets zero instead of one, permanently). See the commit messages for the reasoning behind each step.

## The SQL change

The change to `find_batches_v2` ships as a `CREATE OR REPLACE FUNCTION`, so the raw diff shows the whole function as new. The function body differs from `V19` by **one predicate** (the surrounding added lines are explanatory comments):

```diff
-            IF l_partition_batch_count = 0
+            IF l_global_bytes = 0
                 OR (l_partition_bytes < l_request.max_partition_fetch_bytes
                     AND l_global_bytes < fetch_max_bytes)
```

```
diff <(sed -n '/^CREATE OR REPLACE/,$p' storage/inkless/src/main/resources/db/migration/V19__Find_batches_bounded_scan.sql) \
     <(sed -n '/^CREATE OR REPLACE/,$p' storage/inkless/src/main/resources/db/migration/V27__Find_batches_request_level_min_one_message.sql)
```

The bounded O(result) scan and its index-scan plan dependency, error handling, request ordering and `max_batches_per_partition` are untouched.

`InMemoryControlPlane` (selectable for local and system-test deployments) previously applied `fetch.max.bytes` per partition and ignored both the per-partition byte cap and the batch cap; it now enforces all three so the same expectation can be asserted against both control planes. Tests that relied on the in-memory plane ignoring the per-partition cap see fewer batches.

## Behaviour change

A partition with data available can now return an **empty** response when the request budget is spent. `FetchCompleter` maps an empty batch list to `MemoryRecords.EMPTY` with `Errors.NONE` and the real high watermark — the response a caught-up partition already gets — so lag stays visible.

That empty path had a latent gap this change surfaces: it dropped `lastStableOffset` (passing `OptionalLong.empty()` where the served branch carries the high watermark), which reaches a `READ_COMMITTED` consumer as `INVALID_LAST_STABLE_OFFSET` and skips `SubscriptionState.updateLastStableOffset`, staling `currentLag`/`endOffsets`. It was harmless while only a caught-up partition reached the branch; under request-level budgeting a lagging partition reaches it on every round it waits its turn. The empty success path now carries the high watermark as the LSO, and the error path keeps it unset.

Response size is clamped again as a consequence: overshoot drops from one batch **per partition** to one **per request**.

Batches served per partition, before → after: consumer profile (4 partitions, 4 rounds) `[32, 4, 4, 4]` → `[8, 8, 8, 8]`; consolidation budget ratio (10 partitions, 10 rounds) `[50 x5, 10 x5]` → `25` each, at **300 → 250** total batches fetched — so this is cheaper, not a fairness/throughput tradeoff.

## Testing

Cross-request budget fairness is pinned end to end: `FindBatchesFairnessTest` replays each caller's move-to-end rotation against the real SQL and asserts every partition advances equally over many rounds (consumer and consolidation profiles); `FindBatchesJobTest` and `AbstractControlPlaneTest` assert that partitions past the budget return no batches, the latter against both control planes. `FetchCompleterTest` covers the empty-but-lagging response carrying the LSO; `ConsolidationFetcherThreadTest` and `ReplicaManagerInklessTest` guard the empty-response and order-preservation plumbing.

Five parameterised rows across three existing tests asserted the floor; they are **relocated** into `findServesOnlyTheFirstPartitionWhenTheRequestBudgetIsSpent` with the second partition now expected empty. No existing expectation was adjusted in place.

## Follow-up (not in this PR)

`fetchParamsWithNewMaxBytes` is an identity function — `Math.max((maxBytes * percentage).toInt, maxBytes)` with every caller passing `percentage <= 1` — so both the classic and diskless legs receive the full `fetch.max.bytes`, contradicting its docstring. Independent defect, deliberately deferred: the obvious fix splits by partition count, which can cut diskless catch-up N-fold on mixed requests and still would not bound consolidating fetches. Filed as a gap with measurements needed first.

The order invariant that `fetchDisklessMessages` depends on is carried by javadoc and tests, not by a type: `Seq[(TopicIdPartition, PartitionData)]` cannot express that its order is load-bearing. A per-caller order-propagation test would guard the producers directly; a type-level change would not (a `Seq`/`LinkedHashMap` is ordered but cannot encode that the order is the rotation order).


storage/inkless/src/main/resources/db/migration/V27__Find_batches_request_level_min_one_message.sql
```diff
@@ -0,0 +1,163 @@
...
+-- budget-crossing batch is included, responses stay in request order, and max_batches_per_partition still
+-- caps each partition. Partitions past the budget now return an empty batch list, which FetchCompleter
+-- already maps to MemoryRecords.EMPTY with Errors.NONE and the real high watermark -- the same response a
+-- caught-up partition gets today, so this is an existing path reached for a second reason.
+--
+-- Do not reintroduce a per-partition floor, and do not "fix" this by dividing the budget between
+-- partitions (fair-share/water-filling): that keeps every partition in every response, so it restores the
+-- per-partition scan and GET cost, and it gives a lagging partition 1/n of the budget per round instead of
+-- a full budget on its turn.
+CREATE OR REPLACE FUNCTION find_batches_v2(
```
AnatolyPopov, line 35: Should it be v3 instead since we have find_batches_v2 already?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2284,7 +2288,14 @@ class ReplicaManager(val config: KafkaConfig,
   def fetchDisklessMessages(params: FetchParams,
                             fetchInfos: Seq[(TopicIdPartition, PartitionData)]): CompletableFuture[Seq[(TopicIdPartition, FetchPartitionData)]] = {
     inklessFetchHandler match {
-      case Some(handler) => handler.handle(params, fetchInfos.toMap.asJava).thenApply(_.asScala.toSeq)
+      case Some(handler) =>
+        // find_batches spends the shared fetch.max.bytes in request order, and FetchSession's move-to-end
+        // rotation is what makes that order fair across rounds -- so the rotated order must survive here.
+        // An immutable.Map would be a HashMap above 4 entries: its iteration order ignores insertion and
+        // is stable for a given key set, which pins the same partitions last on every fetch.
```
AnatolyPopov: But I believe that we need then to preserve order in DelayedFetch that relies on Map order and the rotation does not really help there. For this to work we need to change lines 294-297 in DelayedFetch to be
```
 val disklessFetchInfos = disklessFetchPartitionStatus.asScala.iterator.map { case (tp, status) =>
          tp -> status.fetchInfo
}.toSeq
replicaManager.fetchDisklessMessages(disklessParams, disklessFetchInfos)
```

storage/inkless/src/main/resources/db/migration/V25__Find_batches_request_level_min_one_message.sql
```diff
@@ -0,0 +1,163 @@
...
+-- ReplicaManager.readFromLocalLog, minOneMessage starts true and is cleared as soon as any partition
+-- returns bytes, while limitBytes clamps every partition after it -- so a partition reached once the
+-- budget is spent returns zero bytes.
+--
+-- That zero-byte response is the only input to Kafka's fetch fairness. Both readers of this function move
+-- a partition to the back of their order only when it returned bytes
+-- (FetchSession.IncrementalFetchContext.PartitionIterator: recordsSize > 0;
+-- AbstractFetcherThread.processFetchRequest: validBytes > 0), so a partition that returns nothing keeps
+-- its place and is served first, with a full budget, on the next round. Fairness is therefore a rotation
+-- *across* responses, not a division *within* one.
```
AnatolyPopov:
```suggestion
-- AbstractFetcherThread.processFetchRequest: validBytes > 0 -- except on a partition's first fetch,
-- before lag is known, where partitionStates.updateAndMoveToEnd fires unconditionally), so a partition
-- that returns nothing keeps its place and is served first, with a full budget, on the next round (once
-- lag tracking has started). Fairness is therefore a rotation *across* responses, not a division *within*
-- one.

```

core/src/main/scala/kafka/server/ReplicaManager.scala
AnatolyPopov: Same thing applies to line 173 in DelayedFetch I think

AnatolyPopov (dismissed):

storage/inkless/src/main/resources/db/migration/V27__Find_batches_request_level_min_one_message.sql
jeqo: No need as the function signature hasn't change.

On the rolling upgrade consequences, it may have an impact, but is focused and temporary: consumers on lagging partitions will now receive zero where before received one batch _when response budged is empty_. As the impact is on an edge case, I think keeping the same versions is safe. Otherwise, if affecting all consumers, a new version would be warrant.

storage/inkless/src/main/resources/db/migration/V25__Find_batches_request_level_min_one_message.sql
jeqo: Applied, scoped to the `AbstractFetcherThread` half — `FetchSession`'s condition has no lag clause. Also fixed the sentence it sits in: it called both classes "readers of this function", and neither references `find_batches`; they supply its request order and rotate on the bytes it returns.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: You are right, this actually fulfills the fix, thanks!

My earlier `LinkedHashMap` change did run here, it was just a no-op. the Seq it received was already shuffled, so it preserved the shuffle. Of the four callers of `fetchDisklessMessages` only `ReplicaManager:2799` was passing an ordered Seq, and that's the tiered-storage branch. Chasing that down found a second producer with the same defect: `buildConsolidationSupplementFetchInfos` iterated the `supplements` HashMap and looked up in `fetchInfos`. Inverting it to walk the ordered `fetchInfos` fixes the order and drops the intermediate `toMap`. Both fixed, both with tests verified red against the old shapes first.

On 173 -- that one is fine as-is: the lambda returns a `FindBatchRequest`, not a tuple, so it's an `ArrayBuffer` and order survives. It's also unbudgeted (`findDisklessBatches` passes `Int.MaxValue`), so order can't affect the outcome there. Added a comment saying so, since the tuple/non-tuple distinction is exactly what caused the bug above.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2295,10 +2295,23 @@ class ReplicaManager(val config: KafkaConfig,
...
+   * contract. This is the last hop that can preserve the order, so every producer of the Seq handed in
+   * here has to preserve it too.
+   */
   def fetchDisklessMessages(params: FetchParams,
                             fetchInfos: Seq[(TopicIdPartition, PartitionData)]): CompletableFuture[Seq[(TopicIdPartition, FetchPartitionData)]] = {
     inklessFetchHandler match {
-      case Some(handler) => handler.handle(params, fetchInfos.toMap.asJava).thenApply(_.asScala.toSeq)
+      case Some(handler) =>
+        val ordered = mutable.LinkedHashMap.from(fetchInfos).asJava
+        fetchInfos.foreach { case (tp, data) => ordered.put(tp, data) }
```
AnatolyPopov: This I guess is not needed anymore, since the new LinkedHashMap is created with the elements of fetchInfos already

AnatolyPopov:
On top of one leftover I posted above, this comment https://github.com/aiven/inkless/blob/main/core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala#L193C1-L198C87 seems to be now outdated. Could you check it @jeqo?
Otherwise LGTM.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: good catch. was too lazy to verify before pushing 😅

AnatolyPopov (approved):
LGTM, thanks for addressing everything!

PR #752, opened 2026-08-14
fix(inkless): repair missing control-plane rows on CreatePartitions retry [KC-420]

viktorsomogyi:
A failed control-plane write after KRaft already applied the increase left partitions advertised but unusable; retry returned INVALID_PARTITIONS and skipped the insert. Also treat a diskless LATEST below the committed seal as the seal, so a placeholder row cannot truncate the classic prefix.

### DIFF
--- a/core/src/main/scala/kafka/server/ControllerApis.scala
+++ b/core/src/main/scala/kafka/server/ControllerApis.scala
@@ -987,14 +987,12 @@ class ControllerApis(
 
       case Some(cp) =>
         val eligibleRequests = (topics.asScala zip results.asScala)
-          // NONE is the first increase. INVALID_PARTITIONS (and TOPIC_ALREADY_EXISTS, copied from
-          // create-topic) is a retry after KRaft already applied, except a decrease, which uses the
-          // same error code and must not write. disklessPartitionCreateRequests drops those.
+          // NONE is the first increase. INVALID_PARTITIONS is a retry after KRaft already applied,
+          // except a decrease, which uses the same error code and must not write.
+          // disklessPartitionCreateRequests drops those.
           .filter { case (_, res) =>
             val code = res.errorCode()
-            code == Errors.NONE.code() ||
-              code == Errors.TOPIC_ALREADY_EXISTS.code() ||
-              code == Errors.INVALID_PARTITIONS.code()
+            code == Errors.NONE.code() || code == Errors.INVALID_PARTITIONS.code()
           }
           // In contrast to the topic creation, we only create new partitions to existing topics.
           // Hence, the topics themselves must be in the metadata already, no need to wait.
--- a/core/src/main/scala/kafka/server/ControllerApis.scala
+++ b/core/src/main/scala/kafka/server/ControllerApis.scala
@@ -53,7 +53,7 @@ import org.apache.kafka.common.Uuid
 import org.apache.kafka.controller.ControllerRequestContext.requestTimeoutMsToDeadlineNs
 import org.apache.kafka.controller.{Controller, ControllerRequestContext}
 import org.apache.kafka.image.publisher.ControllerRegistrationsPublisher
-import org.apache.kafka.metadata.{BrokerHeartbeatReply, BrokerRegistrationReply}
+import org.apache.kafka.metadata.{BrokerHeartbeatReply, BrokerRegistrationReply, PartitionRegistration}
 import org.apache.kafka.common.security.auth.KafkaPrincipal
 import org.apache.kafka.common.security.auth.SecurityProtocol
 import org.apache.kafka.raft.RaftManager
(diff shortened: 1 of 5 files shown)


storage/inkless/src/main/java/io/aiven/inkless/control_plane/CreateTopicAndPartitionsRequest.java
```diff
@@ -20,14 +20,14 @@
 import org.apache.kafka.common.Uuid;
 
 /**
- * Request to create control-plane {@code logs} rows for a topic's partitions.
+ * Request to create control-plane {@code logs} rows for {@code [firstPartition, numPartitions)}.
+ * {@code numPartitions} is the exclusive end of that range, not necessarily the topic's partition
+ * count: a retry that fills holes may emit several such ranges.
```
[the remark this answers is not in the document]
viktorsomogyi: won't do

core/src/main/scala/kafka/server/ControllerApis.scala
```diff
@@ -987,40 +987,102 @@ class ControllerApis(
 
       case Some(cp) =>
         val successfulCreations = (topics.asScala zip results.asScala)
-          // It's OK if we retry creating for already existing topics,
-          // this may save some trouble when Inkless creation failed for some reason and the user retries.
-          .filter { case (_, res) => res.errorCode() == Errors.NONE.code() || res.errorCode() == Errors.TOPIC_ALREADY_EXISTS.code() }
+          // NONE is the first increase. INVALID_PARTITIONS (and TOPIC_ALREADY_EXISTS, copied from
+          // create-topic) is a retry after KRaft already applied, except a decrease, which uses the
+          // same error code and must not write. disklessPartitionCreateRequests drops those.
+          .filter { case (_, res) =>
```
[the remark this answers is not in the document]
viktorsomogyi: fair

core/src/test/scala/unit/kafka/server/ControllerApisTest.scala
```diff
@@ -1112,6 +1125,196 @@ class ControllerApisTest {
...
+    controllerApis = createControllerApis(None, controller, inklessControlPlane = Some(controlPlane))
+
+    val request = new CreatePartitionsRequestData().setValidateOnly(false)
+    request.topics().add(topic)
+    controllerApis.createPartitions(ANONYMOUS_CONTEXT, request, _ => Set(topicName)).get()
+
+    verify(controlPlane).createTopicAndPartitions(ArgumentMatchers.eq(singleton(
+      new CreateTopicAndPartitionsRequest(topicId, topicName, 3, 4))))
+  }
+
```
jeqo, line 1317: for completeness, we can add the scenario that should still fail and do not insert rows:
```suggestion

  @Test
  def testCreateDisklessPartitionsDecreaseRetryDoesNotInsert(): Unit = {
    // topic image already has 4 partitions; a decrease to 2 is rejected by KRaft with
    // INVALID_PARTITIONS and must NOT write control-plane rows.
    val topic = new CreatePartitionsTopic().setName(topicName).setAssignments(null).setCount(2)
    val result = new CreatePartitionsTopicResult().setName(topicName).setErrorCode(INVALID_PARTITIONS.code())
    // ... arrange metadataCache image with 4 partitions for topicId (as the other retry tests do) ...
    // invoke createDisklessPartitions(...)
    verify(controlPlane, never()).createTopicAndPartitions(any())
  }

```

core/src/main/scala/kafka/server/ControllerApis.scala
```diff
@@ -986,41 +986,103 @@ class ControllerApis(
...
+    if (retry && count < imagePartitionCount(topicId)) {
+      Seq.empty
+    } else {
+      val firstPartition =
+        if (retry) 0
+        else priorTopicStates.get(topicName) match {
+          case Some(state) if state.topicId == topicId => math.min(state.numPartitions, count)
+          case _ => 0
+        }
+      val partitions = bornDisklessPartitions(topicId, count).filter(_ >= firstPartition)
```
jeqo: `imagePartitionCount` and `bornDisklessPartitions` both access `metadataCache.currentImage()` snapshot at different points, potentially getting different images; maybe:  

```suggestion
    val retry = errorCode != Errors.NONE.code()
    val imagePartitions = Option(metadataCache.currentImage().topics().getTopic(topicId))
        .map(_.partitions())
        .getOrElse(util.Collections.emptyMap())
    if (retry && count < imagePartitions.size()) {
      Seq.empty
    } else {
      val firstPartition =
        if (retry) 0
        else priorTopicStates.get(topicName) match {
          case Some(state) if state.topicId == topicId => math.min(state.numPartitions, count)
          case _ => 0
        }
      val partitions = bornDisklessPartitions(imagePartitions, count)....
```

core/src/test/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPointTest.scala
```diff
@@ -279,6 +279,24 @@ class DisklessLeaderEndPointTest {
...
+    val endPoint = listOffsetEndPointWithPlaceholderEpoch(offset = 200000L, seal = 150000L, disklessLeaderEpoch = 5)
+    assertEquals(new OffsetAndEpoch(200000L, 5), endPoint.fetchLatestOffset(topicPartition, 3))
+  }
+
+  @Test
+  def testFetchLatestOffsetNonEmptyBelowSealUsesSealAndDisklessEpoch(): Unit = {
+    val endPoint = listOffsetEndPointWithPlaceholderEpoch(offset = 10L, seal = 150000L, disklessLeaderEpoch = 5)
+    assertEquals(new OffsetAndEpoch(150000L, 5), endPoint.fetchLatestOffset(topicPartition, 3))
+  }
+
```
jeqo, line 299: for completeness, adding the negative 
```suggestion

@Test
def testFetchLatestOffsetSwitchPendingSealIsUnchanged(): Unit = {
  val endPoint = listOffsetEndPointWithPlaceholderEpoch(
    offset = 0L,
    seal = PartitionRegistration.CLASSIC_TO_DISKLESS_SWITCH_PENDING,
    disklessLeaderEpoch = PartitionRegistration.NO_DISKLESS_LEADER_EPOCH)
  assertEquals(new OffsetAndEpoch(0L, 0), endPoint.fetchLatestOffset(topicPartition, 3))
}

@Test
def testFetchLatestOffsetBornDisklessSealIsUnchanged(): Unit = {
  val endPoint = listOffsetEndPointWithPlaceholderEpoch(
    offset = 0L,
    seal = PartitionRegistration.NO_CLASSIC_TO_DISKLESS_START_OFFSET,
    disklessLeaderEpoch = PartitionRegistration.NO_DISKLESS_LEADER_EPOCH)
  assertEquals(new OffsetAndEpoch(0L, 0), endPoint.fetchLatestOffset(topicPartition, 3))
}

@Test
def testFetchLatestOffsetExactlyAtSealIsUnchanged(): Unit = {
  val endPoint = listOffsetEndPointWithPlaceholderEpoch(offset = 150000L, seal = 150000L, disklessLeaderEpoch = 5)
  assertEquals(new OffsetAndEpoch(150000L, 5), endPoint.fetchLatestOffset(topicPartition, 3))
}

```

core/src/main/scala/kafka/server/ControllerApis.scala
```diff
@@ -986,41 +986,103 @@ class ControllerApis(
...
-          // this may save some trouble when Inkless creation failed for some reason and the user retries.
-          .filter { case (_, res) => res.errorCode() == Errors.NONE.code() || res.errorCode() == Errors.TOPIC_ALREADY_EXISTS.code() }
+        val eligibleRequests = (topics.asScala zip results.asScala)
+          // NONE is the first increase. INVALID_PARTITIONS (and TOPIC_ALREADY_EXISTS, copied from
+          // create-topic) is a retry after KRaft already applied, except a decrease, which uses the
+          // same error code and must not write. disklessPartitionCreateRequests drops those.
+          .filter { case (_, res) =>
+            val code = res.errorCode()
+            code == Errors.NONE.code() ||
+              code == Errors.TOPIC_ALREADY_EXISTS.code() ||
```
jeqo: does this error actually need to be covered? I think it comes from topic creation which is not reachable when creating partitions

jeqo (commented):
Looking good. Left a few comments before approving.

core/src/test/scala/unit/kafka/server/ControllerApisTest.scala
viktorsomogyi: isn't it covered in `testCreateDisklessPartitionsDoesNotInsertOnPartitionDecrease` or do you propose something different?

core/src/main/scala/kafka/server/ControllerApis.scala
viktorsomogyi: fair comment, as I see this error only used on the topic creation path, so we don't need to cover this here

core/src/test/scala/unit/kafka/server/ControllerApisTest.scala
jeqo: You're right, I missed this one and is already covered. Sorry for the noise.

jeqo (approved):

PR #754, opened 2026-08-17
fix(inkless:controller): skip switched partitions in unfence ISR expansion

jeqo:
`expandIsrForDisklessManagedPartitions` re-admitted a returning broker to the ISR of every `diskless.enable=true` partition, selecting on the topic config alone. That is correct for born-diskless partitions, the case #643 targeted: all their data is in object storage, so any live replica is current.

A partition switched from classic breaks that premise. Its records below `classicToDisklessStartOffset` exist only in the replicas' local logs, so a returning replica could hold an incomplete prefix and was still placed in ISR and made electable.

Skip any partition whose seal is not `NO_CLASSIC_TO_DISKLESS_START_OFFSET`, which covers both a committed seal and a switch still PENDING. Those partitions earn ISR through AlterPartition once the leader observes follower fetch state at the seal -- the path #697 added. Removing this shortcut before that path existed would have left switched partitions with no ISR-recovery route at all.

The test drives one unfence across three topics -- born-diskless, switched with a committed seal, and switch-pending. Asserting only that switched partitions stay out would also pass against a guard that skipped everything, so the born-diskless leg is what makes it discriminating. Verified red-then-green: with the guard reverted, the switched assertion fails with expected <false> but was <true>.

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
+++ b/metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java
@@ -2009,6 +2009,14 @@ public class ReplicationControlManager {
             for (var entry : topic.parts.entrySet()) {
                 int partitionId = entry.getKey();
                 PartitionRegistration partition = entry.getValue();
+                // Only born-diskless partitions qualify: a switched (seal committed) or mid-switch
+                // (PENDING) partition still has classic records that exist solely in the replicas'
+                // local logs, so an unfenced replica is not necessarily complete. It earns ISR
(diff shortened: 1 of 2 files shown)


viktorsomogyi (approved):
LGTM, only one comment that Cursor flagged, which is that growing or moving a switched partition (via reassignment) still puts every target broker in ISR immediately, including brokers with no classic prefix. Not introduced here, it's a leftover from earlier commits and I think it's fine to do it as a separate follow-up PR.

jeqo:
@viktorsomogyi good catch, working on it already: https://github.com/aiven/inkless/pull/757

PR #758, opened 2026-08-18
docs(inkless:ai): add writing style rules to AGENTS.md

viktorsomogyi:
Agents write PR text, commit messages, and comments without a shared style. `AGENTS.md` now carries a short writing-style section adapted from the [Google developer documentation style guide](https://developers.google.com/style).

The section covers tone, grammar, formatting, a small word list, commit and PR shape, and comments. Project conventions stay first: conventional-commit titles, Kafka terms such as `replica` and `ISR`, and rare comments that explain why. Upstream Kafka files keep upstream style.

The full Google guide remains the lookup for terms this section doesn't cover.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,8 +61,140 @@ Upstream Kafka build, test, and tooling commands are in the root
 
 **MUST**-follow rules and guidelines for development.
 
-#### Comments
+#### Writing style
 
-Write self-documenting code. Comments should be rare and explain **why** (a
-non-obvious constraint, workaround, or deliberate deviation), never **what** the
-code does. Match the surrounding comment density and never restate the code.
+Applies to Inkless-owned docs, PR titles and bodies, commit messages, and
+comments you write. Upstream Kafka files keep upstream style. Do not rewrite
+existing comments or docs only to match this section.
+
+Follow project conventions first, then this section. It is adapted from the
+[Google developer documentation style guide](https://developers.google.com/style).
+If something is not covered here, look it up there. Prefer a clear sentence
+over a strictly compliant one. When you depart from the guide, stay consistent
+in that file or PR.
+
+##### Tone
+
+Write like a knowledgeable teammate: conversational and direct, not stiff and
+not cute. The job is to give the reader the fact they came for.
+
+- MUST use active voice. Name the actor. "The broker retries the job." not
+  "The job is retried."
+- MUST use present tense for current behavior. "The server sends an
+  acknowledgment." not "The server will send an acknowledgment." Use `will`
+  only when the action is actually later (for example, a deferred archive).
+- MUST NOT use `please` in instructions or PR bodies. MUST NOT call a change
+  `simple`, `easy`, or `just`. Those words hide the real cost.
+- SHOULD use common two-word contractions (`don't`, `isn't`, `can't`,
+  `you're`). Negation contractions are easier to see while scanning than a
+  lone `not`.
+- SHOULD address the reader as `you` in docs and instructions. Use the
+  imperative when telling them to do something ("Run the test."). In code
+  comments, describe what the system does in third person ("The broker
+  retries the job.").
+- SHOULD NOT use first-person plural for the code (`we then update the
+  offset`). `we` is fine only when it means the Inkless project.
+- MUST NOT use slang, memes, exclamation points, or `tl;dr`.
+- SHOULD put the condition or goal before the instruction, so the reader can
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -61,8 +61,138 @@ Upstream Kafka build, test, and tooling commands are in the root
...
+| ----- | ------ |
+| `allows you to` | `lets you` |
+| `utilize`, `leverage` | `use` |
+| `currently`, `at this time`, `as of this writing` | omit; state the fact |
+| `will` for current behavior | present tense |
+| `please note` | the note, with no preamble |
+| `whitelist` / `blacklist` | `allowlist` / `denylist`, or the action ("deny requests from") |
+| `dummy` (for a stand-in value) | `placeholder` |
+| `above` / `below` for a position in a doc | `preceding` / `following` |
+| `abort` (except a POSIX signal) | `stop`, `cancel`, or `end` |
```
jeqo: I think we are already using `abort` in here: https://github.com/aiven/inkless/blame/8f5cd6fed422592bd9733dadf56c120a2a4dd7c4/metadata/src/main/java/org/apache/kafka/controller/ReplicationControlManager.java#L1702-L1703 and in transactions -- so this may conflict

```suggestion
| `abort` in general prose about stopping work | `stop`, `cancel`, or `end` |
```

also maybe worth adding after:
```
Keep a term from the "Avoid" column when it names an existing operation in Kafka or
Inkless rather than describing one. Write the operation's own word, in backticks,
and use the preferred word only in prose that is not naming that operation.
```

AGENTS.md
```diff
@@ -61,8 +61,138 @@ Upstream Kafka build, test, and tooling commands are in the root
...
+  with em dashes or parentheses, give it its own sentence, or introduce it
+  with a colon. An em dash is fine for a brief restatement or interruption
+  that is not a new claim. Parentheses are fine for a bare label that has no
+  thought of its own, such as an abbreviation gloss or a citation.
+- SHOULD make pronoun antecedents obvious. Prefer "this value" over "this."
+  Use singular `they`. Use `that` for restrictive clauses and `which` (with a
+  comma) for nonrestrictive ones.
+- MUST use a colon, not a dash, in a list of term/description pairs:
+  `Term: description.` not `Term - description.`
+- MUST NOT use `e.g.` or `i.e.`; write `for example` or `that is`.
```
EelisK, line 119: Our current AGENTS.md uses `e.g.`. We could update it 😄 
https://github.com/aiven/inkless/blob/3756826bca92f4084db2616b7c975d3981c2b86c/AGENTS.md?plain=1#L23

jeqo (commented):
I'm testing it now and found a small conflict. Otherwise looks good, thanks!

AGENTS.md
```diff
@@ -61,8 +61,138 @@ Upstream Kafka build, test, and tooling commands are in the root
...
+- SHOULD NOT end a list with `etc.`; introduce the list as incomplete
+  (`such as`) or name the items.
+- MUST use one space between sentences.
+- MUST NOT put a period on a heading.
+
+##### Formatting
+
+- MUST use sentence case for headings and PR section titles: "Test plan" not
```
EelisK, line 129: Could make AGENTS.md consistent with this rule  🔍 
https://github.com/aiven/inkless/blob/8d93db5fc9d13f843f785acbc11ff4ccedb49b59/AGENTS.md?plain=1#L1
https://github.com/aiven/inkless/blob/8d93db5fc9d13f843f785acbc11ff4ccedb49b59/AGENTS.md?plain=1#L25
https://github.com/aiven/inkless/blob/8d93db5fc9d13f843f785acbc11ff4ccedb49b59/AGENTS.md?plain=1#L25
https://github.com/aiven/inkless/blob/8d93db5fc9d13f843f785acbc11ff4ccedb49b59/AGENTS.md?plain=1#L27
https://github.com/aiven/inkless/blob/8d93db5fc9d13f843f785acbc11ff4ccedb49b59/AGENTS.md?plain=1#L53

AGENTS.md
viktorsomogyi: Yea, I agree on both accounts.

AGENTS.md
viktorsomogyi: Let's hope editing AGENTS.md based on the guidelines of that AGENTS.md doesn't create a black hole. 😄 
If it does, sorry for that.

viktorsomogyi: We didn't implode 🎉

jeqo (approved):

EelisK (approved):
LGTM 👍

PR #762, opened 2026-08-19
fix(inkless:fetch): preserve legacy fetch request partition identity in response [KC-353]

tvainika:
Fetch versions 7-12 omit topic IDs, so the request keys a diskless partition with Uuid.ZERO_UUID. ReplicaManager resolves the real topic ID to route the diskless read. Restore the client's original partition identity in respond() before calling back into KafkaApis.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -2422,6 +2422,12 @@ class ReplicaManager(val config: KafkaConfig,
     val disklessFetchInfos = new mutable.ArrayBuffer[(TopicIdPartition, PartitionData)]()
     val classicFetchInfos = new mutable.ArrayBuffer[(TopicIdPartition, PartitionData)]()
     val immediateFetchResponses = new mutable.ArrayBuffer[(TopicIdPartition, FetchPartitionData)]()
+    // Legacy fetch versions (<13) omit the topic ID, so the request keys this partition with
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2613,8 +2620,17 @@ class ReplicaManager(val config: KafkaConfig,
...
-    def respond(response: Seq[(TopicIdPartition, FetchPartitionData)]): Unit =
-      responseCallback(response ++ immediateFetchResponses)
+    def respond(response: Seq[(TopicIdPartition, FetchPartitionData)]): Unit = {
+      // Restore the client's original partition identity (see disklessTopicIdOverrides above)
+      // before handing the response back to the fetch session/request layer.
+      val restored = response.map { case (backfilledTp, data) =>
+        disklessTopicIdOverrides.get(backfilledTp) match {
+          case Some(originalTp) => originalTp -> data
+          case None => backfilledTp -> data
+        }
```
jeqo: we can avoid the checks when no overrides (default case):
```suggestion
        if (disklessTopicIdOverrides.isEmpty) response
        else response.map { case (backfilledTp, data) =>
           disklessTopicIdOverrides.get(backfilledTp) match {
            case Some(originalTp) => originalTp -> data
            case None => backfilledTp -> data
          }
        }
```

jeqo (changes requested):

jeqo (approved):

PR #767, opened 2026-08-21
fix(inkless:metadata): accept pipe as a client.id AZ separator [KC-443]

jeqo:
MirrorMaker 2 appends `|<source>-><target>|<connector>|<role>` to any configured `client.id` via `MirrorConnectorConfig.addClientId()`. With the old regex, that suffix was swallowed into the `diskless_az` capture, so a `client.id` of `diskless_az=euc1-az2` arrived as
`diskless_az=euc1-az2|primary->backup|MirrorSourceConnector|producer` and AZ extraction returned null after failing every hyphen-truncation candidate.

Treat `|` as both a valid leading separator and an AZ stop character so MM2's pipe-appended `client.id` parses the same as the comma-separated form Connect produces. This is symmetric with the existing comma handling and removes the fragility for any client that suffixes client.id with a pipe.

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/metadata/ClientAZExtractor.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/metadata/ClientAZExtractor.java
@@ -23,7 +23,8 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 class ClientAZExtractor {
-    private static final Pattern CLIENT_AZ_PATTERN = Pattern.compile("(^|[ ,])diskless_az=(?<az>[^=, ]*)");
+    // Commas as default separator and pipes added to support MirrorConnectorConfig.addClientId appends to client.id
+    private static final Pattern CLIENT_AZ_PATTERN = Pattern.compile("(^|[ ,|])diskless_az=(?<az>[^=,| ]*)");
 
     // Visible for testing
     static String getClientAZ(final String clientId) {
--- a/docs/inkless/CLIENT-BROKER-AZ-ALIGNMENT.md
+++ b/docs/inkless/CLIENT-BROKER-AZ-ALIGNMENT.md
@@ -148,13 +148,21 @@ The Inkless broker parses the `client.id` field using a regex pattern:
 ```java
 // From ClientAZExtractor.java
 private static final Pattern CLIENT_AZ_PATTERN = 
-    Pattern.compile("(^|[ ,])diskless_az=(?<az>[^=, ]*)");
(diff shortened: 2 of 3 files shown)


storage/inkless/src/main/java/io/aiven/inkless/metadata/ClientAZExtractor.java
```diff
@@ -23,7 +23,10 @@
 import java.util.regex.Pattern;
 
 class ClientAZExtractor {
-    private static final Pattern CLIENT_AZ_PATTERN = Pattern.compile("(^|[ ,])diskless_az=(?<az>[^=, ]*)");
+    // '|' is a separator and a stop character because MirrorMaker 2's
+    // MirrorConnectorConfig.addClientId appends |<source>-><target>|<connector>|<role>
+    // to any configured client.id; without it that suffix is swallowed into the AZ capture.
+    private static final Pattern CLIENT_AZ_PATTERN = Pattern.compile("(^|[ ,|])diskless_az=(?<az>[^=,| ]*)");
```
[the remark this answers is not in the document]
EelisK: This is valid

storage/inkless/src/main/java/io/aiven/inkless/metadata/ClientAZExtractor.java
```diff
@@ -23,7 +23,10 @@
 import java.util.regex.Pattern;
 
 class ClientAZExtractor {
-    private static final Pattern CLIENT_AZ_PATTERN = Pattern.compile("(^|[ ,])diskless_az=(?<az>[^=, ]*)");
+    // '|' is a separator and a stop character because MirrorMaker 2's
+    // MirrorConnectorConfig.addClientId appends |<source>-><target>|<connector>|<role>
+    // to any configured client.id; without it that suffix is swallowed into the AZ capture.
```
EelisK: nit: this is a 3-line comment explaining a single character in a regex. Consider removing it or making it more concise.

```suggestion
    // MM2's MirrorConnectorConfig.addClientId appends |<source>-><target>|<connector>|<role> to client.id
```

EelisK (commented):
Good catch 👍

storage/inkless/src/main/java/io/aiven/inkless/metadata/ClientAZExtractor.java
jeqo: Fair point. Moving the detailed format to doc and kept a simpler doc line here.

EelisK (approved):

PR #770, opened 2026-08-26
fix(inkless-sync): use trunk merge base for --before-version

tvainika:
main-sync.sh --before-version resolved the sync target from a version tag, then took its parent commit. Release tags and branches such as apache/4.3 can carry backports and fixes that never land on apache/trunk. Resolving from the tag pulled those release-only commits into the trunk sync, mixing history that does not belong there.

Resolve the target as the merge base of apache/trunk and apache/$BEFORE_VERSION instead, so the sync only replays commits that are genuinely part of trunk history up to where the release branch diverged.

tag_exists() used git rev-parse on the bare ref, so it matched branches as well as tags, which let a mistyped or branch-only argument pass validation. It now checks refs/tags/<ref> specifically. Added branch_exists(), checking refs/heads/ and refs/remotes/, and use it in main-sync.sh to validate apache/$BEFORE_VERSION as a branch, since --before-version now expects a release branch rather than a tag.

### DIFF
--- a/inkless-sync/README.md
+++ b/inkless-sync/README.md
@@ -56,12 +56,18 @@ Context for the agent:
 ./inkless-sync/main-sync.sh
 ```
 
-### Sync Before New Kafka Version
+### Pin main to a release-branch cut
 
 ```bash
-# Sync to the trunk commit where the 4.3 release branch diverged from trunk
+# Last trunk commit that is also on apache/4.3 (4.3.0-SNAPSHOT).
+# Use this to cut inkless-4.3; do not pass 4.4.
 ./inkless-sync/main-sync.sh --before-version 4.3
 ```
+
+`--before-version 4.3` is the merge base of `apache/trunk` and `apache/4.3`, not a
+tag parent. Trunk commits after that cut never landed on `apache/4.3` (or landed
+later as cherry-picks with different SHAs). Merging them into main would put them
+on `inkless-4.3`.
 
 ### Dry Run (Preview)
 
--- a/inkless-sync/README.md
+++ b/inkless-sync/README.md
@@ -59,7 +59,7 @@ Context for the agent:
 ### Sync Before New Kafka Version
 
 ```bash
-# Sync to the commit just before a version tag (e.g., before 4.3)
+# Sync to the trunk commit where the 4.3 release branch diverged from trunk
 ./inkless-sync/main-sync.sh --before-version 4.3
 ```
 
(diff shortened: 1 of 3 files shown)


inkless-sync/README.md
```diff
@@ -59,7 +59,7 @@ Context for the agent:
 ### Sync Before New Kafka Version
```
jeqo:
```suggestion
### Pin main to a release-branch cut
```

inkless-sync/README.md
```diff
@@ -59,7 +59,7 @@ Context for the agent:
...
 
 ```bash
-# Sync to the commit just before a version tag (e.g., before 4.3)
+# Sync to the trunk commit where the 4.3 release branch diverged from trunk
 ./inkless-sync/main-sync.sh --before-version 4.3
 ```
```
jeqo, line 65:
```suggestion
```

`--before-version 4.3` is the merge base of `apache/trunk` and `apache/4.3`, not a tag parent. Trunk commits after that cut never landed on `apache/4.3` (or landed later as cherry-picks with different SHAs). Merging them into main would put them on `inkless-4.3`.

```

File: inkless-sync/README.md [hunk 3]
```diff
@@ -59,7 +59,7 @@ Context for the agent:
 ### Sync Before New Kafka Version
 
 ```bash
-# Sync to the commit just before a version tag (e.g., before 4.3)
+# Sync to the trunk commit where the 4.3 release branch diverged from trunk
```
jeqo:
```suggestion
# Last trunk commit that is also on apache/4.3 (4.3.0-SNAPSHOT).
# Use this to cut inkless-4.3; do not pass 4.4.
```

inkless-sync/main-sync.sh
```diff
@@ -161,15 +162,16 @@ phase_prepare() {
 
     # Determine sync target
     if [[ -n "$BEFORE_VERSION" ]]; then
-        log_info "Finding last commit before version ${BEFORE_VERSION}..."
-        # Find the commit just before the version tag
-        if ! tag_exists "${BEFORE_VERSION}"; then
-            log_error "Tag ${BEFORE_VERSION} not found"
+        log_info "Finding trunk commit before version ${BEFORE_VERSION} diverged..."
```
jeqo: align it with change
```suggestion
        log_info "Finding merge base of ${APACHE_REMOTE}/trunk and ${APACHE_REMOTE}/${BEFORE_VERSION}..."
```

jeqo (commented):
LGTM. Added a few doc/logging suggestions

jeqo (approved):

PR #772, opened 2026-08-27
fix(inkless:consume): size fetch buffers to the payload [KC-481]

jeqo:
The diskless fetch path allocated a 1 MiB scratch buffer per object to carry a payload that is usually tens of kilobytes. A lagging fetch touches about one object per batch, so that allocation dominates the read path.

The size the caller already has isn't that payload. On the hot path `FixedBlockAlignment` requests a full `inkless.consume.cache.block.bytes` block (16 MiB by default) with no clip to the object. Sizing to `range.bufferSize()` would allocate a full block to deliver tens of kilobytes, since the object itself is capped by `produce.buffer.max.bytes`. Every backend already knows the true length before anything reads the channel. This PR sizes the allocation to that length.

`readToByteBuffer` drained every channel into 1 MiB scratches and copied. This PR changes that in two steps:

1. The scratch path fills each buffer before allocating the next, so buffer count tracks the payload, not the read count.
2. A channel can declare its exact content length. All four backends do. `readToByteBuffer` then allocates once and reads straight into that destination.

The scratch path stays for any channel that doesn't declare a length. After this PR no backend uses it: S3, GCS, Azure, and in-memory all declare a length. `BaseStorageTest` asserts that, so dropping a declaration fails the shared tests instead of silently allocating a 1 MiB scratch per fetch. Deleting the fallback means narrowing `fetch` to `SizedReadableByteChannel` and making `wrap` always sized. That change is broad (the fetch signature, wrap, the consume-test stubs, and the scratch-path suite) and is a follow-up.

## Impact

The two paths differ by more than the numbers below suggest, and it is worth saying why: the same 1 MiB scratch buffer was proportionate on one and dominant on the other.

**Hot path.** `FixedBlockAlignment` requests a full cache block (16 MiB by default), which is always at least the object size, since an object is capped by `produce.buffer.max.bytes` (8 MiB by default). A hot fetch therefore reads essentially the whole object: a 4 MiB object cost four scratch buffers plus the EOF probe, about 1.25x overhead. Add the cache in front of it - a hot object is fetched once and reused - and the cost was real but not visible.

**Lagging path.** `BoundingRangeAlignment` collapses an object's ranges to one min-to-max span, and for a lagging consumer that span is only its own partition's slice of an object shared with many partitions: tens of kilobytes. The same 1 MiB buffer is then 16-25x the payload. Three things compound:

1. The scratch-to-payload ratio inverts, as above.
2. Objects per fetch tracks batch count, because `planJobs` groups by object key across all partitions and then collapses each object to one range. So that ratio is paid per object, many times per fetch request.
3. The cold path bypasses the cache, so every fetch pays it again rather than amortizing over the cache lifetime.

The fixed EOF-probe buffer is the clearest illustration: a full 1 MiB regardless of payload, so for a small cold read it was half the allocation on its own. That is what the first commit removes, before the sized path removes the rest.

64 KiB is the common cache-aligned fetch against a small object. 8 MiB is the object-size cap. S3 and GCS already filled a whole scratch per read. Azure used to stream through `BlobInputStream` and the JDK 8 KiB channel adapter; that path is gone.

- 64 KiB dest: 2.06 MiB (S3/GCS) or ~9 MiB (Azure-as-stream) → 64 KiB.
- 8 MiB dest: 17 MiB (S3/GCS) or ~1 GiB (Azure-as-stream) → 8 MiB.

The sized path also drops the extra copy of the payload. The reader fills the destination from the channel instead of joining scratches. An empty range no longer allocates 1 MiB to read nothing.

GCS `setChunkSize(0)` also drops the reader's staging buffer. A positive chunk size allocated a second buffer of that size (8 MiB by default) on first read; the client now writes straight into the dest.

S3 and Azure hand the materialized response to `SizedReadableByteChannel.of(ByteBuffer)`. The old JDK channel over an `InputStream` moved every byte through an 8 KiB staging array, so an 8 MiB fetch paid about a thousand reads and a full extra copy of a payload that was already exactly sized. Azure also drops `BlobInputStream`'s 4 MiB chunk.

No new or modified config, no new or renamed metrics, no API or protocol change.

## Operator impact

**GCS fetch TTFB drops, and dashboards or alerts on it need re-baselining.** `TimingReadableByteChannel` records time-to-first-byte on the first read that returns data. The buffered GCS reader filled the destination before returning, so with a 1 MiB scratch the first read came back only after about 1 MiB (or the whole payload) had transferred, and TTFB was close to whole-fetch time for a small range. The unbuffered reader returns on the first socket read, so the reported number is now much smaller and much closer to actual time-to-first-byte. Nothing else about the metric changed: same name, same units, same call site.

S3 TTFB is unaffected: it materializes the response inside `fetch`, so its first read already returned after the whole transfer. Azure now does the same, so its TTFB also becomes whole-download time - Azure isn't in production, so there is nothing to re-baseline there.

**Which metric catches a bad fetch.** A backend that declares a content length it doesn't deliver throws `IOException` from `readToByteBuffer`. That is wrapped as `FileFetchException`, then converted to a per-object `Failure`. The FETCH completes. The partition is `KAFKA_STORAGE_ERROR` and marks `PartitionStorageErrorRate`. `FileFetchErrorRate` is only marked when the whole Reader future fails, which this path does not. Content corruption is unaffected and already covered - `createMemoryRecords` calls `ensureValid()` before patching offsets, so a CRC or size failure is classified per partition as `CORRUPT_MESSAGE` and marks `PartitionCorruptRecordRate` while other partitions in the same request still succeed. A missing extent marks `PartitionStorageErrorRate`, a short batch count marks `PartitionPartialFetchRate`.

One pre-existing limitation worth naming, unchanged by this PR: those per-partition counters distinguish the failure shapes, but at request level only the generic `FailedFetchRequestsPerSec` is marked, so a steady trickle can't be attributed to one error class from request-level metrics alone.

## Commits

- `fix(inkless:consume): fill fetch scratch buffers before allocating another`: bound the scratch path. Fill each scratch before allocating the next.
- `feat(inkless:consume): read sized channels into one exact allocation`: the sized channel, the exact destination, `wrap`, and in-memory adopting it.
- `fix(inkless:storage): align in-memory ranged fetches with the other backends`: clip over-runs, short-circuit empty ranges, and read the clip through `of(ByteBuffer)`.
- `feat(inkless:storage): report content length from S3 fetches`: S3 declares the length it already knew and reads its response buffer directly.
- `feat(inkless:storage): report content length from GCS fetches`: GCS declares the length it already knew and reads straight into the dest (`setChunkSize(0)`).
- `feat(inkless:storage): report content length from Azure fetches`: Azure downloads the range with `downloadContentWithResponse` and reads that buffer directly. The shared tests assert every backend declares a length.

See the commit messages for the reasoning behind each step.

## Test plan

`ObjectFetcherTest` is new. `RecordingReadChannel` is a shared fixture: it caps every read at 8 KiB and records dest identity, so the assertions check allocation, not only output bytes.

Unsized reads cover empty through just over 2 MiB. Scratch count tracks payload size: one 1 MiB scratch below 1 MiB, a second scratch at exactly 1 MiB to observe EOF. Sized reads expect dest capacity equal to the payload plus a 1-byte over-delivery probe. An empty sized channel is the probe only. An upper bound on capacity would stay green with a leftover 1 MiB scratch once the payload is at least 1 MiB.

Both paths retry a `0`-byte read that isn't EOF. A declared length that doesn't match what the channel delivers fails. Both `of` factories throw `ClosedChannelException` after `close`. The delegate factory forwards `close` and `isOpen` and rejects a null delegate. The 8 KiB cap on `RecordingReadChannel` is a short-read stand-in; Azure no longer returns 8 KiB per `read()`.

`FileFetchJobTest.testSizedChannelReachesTheReader` runs the same fake through the real job. The job wraps the channel for TTFB, then calls the real `readToByteBuffer`. Dest plus probe capacities prove the wrapper carried the length. Without that, the dest would be a 1 MiB scratch and the test would still see the right bytes.

`TimingReadableByteChannelTest` covers `wrap`: a sized channel keeps its length, a plain channel doesn't claim one. The TTFB callback fires only on a strictly positive read.

`BaseStorageTest` gains an offset-0 over-run (the cache-aligned shape), `testFetchWithMaxRange` (`ByteRange.maxRange()`), and `testFetchDeclaresContentLength` on the exact range, two clipped ranges, and two empty ranges including an offset past the object end, so each backend's length arithmetic is checked against the same expectations. Azure no longer overrides the out-of-range case: `available <= 0` matches the other backends and the fixture case runs as-is. `InMemoryStorage` gained the empty-range short-circuit the other three already had, so an empty range is an empty read on every backend rather than an out-of-range failure on one. `InMemoryStorageTest` extends the fixture, which is what surfaced the off-by-one.

These fixture cases run on every backend: in-memory and the GCS and Azure container suites under `:storage:inkless:test`, and the S3 container suite under `:storage:inkless:integrationTest`. `S3StorageTest.testUploadUndersizedStream` executes for the first time in this PR - it carried `@Override` without `@Test`, which JUnit does not inherit, so it had never run since it was written.

## Scope

This lands the per-object allocation fix and nothing else. No config, no metrics, no API change.

The scratch-buffer fallback is the non-production path. Deleting it means narrowing `fetch` to `SizedReadableByteChannel` and making `wrap` always sized. That change is broad: the fetch signature, wrap, the consume-test stubs, and the scratch-path suite. It is a follow-up.

It doesn't bound how many fetched extents a broker holds at once. A request still assembles every extent before returning any of them. That peak is untouched here and is being worked on separately.

[KC-481]

[KC-481]: https://aiven.atlassian.net/browse/KC-481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/ObjectFetcher.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/ObjectFetcher.java
@@ -41,17 +41,19 @@ public interface ObjectFetcher extends Closeable {
     /**
      * Reads the channel into a single buffer.
      *
-     * <p>Each scratch buffer is filled before another is allocated.
-     * Otherwise, a channel that returns small amounts of data per read causes
-     * one 1 MiB allocation per read,
-     * which creates unnecessary garbage-collection and out-of-memory pressure.
+     * <p>If the channel is a {@link SizedReadableByteChannel}, the destination is that length.
+     * A mismatch fails the fetch. Otherwise the channel is drained into 1 MiB scratch buffers
+     * that are filled before another is allocated, then copied.
      *
      * <p>A {@code 0}-byte read is not EOF.
-     * The loop retries until {@code tempBuffer} is full or the channel returns {@code -1},
-     * so the channel must eventually return data or EOF.
+     * Both paths retry until the destination is full or the channel returns {@code -1},
+     * so a channel must eventually return data or EOF.
+     * A channel that returns {@code 0} indefinitely never completes the read.
      */
     default ByteBuffer readToByteBuffer(final ReadableByteChannel readableByteChannel) throws IOException {
-        final ByteBuffer byteBuffer;
+        if (readableByteChannel instanceof SizedReadableByteChannel sized) {
+            return readExactly(sized, sized.contentLength());
+        }
         final List<ByteBuffer> buffers = new ArrayList<>(5);
         int readSize;
         int totalSize = 0;
@@ -66,8 +68,27 @@ public interface ObjectFetcher extends Closeable {
                 totalSize += tempBuffer.remaining();
             }
         } while (readSize >= 0);
-        byteBuffer = ByteBuffer.allocate(totalSize);
+        final ByteBuffer byteBuffer = ByteBuffer.allocate(totalSize);
         buffers.forEach(byteBuffer::put);
         return byteBuffer.flip();
+    }
+
+    private static ByteBuffer readExactly(final ReadableByteChannel channel, final int contentLength)
+        throws IOException {
+        final ByteBuffer buffer = ByteBuffer.allocate(contentLength);
+        int readSize = 0;
+        while (buffer.hasRemaining() && readSize >= 0) {
+            readSize = channel.read(buffer);
+        }
+        if (buffer.hasRemaining()) {
+            throw new IOException(
+                "Channel delivered " + buffer.position() + " of " + contentLength + " bytes");
+        }
+        // Only a strictly positive read proves over-delivery.
+        // A 0 means no data is available yet, which is not evidence either way.
+        if (channel.read(ByteBuffer.allocate(1)) > 0) {
+            throw new IOException("Channel delivered more than " + contentLength + " bytes");
+        }
+        return buffer.flip();
     }
 }
--- a/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/ObjectFetcher.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/ObjectFetcher.java
@@ -30,14 +30,26 @@ import io.aiven.inkless.common.ObjectKey;
 public interface ObjectFetcher extends Closeable {
 
     /**
-     * Use large enough buffer for reading the blob content to byte buffers to reduce required number
-     * of allocations. Usually default implementations in input streams use 16 KiB buffers. The expectation
-     * of blob sizes from cloud storages are multi-megabyte.
+     * Use a large enough buffer when reading blob content to reduce the number of allocations.
+     * Cloud storage blobs are expected to be multiple megabytes,
+     * while channels may return much less data per read.
      */
     int READ_BUFFER_1MiB = 1024 * 1024;
 
     ReadableByteChannel fetch(ObjectKey key, ByteRange range) throws StorageBackendException, IOException;
 
+    /**
+     * Reads the channel into a single buffer.
+     *
+     * <p>Each scratch buffer is filled before another is allocated.
+     * Otherwise, a channel that returns small amounts of data per read causes
+     * one 1 MiB allocation per read,
+     * which creates unnecessary garbage-collection and out-of-memory pressure.
+     *
+     * <p>A {@code 0}-byte read is not EOF.
+     * The loop retries until {@code tempBuffer} is full or the channel returns {@code -1},
+     * so the channel must eventually return data or EOF.
+     */
     default ByteBuffer readToByteBuffer(final ReadableByteChannel readableByteChannel) throws IOException {
         final ByteBuffer byteBuffer;
         final List<ByteBuffer> buffers = new ArrayList<>(5);
@@ -45,11 +57,13 @@ public interface ObjectFetcher extends Closeable {
         int totalSize = 0;
         do {
             final ByteBuffer tempBuffer = ByteBuffer.allocate(READ_BUFFER_1MiB);
-            readSize = readableByteChannel.read(tempBuffer);
-            if (readSize > 0) {
+            do {
+                readSize = readableByteChannel.read(tempBuffer);
+            } while (readSize >= 0 && tempBuffer.hasRemaining());
+            if (tempBuffer.position() > 0) {
                 buffers.add(tempBuffer);
                 tempBuffer.flip();
-                totalSize += readSize;
+                totalSize += tempBuffer.remaining();
             }
         } while (readSize >= 0);
         byteBuffer = ByteBuffer.allocate(totalSize);
--- a/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/TimingReadableByteChannel.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/TimingReadableByteChannel.java
@@ -55,7 +55,18 @@ public class TimingReadableByteChannel implements ReadableByteChannel {
     // Not volatile: channel is confined to a single thread within FileFetchJob.doWork().
     private boolean firstByteRecorded;
 
-    public TimingReadableByteChannel(ReadableByteChannel delegate, Time time, Instant startTime, Consumer<Long> ttfbCallback) {
+    /**
+     * Returns a timing decorator that preserves the delegate's {@link SizedReadableByteChannel}
+     * capability, so a sized channel still reaches {@link ObjectFetcher#readToByteBuffer}.
+     */
+    public static TimingReadableByteChannel wrap(ReadableByteChannel delegate, Time time, Instant startTime, Consumer<Long> ttfbCallback) {
+        if (delegate instanceof SizedReadableByteChannel sized) {
+            return new Sized(sized, time, startTime, ttfbCallback);
+        }
+        return new TimingReadableByteChannel(delegate, time, startTime, ttfbCallback);
+    }
+
+    private TimingReadableByteChannel(ReadableByteChannel delegate, Time time, Instant startTime, Consumer<Long> ttfbCallback) {
         this.delegate = Objects.requireNonNull(delegate, "delegate");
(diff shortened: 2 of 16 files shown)


storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/TimingReadableByteChannel.java
```diff
@@ -55,7 +55,18 @@ public class TimingReadableByteChannel implements ReadableByteChannel {
...
+     * capability, so a sized channel still reaches {@link ObjectFetcher#readToByteBuffer}.
+     */
+    public static TimingReadableByteChannel wrap(ReadableByteChannel delegate, Time time, Instant startTime, Consumer<Long> ttfbCallback) {
+        if (delegate instanceof SizedReadableByteChannel sized) {
+            return new Sized(sized, time, startTime, ttfbCallback);
+        }
+        return new TimingReadableByteChannel(delegate, time, startTime, ttfbCallback);
+    }
+
+    private TimingReadableByteChannel(ReadableByteChannel delegate, Time time, Instant startTime, Consumer<Long> ttfbCallback) {
```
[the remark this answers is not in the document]
jeqo: io.aiven.inkless isn't published API — "no API change" in the description refers to the Kafka protocol and configs.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/ObjectFetcher.java
```diff
@@ -30,30 +30,65 @@
...
+        }
+        if (buffer.hasRemaining()) {
+            throw new IOException(
+                "Channel delivered " + buffer.position() + " of " + contentLength + " bytes");
+        }
+        // Only a strictly positive read proves over-delivery.
+        // A 0 means no data is available yet, which is not evidence either way.
+        if (channel.read(ByteBuffer.allocate(1)) > 0) {
+            throw new IOException("Channel delivered more than " + contentLength + " bytes");
+        }
```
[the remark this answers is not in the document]
jeqo: Retrying the probe would be unbounded, and it runs after the destination is already full — the channel has delivered exactly contentLength and the read has succeeded. A channel that returns 0 indefinitely would then hang a completed fetch, which is worse than the gap it closes. The data-corrupting case is under-delivery, and that's caught deterministically by buffer.hasRemaining() after the loop, not by the probe. Over-delivery means the backend lied about its declared length; we return exactly what it declared, and the probe catches the lie whenever the extra byte is available. Keeping the strictly-positive check.

jeqo:
Measured this on a dev env to size the impact. It removes the scratch-buffer allocation entirely and lands on the theoretical floor for the read path.

**Measured** — 120 s async-profiler allocation capture during a lagging-consumer drain, one topic / 128 partitions, 1 KiB messages, 50 MB/s. Attribution by stack frame, normalised per byte of payload because the two windows carried different load:

- `ObjectFetcher` read path: 3.30 → **1.04** bytes per payload byte (**3.19x** less).
- Fetch-read total, including `FetchCompleter.extractRecords`: 5.38 → 3.37 (1.60x less).

The 1 MiB scratch buffers go from 12,005 events to zero — that bucket is absent from the histogram, not merely smaller. 1.04 is the floor: one allocation of exactly the payload, plus array headers.

Also gone, which I had not expected: the JDK `Channels.newChannel(InputStream)` staging buffers. `TimingReadableByteChannel.read` dropped from ~1.1M to 312 bytes and the 8 KiB bucket disappeared.

**The win is larger on smaller payloads.** The old path allocated `ceil(payload / 1 MiB) + 1` scratches plus one destination, so overhead scales inversely with payload. The measured ~1 MB payloads sit at the flat end (3.19x). A typical lagging-path slice is tens of kilobytes, which is 16–25x on the same formula. The measured point sits on that curve.

**What this does not claim.** It measures allocation rate, not retained memory. In-flight fetch extents are a separate term, so this should not be expected to stop an OOM on its own. Bounding in-flight extent bytes is the next lever.

Nor is it a whole-broker allocation reduction. `FetchCompleter.extractRecords` is now the largest allocator in the fetch path, which is why fetch-read improvement stops at 1.60x. That is the assembly copy from extents into the response — data already in memory, copied again — and looks like a worthwhile follow-up, separate from this PR.

giuseppelillo (approved):

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
```diff
@@ -155,13 +148,20 @@ public ReadableByteChannel fetch(ObjectKey key, ByteRange range) throws StorageB
...
+            // A chunk size of 0 makes the client read straight into the destination buffer.
+            // Any positive size stages a second buffer of that size first,
+            // and the caller already allocates a destination the size of the whole payload.
+            reader.setChunkSize(0);
             if (range != null) {
                 reader.limit(range.endOffset() + 1);
                 reader.seek(range.offset());
             }
-            return reader;
+            return SizedReadableByteChannel.of(reader, Math.toIntExact(contentLength));
```
viktorsomogyi, line 164: `contentLength` is a `long` and when it's larger than what can be represented on an `int`, it'll throw an `ArithmeticException` which gets around the catch. I think we should instead do a check and throw `StorageBackendException` if it's larger than `Integer.MAX_VALUE`, or just catch and rethrow as StorageBackendException. WDYT?

viktorsomogyi: Ah, it's merged, then maybe a followup if you think it's warranted 😄

jeqo: @viktorsomogyi I've created a follow-up fix: https://github.com/aiven/inkless/pull/778

PR #774, opened 2026-08-28
fix(inkless:consume): serve consumer reads from a lagging AZ replica [KC-461]

jeqo:
A consumer in the same availability zone (AZ) can land on a consolidating replica that has the requested records on disk but has not committed them yet. The local log then returns an empty response, so the consumer looks caught up. This change serves those reads from object storage until the replica's high watermark moves past the fetch offset.

The routing decision used the log end offset, which is not a bound the local read can honor: a consumer read stops at the high watermark. Every fetch for an offset between the two was sent to the local log, which had nothing to return. Followers still read through the log end offset, so they keep copying uncommitted records and the high watermark can still advance.

## How it reached consumers

What the consumer saw next depended on the Fetch API version it speaks, because that is what decides whether the broker will serve it from a replica at all.

A client on Fetch v11 or later gets client metadata attached to its request (`KafkaApis` populates it for v11+, with or without `client.rack`), so leader-only enforcement is off and the read is allowed on the replica. It came back empty, and the consumer made progress again once the high watermark advanced: slow and silent, but self-correcting.

A client on an older Fetch version sends no client metadata, which makes the broker enforce leader-only reads. The existing override for that case covered only topics switched from classic, so on a born-diskless consolidating topic the read was refused with `NOT_LEADER_OR_FOLLOWER`. The client refreshed metadata, the transformer pointed it back at the same replica, and the partition stopped progressing while reporting no error to either side. That is the shape KC-461 was reported as: a partition that reads as empty while its data is committed and durable.

Neither condition appears in our test setups. They use the Java client, so client metadata is always present, and single-replica topics make the advertised replica the leader, so leader-only enforcement never engages. Reproducing the refusal needs a client that negotiates Fetch below v11 against a topic with more than one replica.

## Commits

This pull request is nine commits. See the commit messages for the reasoning behind each step:

- `refactor(inkless:consume): one helper for the consolidating-topic check`: six inline spellings, one of them with the operands the other way round, become one topic-scoped helper. No behavior change.
- `fix(inkless:consume): wait for local HW on AZ replicas`: serve consumer fetches from object storage when the replica has not committed the offset. Followers keep reading the local log.
- `fix(inkless:consume): keep the seal window leader-only`: while a switch is pending there is no diskless leg to fall back on, so a replica read can answer out-of-range for records the leader holds, and a supplement can only come back empty.
- `fix(inkless:consume): keep parked fetches waiting on AZ replicas`: a waiting fetch on that replica is no longer completed as "this broker is not the leader".
- `test(inkless:consume): cover below-seal reads on a recovered replica`: a replica that lost its local data still returns records from before the switch.
- `feat(inkless:consume): surface local HW lag on consolidating replicas`: a gauge for partitions in that state and a meter for the consumer fetches that land in it.
- `test(inkless:consume): cover a pre-v11 consumer read from a replica`: the reported failure, reproduced against a running cluster.
- `feat(inkless:consume): report replicas missing the classic prefix`: a lag gauge for what a replica has not replicated from before the switch offset, and a meter for the reads it cannot serve.
- `fix(inkless:consume): wait out a below-seal offset this replica lacks`: those reads wait for the replica instead of resetting the consumer.

## Operator impact

While a replica's high watermark lags its log end offset, consumer reads go to object storage. `ConsolidationLocalBytesPerSec`, `ConsolidationSupplementRate`, and `ConsolidationSupplementBytesPerSec` fall. Diskless fetch volume and object-storage GET cost rise.

Two new metrics report that state:

- `ConsolidationHighWatermarkLagPartitionCount`: partitions on this broker whose local high watermark sits below their local log end offset. Alert on this. It holds whether or not anyone is reading.
- `ConsolidationHighWatermarkLagFetchRate`: consumer fetches that land in the window, one mark per affected partition. Use it to size the impact.
- `DisklessSwitchedPrefixLag`: records from before the switch offset that this broker's replicas have not replicated. Zero means every replica here holds them. A value that falls is a replica catching up; one that stops falling is a replica that has stopped.
- `DisklessSwitchedPrefixMissingFetchRate`: consumer reads that hit that range. They wait for the replica now rather than resetting, so this is the only signal that they are waiting.

A small value on either is normal on a busy topic, because the high watermark trails an append briefly. Followers never increment the meter, and neither do fetches while a switch is pending, which are served locally.

Neither replaces `ConsolidationLocalLag`. That measures how far behind consolidation is in copying diskless data into the local log. It can read zero while the high watermark sits below the local log end offset.

Follower copies of records from before the switch no longer count toward `ConsolidationLocalBytesPerSec`. Re-baseline dashboards you key on that meter.

No config changes. Two new metrics.

## Test plan

`ReplicaManagerInklessTest` adds cases on `fetchMessages` and the delayed-fetch path. Each case fails if you remove the production check it covers:

- A consumer asking for an offset the replica has written but not committed is served from object storage, and `ConsolidationHighWatermarkLagFetchRate` advances by one.
- The gauge counts a partition in that state against a real log (log end 10, checkpointed high watermark 4), and stops counting it once the high watermark advances.
- A follower asking for those same offsets is served from the local log, and the meter does not move. Nor does it move for a fetch while the switch is pending, which is served locally too.
- A fetch for records that live only in remote storage stays on the local path.
- A fetch for records from before the topic switched stays on the local path and does not merge object-storage data. Above the replica's high watermark that reply is empty and returns at once, so the consumer re-fetches until the high watermark advances.
- A consumer whose request carries no client metadata may read from a non-leader replica of a consolidating topic.
- A share fetch, and a fetch while the switch is still pending, still require the partition leader.
- A fetch on that non-leader replica reaches the fetch purgatory instead of being completed as "this broker is not the leader".
- A read for records from before the switch that this replica has not replicated waits in the purgatory and returns empty, rather than out of range. It fails if any of the three sites disagree: an error from the read, an immediate empty response from the purgatory, or a supplement issued below the switch offset.
- The lag is what a replica has not replicated from before the switch offset: it falls as the replica catches up, is zero once it arrives, and does not go negative when a consolidated suffix puts the log end offset past that point.

`InklessConsolidatedDisklessTopicsTest.testPreKip392ConsumerReadsConsolidatedPartitionFromReplica` reproduces the reported failure against a running cluster. It produces to a born-consolidated topic, waits for both replicas to consolidate and commit the range, then fetches at Fetch v10 from the replica that is not the KRaft leader. Without the change it fails with `NOT_LEADER_OR_FOLLOWER`. No current client sends a request that old, so `PreKip392FetchClient` builds one and sends it to a single broker's listener. Fetch v10 also predates topic ids, so this covers the topic-id backfill too; the assertions keep the two apart, `NOT_LEADER_OR_FOLLOWER` for the replica read and `UNKNOWN_TOPIC_ID` for the backfill.

`InklessConsolidatedDisklessTopicsTest.testWipedReplicaServesBelowSealReadAtRequestedOffset` wipes a follower's log directory, restarts the broker, and reads an offset from before the switch on that replica. The first record in the reply is the requested offset, not the switch offset. A route that answered from object storage and dropped the range in between fails this test.

`DelayedFetchTest` and `ReplicaManagerTest` cover the unchanged classic snapshot path through the shared helper.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -27,7 +27,7 @@ import io.aiven.inkless.consolidation.{ConsolidatedDisklessLogPruner, Consolidat
 import kafka.cluster.Partition
 import kafka.log.LogManager
 import kafka.server.QuotaFactory.QuotaManagers
-import kafka.server.ReplicaManager.{AtMinIsrPartitionCountMetricName, ConsolidationFetchBytesInPerSecMetricName, ConsolidationLocalBytesPerSecMetricName, ConsolidationSupplementBytesPerSecMetricName, ConsolidationSupplementRateMetricName, FailedIsrUpdatesPerSecMetricName, IsrExpandsPerSecMetricName, IsrShrinksPerSecMetricName, LeaderCountMetricName, OfflineReplicaCountMetricName, PartitionCountMetricName, PartitionsWithLateTransactionsCountMetricName, ProducerIdCountMetricName, ReassigningPartitionsMetricName, SealedPartitionsCountMetricName, DisklessSwitchedReplicasOutsideIsrCountMetricName, UnderMinIsrPartitionCountMetricName, UnderReplicatedPartitionsMetricName, createLogReadResult, isListOffsetsTimestampUnsupported}
+import kafka.server.ReplicaManager.{AtMinIsrPartitionCountMetricName, ConsolidationFetchBytesInPerSecMetricName, ConsolidationHighWatermarkLagFetchRateMetricName, ConsolidationHighWatermarkLagPartitionCountMetricName, ConsolidationLocalBytesPerSecMetricName, ConsolidationSupplementBytesPerSecMetricName, ConsolidationSupplementRateMetricName, FailedIsrUpdatesPerSecMetricName, IsrExpandsPerSecMetricName, IsrShrinksPerSecMetricName, LeaderCountMetricName, OfflineReplicaCountMetricName, PartitionCountMetricName, PartitionsWithLateTransactionsCountMetricName, ProducerIdCountMetricName, ReassigningPartitionsMetricName, SealedPartitionsCountMetricName, DisklessSwitchedReplicasOutsideIsrCountMetricName, UnderMinIsrPartitionCountMetricName, UnderReplicatedPartitionsMetricName, createLogReadResult, isListOffsetsTimestampUnsupported}
 import kafka.server.metadata.InklessMetadataView
 import kafka.server.share.DelayedShareFetch
 import kafka.utils._
@@ -118,6 +118,9 @@ object ReplicaManager {
   private val ConsolidationSupplementRateMetricName = "ConsolidationSupplementRate"
   private val ConsolidationSupplementBytesPerSecMetricName = "ConsolidationSupplementBytesPerSec"
   private val ConsolidationLocalBytesPerSecMetricName = "ConsolidationLocalBytesPerSec"
+  private val ConsolidationHighWatermarkLagFetchRateMetricName = "ConsolidationHighWatermarkLagFetchRate"
+  private val ConsolidationHighWatermarkLagPartitionCountMetricName =
+    "ConsolidationHighWatermarkLagPartitionCount"
 
   private[server] val GaugeMetricNames = Set(
     LeaderCountMetricName,
@@ -129,6 +132,7 @@ object ReplicaManager {
     ReassigningPartitionsMetricName,
     SealedPartitionsCountMetricName,
     DisklessSwitchedReplicasOutsideIsrCountMetricName,
+    ConsolidationHighWatermarkLagPartitionCountMetricName,
     PartitionsWithLateTransactionsCountMetricName,
     ProducerIdCountMetricName
   )
@@ -140,7 +144,8 @@ object ReplicaManager {
     ConsolidationFetchBytesInPerSecMetricName,
     ConsolidationSupplementRateMetricName,
     ConsolidationSupplementBytesPerSecMetricName,
-    ConsolidationLocalBytesPerSecMetricName
+    ConsolidationLocalBytesPerSecMetricName,
+    ConsolidationHighWatermarkLagFetchRateMetricName
   )
 
   private[server] val MetricNames = GaugeMetricNames.union(MeterMetricNames)
@@ -395,12 +400,26 @@ class ReplicaManager(val config: KafkaConfig,
   metricsGroup.newGauge(SealedPartitionsCountMetricName, () => sealedPartitionsCount)
   metricsGroup.newGauge(DisklessSwitchedReplicasOutsideIsrCountMetricName,
     () => disklessSwitchedReplicasOutsideIsrCount)
+  metricsGroup.newGauge(ConsolidationHighWatermarkLagPartitionCountMetricName,
+    () => consolidationHighWatermarkLagPartitionCount)
   metricsGroup.newGauge(PartitionsWithLateTransactionsCountMetricName, () => lateTransactionsCount)
   metricsGroup.newGauge(ProducerIdCountMetricName, () => producerIdCount)
 
   private def reassigningPartitionsCount: Int = leaderPartitionsIterator.count(_.isReassigning)
 
   private def sealedPartitionsCount: Int = leaderPartitionsIterator.count(_.isSealed)
+
+  /**
+   * Counts consolidating partitions holding data they cannot serve yet, that is a local high
+   * watermark below the local log end offset. Consumer reads of that range go to Inkless, which
+   * costs object-storage traffic, and below a classic-to-diskless seal Inkless has nothing to serve.
+   * Unlike ConsolidationHighWatermarkLagFetchRate this does not depend on consumer traffic, so a
+   * stalled replica stays visible after consumers have moved past its log end offset.
+   */
+  private[server] def consolidationHighWatermarkLagPartitionCount: Int = onlinePartitionsIterator.count { partition =>
+    consolidationActiveFor(partition.topic) &&
+      partition.log.exists(log => log.highWatermark < log.logEndOffset)
+  }
 
   private def lateTransactionsCount: Int = {
     val currentTimeMs = time.milliseconds()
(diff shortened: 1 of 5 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2923,6 +2986,50 @@ class ReplicaManager(val config: KafkaConfig,
...
+   *
+   * Every path serving such a fetch must reach the same answer, `DelayedFetch` included: its
+   * offset-snapshot check also runs with `fetchOnlyLeader`, and a `NotLeaderOrFollowerException`
+   * there force-completes the parked request (case A) instead of waiting for the high watermark or
+   * `minBytes`.
+   *
+   * Share fetches are excluded on the flag rather than on convention: they set
+   * `CONSUMER_REPLICA_ID`, so `isFromConsumer` holds, and only `KafkaApis` always supplying
+   * `clientMetadata` keeps them leader-only today.
+   */
```
viktorsomogyi, line 3028: Seems a bit excessive for a private method. I think we should keep such documentation minimal to avoid code-doc drift.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2923,6 +2986,50 @@ class ReplicaManager(val config: KafkaConfig,
...
+   *
+   * Switch-pending (`-2`) partitions are excluded. Those serve every offset from the local log
+   * because diskless holds nothing until the seal commits, so a fetch above a lagging replica's log
+   * end offset has no diskless path to fall back on and would answer OFFSET_OUT_OF_RANGE for records
+   * the leader still holds. Keeping the leader-only requirement for the seal window makes the
+   * consumer retry until the switch completes instead.
+   *
+   * Managed replicas are not tested here: `KafkaConfig.validateValues` requires
+   * `diskless.managed.replicas.enable` whenever consolidation is enabled.
+   */
```
viktorsomogyi, line 3004: Seems a bit excessive for a private method. I think we should keep such documentation minimal to avoid code-doc drift.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -443,13 +448,27 @@ class ReplicaManager(val config: KafkaConfig,
...
 
+  /**
+   * Counts consolidating partitions holding data they cannot serve yet, that is a local high
+   * watermark below the local log end offset. Consumer reads of that range go to Inkless, which
+   * costs object-storage traffic, and below a classic-to-diskless seal Inkless has nothing to serve.
+   * Unlike ConsolidationHighWatermarkLagFetchRate this does not depend on consumer traffic, so a
+   * stalled replica stays visible after consumers have moved past its log end offset.
+   */
+  private[server] def consolidationHighWatermarkLagPartitionCount: Int = onlinePartitionsIterator.count { partition =>
+    consolidationActiveFor(partition.topic) &&
```
viktorsomogyi: Seems like the gauge counts switching partitions with no seal offset (or seal pending). This may cause false alerts. These partitions correctly fetch until HW and don't go to the diskless storage, yet it seems like they're included in the gauge.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -2470,13 +2496,37 @@ class ReplicaManager(val config: KafkaConfig,
...
+              // Two ranges must stay on the local read path even when they sit above the isolation
+              // frontier, because diskless cannot answer them: [logStartOffset, localLogStartOffset)
+              // lives only in the remote tier once ConsolidatedDisklessLogPruner has pruned the
+              // diskless batches, and [0, classicToDisklessStartOffset) is the classic prefix the
+              // control plane never held. Both switch sentinels are negative, so a partition that
+              // never switched keeps the frontier at the isolation bound. Without a local log there
+              // is nothing to read locally and diskless is authoritative for the whole range.
+              val readableFrontier =
+                if (localLog.isEmpty) 0L
+                else Math.max(Math.max(localReadFrontier, localLogStartOffset), classicToDisklessStartOffset)
```
viktorsomogyi, line 2535: It's more of a general comment but the related code part might be these 3 lines. So let's assume the following situation: 2 replicas, the leader has all offsets but the follower has seal at 60, consumer at 55, this replica at LEO = 10, HW = 10 due to some data recovery event.

1. readableFrontier = max(10, 0, 60) = 60
2. 55 < 60 → local path.
3. UnifiedLog then sees fetch 55 and LEO = 10. In Kafka, fetchOffset > logEndOffset is OFFSET_OUT_OF_RANGE.

This fetch offset is valid on the leader, however the consumer was sent here by AZ metadata, which advertises this replica as the leader. This causes an offset reset in the consumer. I think we should carve this case out and return NONE to the consumers instead of an error.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Agreed, fair concern. Trimming this

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: and this one too.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: Good catch, I've missed the switch scenario here. Fixing this by excluding pending switches.

core/src/main/scala/kafka/server/ReplicaManager.scala
jeqo: You are right. This seems to be the same behavior on main, but with frontier at LEO, so 55 goes diskless which does not contain anything below the seal, so it also answers with out-of-range.

Yes, empty response is a better response here. As the metadata transformer lacks context to filter out a replica in this state, replying with an error that triggers a metadata refresh gets us nowhere, since the refresh returns the same replica.
I'm adding some metrics here to know when we are in this state, so we can alert if it's making progress or not; then reply with empty response.
Also covered the empty-and-wait instead of empty-and-immediate reply which would cause a hot re-fetch loop.

We could consider improving the scenario where if a replica is recovering, and this may take long, then the transformer prefers replicas that hold the prefix. This way we could reply with NOT_LEADER_OR_FOLLOWER error and trigger a metadata refresh on the client to target a healthy replica; but this should be a follow-up.

I'll update the PR body to specify the actual coverage.

viktorsomogyi: > We could consider improving the scenario where if a replica is recovering, and this may take long, then the transformer prefers replicas that hold the prefix. This way we could reply with NOT_LEADER_OR_FOLLOWER error and trigger a metadata refresh on the client to target a healthy replica; but this should be a follow-up.

I agree with this, although this improvement likely would target the metadata transformer I guess, to advertise the best replicas that are caught up for lagging consumers.

viktorsomogyi (dismissed):
LGTM, I think we can merge it after the conflicts are resolved.

viktorsomogyi (approved):

PR #776, opened 2026-09-01
docs(inkless:sync): add sync PR merge steps to main-sync plan

jeqo:
The Recommended Workflow stopped at Step 9 with no push or merge step, so the ruleset handling a sync merge needs was undocumented and had to be rediscovered.

Two rules in the `main` ruleset block a sync PR. Require signed commits fires because upstream Apache commits are occasionally unsigned; signing one rewrites its SHA, which destroys the merge base with apache/kafka and makes every later sync conflict, so the rule has to come off for the merge. Require linear history forbids the merge commit, so no merge-commit button appears even though the ruleset's own merge-method list still shows Merge.

Add Step 10 with the GitHub UI path: clear both rules, confirm the repo-level Allow merge commits toggle, merge with Create a merge commit, then restore the rules. Squash and rebase are called out as unsafe here because they flatten the upstream commits and drop the merge commit the next sync reads as its merge base.

### DIFF
--- a/inkless-sync/MAIN-SYNC-ACTION-PLAN.md
+++ b/inkless-sync/MAIN-SYNC-ACTION-PLAN.md
@@ -112,6 +112,11 @@ git commit -m "sync(test): [description]"
 
 #### Step 10: Merge the sync PR (10 min)
 
+**A repository admin has to merge a sync PR.** Both settings this step changes, the `main` ruleset
+and the repository's merge options, are admin-only. The `maintain` role can't reach them. If you
+ran the sync and aren't an admin, take the PR to review as usual and then hand the merge to an
+admin. Ask internally who currently holds the role.
+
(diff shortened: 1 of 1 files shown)


tvainika:
Instructions look clear, but those ruleset configurations do not seem to be available to everyone (including me). So sync-merging is available only to this repo admins, right?

jeqo:
@jlprat could we check @tvainika claim? I think it's correct, we just need to make sure the list of admins is somewhere so we know who to ping when a merge is needed.

tvainika (changes requested):
To change (temporarily) rules, one needs to be anyway Admin. Can you document this so that only admins can merge sync PRs? No need to document here how to find admins, I think that can be done internally. Then I think we are fine with this process.

tvainika (approved):

PR #778, opened 2026-09-01
fix(inkless:storage): report an oversized fetch as a storage failure [KC-481]

jeqo:
GcsStorage narrowed its long content length with Math.toIntExact, which raises an unchecked ArithmeticException reading "integer overflow". It escapes the StorageBackendException that fetch declares, and names neither the object nor the length.

The int is a property of the SizedReadableByteChannel contract, not of GCS: the payload lands in a single ByteBuffer and a byte[] on the extent. exactLength owns that rule so no backend narrows on its own, and the message carries the key and the length.

InMemoryStorage keeps its own narrowing: it bounds the length by an in-heap byte[], so it cannot exceed the int range by construction.

Unreachable at default caps, since object and range sizes stay well below 2 GiB. Both caps are operator-settable and nothing validates them against this ceiling, so the failure should be diagnosable if anyone raises them.

Testing

SizedReadableByteChannelExactLengthTest covers the boundary: 0, a normal length and Integer.MAX_VALUE pass through; MAX_VALUE + 1 and a negative raise StorageBackendException naming the object and the length.

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/SizedReadableByteChannel.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/common/SizedReadableByteChannel.java
@@ -23,6 +23,8 @@ import java.nio.channels.ClosedChannelException;
 import java.nio.channels.ReadableByteChannel;
 import java.util.Objects;
 
+import io.aiven.inkless.common.ObjectKey;
+
 /**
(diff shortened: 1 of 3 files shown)


viktorsomogyi (approved):
Noticed 2 somewhat unrelated things but I'm gonna file them as a follow-up.

PR #783, opened 2026-09-03
feat(inkless:storage): bound and shorten the GCS upload path

jeqo:
GCS requests were bounded only by the google-cloud client defaults (20s connect, 20s read), neither of which was configurable, and every upload opened a resumable session and paid two round trips. Both costs land on the produce path: a stalled or slow upload holds an uploader thread, shows up as `FileUploadTime`, and cascades into produce latency through `FileCommitWaitTime`. This makes the timeouts configurable, bounds each response read by default, and drops one round trip for small objects. See the commit messages for the reasoning behind each step.

## Commits

- `feat(inkless:storage): bound each GCS request attempt by default` - adds `gcs.connect.timeout` and `gcs.read.timeout`, defaulting to 2s and 1s; maps a timed out request to `StorageBackendTimeoutException`; and reshapes the batch delete so it chunks its work, reports per-key failures, and no longer confirms a key the server refused to delete.
- `feat(inkless:storage): upload GCS objects in a single request` - uses `storage.create` for objects up to 1 MiB, keeps the resumable path above that, and adds an `object-upload` counter so per-request upload activity stays visible.
- `feat(inkless:storage): count GCS error responses by type` - adds `throttling-errors`, `server-errors`, and `other-errors` under the S3 backend's names, counted per request attempt from the response interceptor.

## Config changes

- `gcs.connect.timeout` (new): connection establishment timeout in milliseconds. Default 2000.
- `gcs.read.timeout` (new): response read timeout in milliseconds. Default 1000.

Both mirror the existing `s3.api.call.timeout` and `s3.api.call.attempt.timeout` knobs, except that these carry real defaults rather than deferring to the client's. An operator who wants the old behavior sets the client's values explicitly (20000 each); there is no null or sentinel, because a plain number already expresses it.

Three properties of these bounds are worth knowing before tuning them:

- **The read value bounds each socket read, not a whole request.** Connection setup falls under `gcs.connect.timeout`, request-body writes are not covered, and a response arriving in several reads can exceed it.
- **Whether it bounds the operation depends on the client's idempotency classification.** A single-request upload and a resumable session initiate are non-idempotent, and the client's non-idempotent handler retries nothing, so their ceiling is ours to set: an object at or below 1 MiB ends its attempt on the first stalled read, making the upload budget `produce.max.upload.attempts` stalled reads, about 3s. A read and a resumable upload's chunk write are idempotent and run up to 6 client attempts under a 50s total timeout, so above 1 MiB an upload attempt is bounded by that budget rather than by this value, and lowering the read timeout changes how many attempts fit in the 50s rather than shortening it. Removing the client's retry on the produce path is left as a follow-up.
- **Both values apply to every GCS role.** `SharedState` builds four independent backends (fetch, lagging fetch, produce, background) from the same `inkless.storage.*` config. Connect keeps the wider bound on purpose: it applies only on a connection-pool miss, so a generous ceiling costs nothing in steady state and leaves headroom for the connection churn that accompanies a slowdown, while the read bound applies to every stalled request and is the one worth keeping tight.

A timed out request surfaces as `StorageBackendTimeoutException` from `upload`, single-key `delete`, and a fetch's metadata request. A fetch's body is streamed from the channel `fetch` returns, so a stall mid-download reaches the caller as a plain `IOException` instead -- `ReadableByteChannel.read` cannot throw `StorageBackendException`. `S3Storage` classifies its whole fetch only because it materializes the response into a buffer. The batch delete stops the pass rather than classifying, as S3 does. When an upload exhausts its attempts the file fails, and every produce request in it gets `KAFKA_STORAGE_ERROR`. That is a `RefreshRetriableException` at the client, so producers retry after a metadata refresh rather than losing data.

## Metric changes

- `object-upload-rate` / `object-upload-total` (new): single-request object uploads. `resumable-upload-initiate` and `resumable-chunk-upload` stop advancing for objects at or below the threshold, so dashboards that track upload activity through those two need this one added.
- `throttling-errors`, `server-errors`, `other-errors` (new, each `-rate` / `-total`): error responses per request attempt, under the S3 backend's names so one dashboard serves both. 429 and 503 are throttling, as `SlowDown` is on S3; other 5xx are server errors; other non-2xx, including the 404 for a missing key, are other errors. A request that fails without a response (a configured timeout, a connection reset) is not counted; see the follow-ups.

## Operator impact

Objects up to 1 MiB now issue one request instead of two. With `produce.buffer.max.bytes` at 8 MiB, that is the small-file end of the distribution: a broker rotating full buffers under load keeps streaming, while a low-throughput topic flushing a few kilobytes every `produce.commit.interval.ms` saves a round trip, which is where the fixed cost dominates. The threshold sits below the measured break-even near 2 MiB, above which a single request costs more per megabyte than the round trip it saves.

A GCS request that stalls beyond the new timeouts fails its attempt rather than holding a connection for 20s. For an object at or below 1 MiB, where nothing is retried inside the client, three upload attempts now complete in about 3s instead of a minute. Above the threshold the client's own 50s retry budget dominates and the read timeout does not shorten the upload, so expect `FileUploadTime` maxima to drop at the small end of the distribution and stay where they are at the large end.

A stalled object download is a second place the client's budget wins. `ApiaryUnbufferedReadableByteChannel.read` reopens the stream on a retryable failure and each reopen starts a fresh budget, so a shorter read timeout makes it reopen more often against a backend that is already slow rather than giving up sooner. Bounding a stalled fetch needs a deadline above the channel; see the follow-ups.

Batch deletes change shape rather than budget. The client splits a delete into HTTP batches of at most 100 sub-requests and answers only after processing all of them, so the response wait scales with the batch, and a batch delete is not retried by the client. `GcsStorage.delete(Set)` now deletes in chunks of half that maximum and returns the keys confirmed so far, so a slow batch costs one chunk instead of a whole cleaner pass. A default cycle of 20000 files becomes 400 round trips, well inside `file.cleaner.interval.ms`. `FileCleaner` already treats deletion as partial, so this matches the contract it expects and the S3 backend's shape.

The same rewrite fixes a confirmation bug the previous shape had. `Storage.delete(Iterable)` reports a failed sub-request as the same `false` its callback uses for an absent blob, and the old code ignored the result and confirmed every key in the batch, so a 403 or 5xx on one key still dereferenced its file in the control plane and left a permanent orphan. Driving `Storage.batch()` directly keeps the three outcomes apart: deleted and already-absent are confirmed, a failure is not, and failures are now logged by response code the way the S3 backend logs them. `object-delete` counts confirmed deletions rather than attempted ones, so it can read lower than before on a bucket that is refusing deletes.

## Testing

`GcsStorageConfigTest` covers the new defaults, that both values parse into `Duration` objects, and that a non-positive value is rejected. `GcsStorageMetricsTest` asserts the upload counters on both sides of the 1 MiB threshold. `GcsStorageTest` covers a delete spanning several batches, which the shared suite's ten-key case does not reach, and overrides `testUploadOversizeStream`, since a stream longer than its declared length is now reported by the trailing-byte probe rather than by the size check, matching the S3 backend.

`GcsErrorHandlingTest` uses WireMock for the failures fake-gcs-server cannot produce. A delayed response pins the timeout classification: an upload surfaces `StorageBackendTimeoutException` with `SocketTimeoutException` as its root cause, which exercises the cause-chain walk every classification site depends on. A hand-built multipart batch response pins the delete confirmation on both sides: with one deleted sub-response and two refused, only one key comes back confirmed, and with every sub-response a 404, both keys do. Each error sensor is asserted at exactly one count end to end: a 429 and a 500 on upload, a 404 on fetch, and a 503 on the batch endpoint, which the interceptor never sees and the caller reports.

## Follow-ups

- `configured-timeout-errors` and `io-errors` are still missing on GCS, so nothing reports how often these timeouts fire per attempt. The AWS SDK publishes those itself; the GCS client has no once-per-attempt hook that carries the exception (the response interceptor never runs on an `IOException`, `HttpIOExceptionHandler` receives no exception, `NetHttpTransport` is final, and gax consults the retry strategy up to three times per attempt). Settling where that hook lives is its own change, and it is what would let these defaults be validated in production rather than inferred from `FileUploadTime`.
- A stalled GCS object download is neither classified nor bounded. The body is streamed after `fetch` returns, so a stall reaches the caller as a plain `IOException`, and the client's read channel reopens the stream on each retryable failure with a fresh retry budget, so no number of read timeouts ends the fetch. Both need a wall-clock deadline above the channel, around `ObjectFetcher.readToByteBuffer`, which can throw `StorageBackendException`. Shared across backends, so it belongs in its own change.
- Above 1 MiB an upload is retried twice over: by the client's idempotent handler on the resumable PUT and by `FileUploadJob` on top, while a fetch has only the client's retry. The fix is no client retry on the produce backend and the default on the fetch backend, which needs per-role storage config, since `SharedState` builds all four backends from the same `inkless.storage.*` map. `produce.upload.backoff.ms` (10ms, unjittered) has to be re-sized in the same change: it is masked by the client's backoff today and becomes the only wait between attempts once that is gone. Both should follow the error counters, because removing the client's retry trades a slow success for a fast `KAFKA_STORAGE_ERROR`, and how often a stalled PUT recovers on a retry is the number that decides whether that trade is worth making.

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
@@ -30,9 +30,16 @@ import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 import com.groupcdg.pitest.annotations.CoverageIgnore;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.SocketTimeoutException;
 import java.nio.channels.ReadableByteChannel;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -45,9 +52,18 @@ import io.aiven.inkless.storage_backend.common.KeyNotFoundException;
 import io.aiven.inkless.storage_backend.common.SizedReadableByteChannel;
 import io.aiven.inkless.storage_backend.common.StorageBackend;
 import io.aiven.inkless.storage_backend.common.StorageBackendException;
+import io.aiven.inkless.storage_backend.common.StorageBackendTimeoutException;
 
 @CoverageIgnore  // tested on integration level
 public class GcsStorage extends StorageBackend {
+    private static final Logger LOGGER = LoggerFactory.getLogger(GcsStorage.class);
+
+    // The client splits a delete into HTTP batches of at most 100 sub-requests, and the server has to
+    // process every sub-request before it answers, so the response wait grows with the batch. Half the
+    // client's maximum keeps that wait clear of gcs.read.timeout; a default cleaner cycle of 20000
+    // files then costs 400 round trips, well inside file.cleaner.interval.ms.
+    private static final int MAX_DELETE_BATCH_SIZE = 50;
+
     private volatile Storage storage;
     private String bucketName;
     private ReloadableCredentialsProvider credentialsProvider;
@@ -70,6 +86,8 @@ public class GcsStorage extends StorageBackend {
         this.bucketName = config.bucketName();
 
         final HttpTransportOptions.Builder httpTransportOptionsBuilder = HttpTransportOptions.newBuilder();
+        httpTransportOptionsBuilder.setConnectTimeout((int) config.connectTimeout().toMillis());
+        httpTransportOptionsBuilder.setReadTimeout((int) config.readTimeout().toMillis());
 
         // Create reloadable credentials provider
         this.credentialsProvider = config.reloadableCredentials();
@@ -104,8 +122,24 @@ public class GcsStorage extends StorageBackend {
                         "Object " + key + " created with incorrect length " + transferred + " instead of " + length);
             }
         } catch (final IOException | BaseServiceException e) {
+            if (isTimeout(e)) {
+                throw new StorageBackendTimeoutException("Timed out to upload " + key, e);
+            }
             throw new StorageBackendException("Failed to upload " + key, e);
         }
+    }
+
+    // The GCS client wraps the originating SocketTimeoutException in a BaseServiceException, so the
+    // whole cause chain is inspected.
+    private static boolean isTimeout(final Throwable e) {
+        Throwable cause = e;
+        while (cause != null) {
+            if (cause instanceof SocketTimeoutException) {
+                return true;
+            }
+            cause = cause.getCause() == cause ? null : cause.getCause();
+        }
+        return false;
     }
 
     @Override
@@ -113,28 +147,44 @@ public class GcsStorage extends StorageBackend {
         try {
             storage.delete(this.bucketName, key.value());
         } catch (final BaseServiceException e) {
+            if (isTimeout(e)) {
+                throw new StorageBackendTimeoutException("Timed out to delete " + key, e);
+            }
             throw new StorageBackendException("Failed to delete " + key, e);
         }
     }
 
     @Override
     public Set<ObjectKey> delete(final Set<ObjectKey> keys) throws StorageBackendException {
-        try {
-            final Set<BlobId> ids = keys.stream()
-                    .map(k -> BlobId.of(this.bucketName,k.value()))
-                    .collect(Collectors.toSet());
+        final List<ObjectKey> objectKeys = new ArrayList<>(keys);
+        final Set<ObjectKey> deleted = new HashSet<>();
 
-            // storage.delete returns a List<Boolean> of deleted-vs-already-absent, but a genuine
-            // failure surfaces as a thrown BaseServiceException rather than a per-blob flag, so we
-            // cannot extract a confirmed-deleted subset the way the S3 backend does. This stays
-            // all-or-nothing: on success every key is gone (idempotent), and on failure we delete
-            // nothing and let the FileCleaner cycle retry the whole set.
-            storage.delete(ids);
-            metricCollector.recordBatchDeleteObjects(keys.size());
-            return Set.copyOf(keys);
-        } catch (final BaseServiceException e) {
-            throw new StorageBackendException("Failed to delete " + keys.size() + " keys", e);
+        for (int i = 0; i < objectKeys.size(); i += MAX_DELETE_BATCH_SIZE) {
+            final List<ObjectKey> batch = objectKeys.subList(
+                i,
+                Math.min(i + MAX_DELETE_BATCH_SIZE, objectKeys.size())
+            );
+            final Set<BlobId> ids = batch.stream()
+                    .map(k -> BlobId.of(this.bucketName, k.value()))
+                    .collect(Collectors.toSet());
+            try {
+                // storage.delete returns a List<Boolean> of deleted-vs-already-absent, but a genuine
+                // failure surfaces as a thrown BaseServiceException rather than a per-blob flag, so a
+                // batch stays all-or-nothing: on success every key in it is gone (idempotent), and on
+                // failure none of it is confirmed.
+                storage.delete(ids);
+            } catch (final BaseServiceException e) {
+                // Deletion is idempotent, so stopping here is safe: the keys left unconfirmed stay
+                // marked for deletion and the next FileCleaner cycle retries them.
+                LOGGER.warn("Batch delete failed after {} of {} keys, stopping the pass",
+                    deleted.size(), objectKeys.size(), e);
+                break;
+            }
+            deleted.addAll(batch);
+            metricCollector.recordBatchDeleteObjects(batch.size());
         }
+
+        return deleted;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
```diff
@@ -97,44 +122,86 @@ public void upload(final ObjectKey key, final InputStream inputStream, final lon
...
+            );
+            final Set<BlobId> ids = batch.stream()
+                    .map(k -> BlobId.of(this.bucketName, k.value()))
+                    .collect(Collectors.toSet());
+            try {
+                // storage.delete returns a List<Boolean> of deleted-vs-already-absent, but a genuine
+                // failure surfaces as a thrown BaseServiceException rather than a per-blob flag, so a
+                // batch stays all-or-nothing: on success every key in it is gone (idempotent), and on
+                // failure none of it is confirmed.
+                storage.delete(ids);
```
giuseppelillo: `Storage.delete(Iterable)` does not throw for individual failed subrequests, but ti returns false when deletion fails or access is denied. Because the result is ignored and the entire batch is added to `deleted` at line 200, `FileCleaner` can dereference objects that were never deleted, leaving permanent orphans.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
```diff
@@ -97,44 +122,86 @@ public void upload(final ObjectKey key, final InputStream inputStream, final lon
         }
         try {
             final BlobInfo blobInfo = BlobInfo.newBuilder(this.bucketName, key.value()).build();
-            Blob blob = storage.createFrom(blobInfo, inputStream);
+            final Blob blob;
+            if (length <= DIRECT_UPLOAD_MAX_SIZE) {
+                blob = storage.create(blobInfo, inputStream.readNBytes((int) length));
+            } else {
+                blob = storage.createFrom(blobInfo, inputStream);
+            }
```
giuseppelillo, line 136: This resumable branch still has internal SDK retries. After session initiation, GCS client retries chunk and finalization requests as idempotent operations under its retry budget. Therefore uploads above 1 MiB are not bounded to one `gcs.read.timeout` per FileUploadJob attempt. Either enforce an overall retry budget for resumable uploads or narrow the documented latency guarantee, and test a stalled final PUT after successful session initiation.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
```diff
@@ -174,6 +243,11 @@ public ReadableByteChannel fetch(ObjectKey key, ByteRange range) throws StorageB
             } else if (e.getCode() == 416) {
                 // https://cloud.google.com/storage/docs/json_api/v1/status-codes#416_Requested_Range_Not_Satisfiable
                 throw new InvalidRangeException("Failed to fetch " + key + ": Invalid range " + range, e);
+            } else if (isTimeout(e)) {
+                // Reaches only the metadata request. The body is streamed from the channel this
+                // method returns, so a stall mid-download surfaces to the caller as a plain
+                // IOException; ReadableByteChannel.read cannot throw StorageBackendException.
+                throw new StorageBackendTimeoutException("Timed out to fetch " + key, e);
```
giuseppelillo, line 283: A persistent body-read timeout may not surface as an `IOException`. `ApiaryUnbufferedReadableByteChannel.read` handles a retryable failure by reopening the stream in a loop, with each reopen starting a fresh retry budget. Repeated zero-byte stalls can therefore keep a fetch thread occupied indefinitely. Add an overall fetch deadline or cancellation boundary and cover repeated body stalls.

giuseppelillo (changes requested):
Raise some issues related to the already existing behavior, but I think it's still worth to fix those right now.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
jeqo: Good catch. I'm fixing this by using StorageBatch API to handle the response properly.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
jeqo: Right. Let's fix the docs for now. I'm amending the commit as well.

Motivated by this, but not actually resolving the issue here, is the handling of error metrics as it's done in S3 I'm adding on the last commit. It feels enough the scope of this PR to add it here.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/GcsStorage.java
jeqo: Good catch -- I think this needs to be solved at the caller so it's solved for all backends. Adding it as a follow-up 2 on the PR body.

giuseppelillo (approved):

PR #785, opened 2026-09-03
test(inkless:consolidation): tolerate re-delivery in the consume check

jeqo:
`consumeAndVerify` asserted that each delivered offset was exactly one past the previous one, walking raw delivery order, and counted raw deliveries to decide when to stop polling. Both assume every record arrives once.

Nothing guarantees that. The helper subscribes with a fresh group id and auto.offset.reset=earliest, so the consumer is group-managed: a rebalance before the first auto-commit lands (auto-commit is on by default, every five seconds) resets the partition and the stream restarts at its earliest offset. That is legal at-least-once delivery, and it breaks the helper twice over. Duplicates count toward the wait, so polling can stop before every distinct record has arrived, and the walk over delivery order then sees the offset go backwards.

CI caught it as `"Offset gap or reordering ... expected: <6> but was: <0>"` after a partition had delivered 0..5 and started again from 0. Four tests call the helper, across seven call sites, so any of them can fail this way.

Collect distinct offsets per partition in a sorted set, wait on the distinct count, and assert the distinct offsets form one unbroken run. A replay now adds nothing to the set and changes no outcome.

The check still means what it did: it fails when a partition is genuinely missing an offset, including a gap that a replay would otherwise pad over.

**Testing**

`InklessConsolidatedDisklessTopicsTest` passes across two runs of the full class. The contiguity logic was checked against the delivery sequence from the CI failure (0..5 then 0..5 again, which the previous check rejected and this one accepts), a clean run, a run with one offset missing, and a run with both a replay and a missing offset. The last two still fail.

### DIFF
--- a/core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
+++ b/core/src/test/java/kafka/server/InklessConsolidatedDisklessTopicsTest.java
@@ -81,7 +81,9 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableSet;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


viktorsomogyi (approved):

viktorsomogyi:
@jeqo the commit is unverified, please rebase it and then we can merge it.

PR #786, opened 2026-09-03
fix(inkless:control_plane): force generic find batches planning

giuseppelillo:
`find_batches_v2` runs its `batches`-to-`files` lookup once for every requested partition. With automatic plan caching, a PostgreSQL backend can select a custom plan for that statement and invoke the planner on every lookup, even though each invocation produces the same ordered index-scan plan. Planning then dominates replica CPU and makes performance depend on backend-local plan-cache history.

`plan_cache_mode=auto` (default) normally switches to a generic plan when its estimated cost is competitive with the average custom-plan cost. For this parameterized query, benchmarks showed that in some cases PostgreSQL keeps replanning equivalent index scans instead of reliably selecting the reusable generic plan.

Set `plan_cache_mode` to `force_generic_plan` at the function boundary to bypass that misleading cost decision. PostgreSQL now builds and reuses one generic inner plan per backend while `find_batches_v2` runs. The setting does not affect other control-plane statements, and the generic plan preserves the last-offset index order and budget-bounded early exit.

The migration test checks the function catalog configuration so a future function replacement cannot silently remove the setting.


storage/inkless/src/main/resources/db/migration/V28__Find_batches_generic_plan.sql
EelisK: This needs to be updated to `V29`. Currently it clashes with [V28__Soft_delete_logs.sql]( https://github.com/aiven/inkless/blob/a6716f703723bc52368638e88d595a20049ed137/storage/inkless/src/main/resources/db/migration/V28__Soft_delete_logs.sql)

PR #789, opened 2026-09-08
fix(inkless:storage): count retried GCS error responses

giuseppelillo:
## Background

[PR #783](https://github.com/aiven/inkless/pull/783) added GCS error counters through an `HttpResponseInterceptor`. The Google HTTP client invokes that interceptor only after `HttpRequest.execute()` finishes its internal retry loop. When GCS returns `401`, `HttpCredentialsAdapter` can refresh the token and retry the same request before the interceptor runs, so a successful refresh hides the failed response and repeated authorization failures are undercounted.

The credentials handler also reinstalls itself after a refresh. Wrapping it without restoring the metrics wrapper would therefore count only the first `401` from a request.

Batch requests add a second distinction. The outer POST bypasses the request initializer, but each failed multipart sub-response invokes the unsuccessful-response handler. A `404` delete means the object is already absent, and `StorageBatch` treats that outcome as a confirmed idempotent delete, so counting it as a provider error creates noise during cleaner retries. GCS still bills and rate-limits every submitted delete sub-request, including refused requests, so the operation counter must include all of them.

## Changes

- Record error statuses through an `HttpUnsuccessfulResponseHandler`, which runs before the client decides whether to retry each response.
- Delegate to the existing credentials handler and restore the metrics wrapper after it refreshes a token.
- Keep the response interceptor for operation counters, which should run once after the internal retry loop.
- Exclude an already-absent batch delete from `other-errors`, while continuing to count refused and throttled sub-requests.
- Count every returned batch delete sub-request as an `object-delete` operation; confirmed deletions remain available through cleaner metrics.
- Describe `other-errors` as non-`429` 4xx responses, excluding absent batch deletes. Protocol responses such as `308 Resume Incomplete` also remain excluded.

## Testing

- `./gradlew :storage:inkless:test --tests io.aiven.inkless.storage_backend.gcs.MetricCollectorTest --tests io.aiven.inkless.storage_backend.gcs.integration.GcsErrorHandlingTest --tests io.aiven.inkless.storage_backend.gcs.integration.GcsStorageMetricsTest`
- `./gradlew :storage:inkless:genInklessMetricsDoc`

### DIFF
--- a/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/MetricCollector.java
+++ b/storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/MetricCollector.java
@@ -30,11 +30,13 @@ import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.http.HttpResponse;
 import com.google.api.client.http.HttpResponseInterceptor;
+import com.google.api.client.http.HttpUnsuccessfulResponseHandler;
 import com.google.cloud.BaseServiceException;
 import com.google.cloud.ServiceOptions;
 import com.google.cloud.http.HttpTransportOptions;
 import com.groupcdg.pitest.annotations.CoverageIgnore;
 
+import java.io.IOException;
 import java.util.regex.Pattern;
 
 import static io.aiven.inkless.storage_backend.gcs.MetricRegistry.OBJECT_DELETE;
@@ -162,8 +164,8 @@ public class MetricCollector {
         return sensor;
     }
 
-    // The client builds the batch request without the request initializer, so the response interceptor
-    // never sees it; its caller reports both its operations and its failure.
+    // The client builds the batch request without the request initializer, so the metric handlers
+    // never see it; its caller reports both its operations and its failure.
     void recordBatchDeleteObjects(final int objectCount) {
         if (objectCount > 0) {
             deleteObjectRequests.record(objectCount);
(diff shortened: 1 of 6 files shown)


storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/MetricCollector.java
jeqo, line 1: This change makes error counters cover batch deletes, which #783 didn't. Each key in a batch gets
its own response, and the new handler sees every failed one; which the old interceptor didn't none. 

The comment on :167 was updated from "the response interceptor" to "the metric handlers" but it may not be aligned with the changes introduced. It says the handlers never see the batch request `GcsStorage.delete(Set)` sends. True for the outer POST, which is built without the initializer. Not true for the one sub-request per key inside it: those are built with the initializer, so the new handler runs for every key whose delete fails. The code is the behavior to keep; the comment and the tests should follow it. No test asserts it either way today: `batchDeleteConfirmsAbsentKeys` now ticks `other-errors` by 2 and still passes.

The one thing to decide is the 404 sub-request: the key was already gone, `delete(Set)` confirms it, and it now counts as an error. A cleaner cycle is 20000 keys, so it shows on the dashboard.
                                                                                                          
Count it (what the code does now): a re-delete after a crash between storage delete and control-plane
dereference ticks `other-errors` once per key with nothing wrong in GCS, and the number is
indistinguishable from refused deletes. No code change; fix the comment, assert `errorTotal` in both
batch tests.
                                                                                                          
Don't count it: skip a 404 when the response belongs to a sub-request (`response.getRequest() != request`
identifies one). One special case in the handler; differs from the single-key path, where a 404 counts;
for deletes it matches S3, which does not report missing keys as errors. Same comment and test changes,
plus the skip.
                                                                                                          
Either way, refused (403) and throttled (429) sub-requests count, which #783 did not have. I'd not count the 404 to match S3.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/MetricRegistry.java
jeqo, line 38: Proposal: count every sub-request as an object-delete operation, recordBatchDeleteObjects(results.size()) instead of confirmed.

Today this counts confirmed deletions only, while the error sensors now count every failed sub-request.
A batch of 50 keys a bucket refuses shows 50 errors and 0 operations. Counting all sub-requests makes it 50 and 50, matches the single-key path, and matches what GCS bills and rate-limits (each sub-request is one operation, success or not). Confirmed deletions stay available in the cleaner metrics.

One line in GcsStorage.delete(Set) plus object-delete-total asserted in the batch tests. #783 introduced this and #789 makes it visible, so it fits here; I can also do it as a one-line follow-up if you'd rather keep this PR to MetricCollector.

jeqo (commented):
Fix looks good. I left a couple of comments to decide the direction. Otherwise LGTM.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/MetricRegistry.java
giuseppelillo: Agreed. Commit cd67e69d28 records `results.size()` after a successful batch response, so every returned sub-request advances `object-delete`. Failed deletes remain excluded from the confirmed-key result and continue to be retried by `FileCleaner`. The batch tests now pin the operation totals.

storage/inkless/src/main/java/io/aiven/inkless/storage_backend/gcs/MetricCollector.java
giuseppelillo: Agreed. Commit cd67e69d28 skips `404` only for batch `DELETE` sub-responses, while refused and throttled sub-responses still advance their error counters. It also updates the batch instrumentation comment and adds exact assertions for both the `403` and already-absent cases.

jeqo (approved):

PR #792, opened 2026-09-08
fix(inkless:delete): do not sleep inside the topic purger tick [KC-349]

viktorsomogyi:
KafkaScheduler uses scheduleAtFixedRate. An idle or error sleep inside run() queues missed ticks and the next work cycle bursts past max.batches.per.cycle / interval. Empty cycles return immediately. Errors set nextEligibleMs and later ticks skip until that time.

InklessDisklessTopicDeleteTest asserts the per-cycle cap from TopicPurger logs instead of a COUNT(*) poll.

### DIFF
--- a/core/src/test/java/kafka/server/InklessDisklessTopicDeleteTest.java
+++ b/core/src/test/java/kafka/server/InklessDisklessTopicDeleteTest.java
@@ -32,6 +32,7 @@ import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.apache.kafka.common.test.KafkaClusterTestKit;
 import org.apache.kafka.common.test.TestKitNodes;
+import org.apache.kafka.common.utils.LogCaptureAppender;
 import org.apache.kafka.coordinator.group.GroupCoordinatorConfig;
 import org.apache.kafka.server.config.ServerConfigs;
 import org.apache.kafka.test.TestUtils;
@@ -50,10 +51,8 @@ import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
@@ -64,6 +63,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import io.aiven.inkless.config.InklessConfig;
 import io.aiven.inkless.control_plane.postgres.PostgresControlPlane;
 import io.aiven.inkless.control_plane.postgres.PostgresControlPlaneConfig;
+import io.aiven.inkless.delete.TopicPurger;
 import io.aiven.inkless.storage_backend.s3.S3Storage;
 import io.aiven.inkless.storage_backend.s3.S3StorageConfig;
 import io.aiven.inkless.test_utils.InklessPostgreSQLContainer;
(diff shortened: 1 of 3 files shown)


giuseppelillo (approved):
LGTM. Merging this one, but we have the same issue on `FileCleaner`, could you also fix that one?

storage/inkless/src/main/java/io/aiven/inkless/delete/TopicPurger.java
```diff
@@ -95,12 +94,13 @@ public void run() {
...
             attempts.set(0);
+            nextEligibleMs = 0L;
             metrics.recordTopicPurgerCycleSucceeded();
         } catch (final Exception e) {
             metrics.recordTopicPurgerError();
             final long backoff = errorBackoff.backoff(attempts.incrementAndGet());
-            LOGGER.error("Error while purging deleted logs, waiting for {}", Duration.ofMillis(backoff), e);
-            time.sleep(backoff);
+            nextEligibleMs = time.milliseconds() + backoff;
+            LOGGER.error("Error while purging deleted logs, retrying after {}", Duration.ofMillis(backoff), e);
```
jeqo, line 103: nit: 
```suggestion
            LOGGER.error("Error while purging deleted logs, skipping ticks for the next {}", Duration.ofMillis(backoff), e);
```

storage/inkless/src/main/java/io/aiven/inkless/delete/TopicPurger.java
```diff
@@ -95,12 +94,13 @@ public void run() {
             }
 
             attempts.set(0);
+            nextEligibleMs = 0L;
```
jeqo, line 97: At this point nextEligibleMs is already in the past, so should be fine to delete this one

storage/inkless/src/main/java/io/aiven/inkless/delete/TopicPurger.java
```diff
@@ -42,8 +41,11 @@ public class TopicPurger implements Runnable, Closeable {
     final int maxBatchesPerCycle;
     final TopicPurgerMetrics metrics;
     private final ExponentialBackoff errorBackoff = new ExponentialBackoff(100, 2, 60 * 1000, 0.2);
-    private final Supplier<Long> noWorkBackoffSupplier;
     private final AtomicInteger attempts = new AtomicInteger();
+    // KafkaScheduler uses scheduleAtFixedRate. Sleeping inside run() queues missed ticks, and
+    // the next work cycle bursts past topic.purger.max.batches.per.cycle / interval. Skip the
+    // tick until this time instead.
```
jeqo, line 47: nit: a shorter version
```suggestion
    // Do not sleep in run(): KafkaScheduler uses scheduleAtFixedRate, and a sleep queues missed
    // ticks that then fire back to back past topic.purger.max.batches.per.cycle.
```

core/src/test/java/kafka/server/InklessDisklessTopicDeleteTest.java
```diff
@@ -185,20 +185,23 @@ public void testDeleteTopicsReturnsBeforeBackgroundPurge() throws Exception {
             assertEquals(NUM_PARTITIONS, beforeDelete.logCount(), "Every partition must have a log row");
             assertEquals(0, beforeDelete.deletedLogCount(), "Live logs must not be stamped deleted");
 
-            admin.deleteTopics(Collections.singletonList(TOPIC_NAME)).all().get(30, TimeUnit.SECONDS);
+            try (LogCaptureAppender appender = LogCaptureAppender.createAndRegister(TopicPurger.class)) {
```
jeqo, line 188: nit: pin the log level explicitly
```suggestion
            try (LogCaptureAppender appender = LogCaptureAppender.createAndRegister(TopicPurger.class)) {
                appender.setClassLogger(TopicPurger.class, Level.INFO)
```

core/src/test/java/kafka/server/InklessDisklessTopicDeleteTest.java
```diff
@@ -275,40 +278,32 @@ private void waitUntilTopicGone(Admin admin) throws InterruptedException {
...
+     */
+    private static void assertSaturatedAcrossMultipleCycles(LogCaptureAppender appender, long batchesAtDelete) {
+        long saturated = appender.getMessages().stream()
+            .filter(message -> message.contains("per-cycle cap reached"))
+            .count();
+        assertTrue(saturated >= batchesAtDelete - NUM_BROKERS,
+            "TopicPurger must hit the per-cycle cap across multiple cycles (max "
+                + MAX_BATCHES_PER_CYCLE + " per cycle); saturated=" + saturated
+                + " batchesAtDelete=" + batchesAtDelete
+                + " messages=" + appender.getMessages());
```
jeqo, line 306: Maybe worth adding a doc for `batchesAtDelete - NUM_BROKERS` as it's not obvious where this is coming from.

We could also consider to move the saturated calculation to a wait if condition, to expect all brokers to do the logging:

```
final long expected = batchesAtDelete - 1; // every deleting cycle but the last hits the cap
TestUtils.waitForCondition(() -> countSaturated(appender) >= expected, 10_000,
    () -> "TopicPurger must hit the per-cycle cap on every cycle but the last; saturated="
        + countSaturated(appender) + " batchesAtDelete=" + batchesAtDelete
        + " messages=" + appender.getMessages());

private static long countSaturated(LogCaptureAppender appender) {
    return appender.getMessages().stream()
        .filter(message -> message.contains(TopicPurger.CAP_REACHED_LOG_FRAGMENT))
        .count();
}

```

jeqo (commented):
The fix looks good. Added a few comments.

Also, a related finding is that `FileCleaner` may have a similar issue as it's sleeping and it's scheduled thread. We could handle it here or in a separate PR, as you prefer.

PR #795, opened 2026-09-09
fix(inkless:delete): do not sleep inside the file cleaner tick [KC-349]

viktorsomogyi:


viktorsomogyi:
@jeqo @giuseppelillo addressed the comments from the earlier PR (#792 )

storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
```diff
@@ -54,13 +53,17 @@ public class FileCleaner implements Runnable, Closeable {
...
-    private final Supplier<Long> noWorkBackoffSupplier;
 
     /**
      * The counter of cleaning attempts.
      */
     private final AtomicInteger attempts = new AtomicInteger();
 
+    // KafkaScheduler uses scheduleAtFixedRate. Sleeping inside run() queues missed ticks, and
+    // the next work cycle bursts past file.cleaner.max.files.per.cycle / interval. Skip the
+    // tick until this time instead.
```
jeqo, line 64: This one is keeping the TopicPurger comment before the fix on this PR, let's align them?
```suggestion
    // Do not sleep in run(): KafkaScheduler uses scheduleAtFixedRate, and a sleep queues missed
    // ticks that then fire back to back past file.cleaner.max.files.per.cycle.
```

storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
```diff
@@ -133,12 +131,13 @@ public void run() {
             }
 
             attempts.set(0);
+            nextEligibleMs = 0L;
```
jeqo, line 134: same here, we can drop it

storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
```diff
@@ -133,12 +131,13 @@ public void run() {
...
             metrics.recordFileCleanerCycleSucceeded();
         } catch (final Exception e) {
             metrics.recordFileCleanerError();
             final long backoff = errorBackoff.backoff(attempts.incrementAndGet());
-            LOGGER.error("Error while deleting files, waiting for {}", Duration.ofMillis(backoff), e);
-            time.sleep(backoff);
+            nextEligibleMs = time.milliseconds() + backoff;
+            LOGGER.error("Error while deleting files, retrying after {}", Duration.ofMillis(backoff), e);
```
jeqo, line 140: and here :) 
```suggestion
            LOGGER.error("Error while deleting files, skipping ticks for the next {}", Duration.ofMillis(backoff), e);
```

core/src/test/java/kafka/server/InklessDisklessTopicDeleteTest.java
```diff
@@ -291,21 +293,26 @@ private void waitUntilPurged(Uuid topicId) throws InterruptedException {
     }
 
     /**
-     * One cycle cannot finish the drain: each broker deletes at most
-     * {@code topic.purger.max.batches.per.cycle} batches. Count saturated cycles from TopicPurger
-     * logs instead of a {@code COUNT(*)} poll, which can miss leftovers between 1s ticks.
+     * Every deleting cycle but the last hits the cap: {@code capReached} requires {@code moreRemain}.
+     * Two brokers can share the last pair of batches, so the floor is {@code batchesAtDelete - 1}
+     * (one finishing cycle), not {@code batchesAtDelete - NUM_BROKERS}.
```
jeqo, line 298: nit: reads a bit clearer, and mentions the log after commit that should be considered on the validation
```suggestion
     * With one batch per cycle, every deleting cycle hits the cap except the one that removes the
     * last deleted log: {@code capReached} requires {@code moreRemain}. Two brokers can finish
     * concurrently, but each probes {@code moreRemain} before it commits, so at most one of them
     * sees no log left. Hence {@code batchesAtDelete - 1} saturated lines.
     *
     * <p>The purger logs after its transaction commits, so poll the appender instead of counting
     * once after the DB reads zero.
```

storage/inkless/src/main/java/io/aiven/inkless/delete/TopicPurger.java
```diff
@@ -36,15 +36,17 @@
 public class TopicPurger implements Runnable, Closeable {
     private static final Logger LOGGER = LoggerFactory.getLogger(TopicPurger.class);
 
+    // Shared with InklessDisklessTopicDeleteTest so the saturation assertion cannot drift.
```
jeqo, line 39: nit: I wonder if keeping plain text is ok as failing test will capture the drift and force a fix -- otherwise either remove the comment as public makes the usage explicit or reduce it to
```suggestion
    // Shared with integration test.
```

storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
viktorsomogyi: Added in the latest batch

storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
viktorsomogyi: addressed in the last batch

storage/inkless/src/main/java/io/aiven/inkless/delete/FileCleaner.java
viktorsomogyi: addressed in the last batch

storage/inkless/src/main/java/io/aiven/inkless/delete/TopicPurger.java
viktorsomogyi: addressed in the last batch

giuseppelillo (approved):

PR #802, opened 2026-09-16
fix(inkless:consolidation): start consolidation when remote storage is enabled later [KC-540]

viktorsomogyi:
## Summary
Enabling `remote.storage.enable` on an already-diskless topic is a config-only write today, so brokers never re-enter become-leader and consolidation never arms. When cluster consolidation is on, the controller co-commits a leader-epoch bump with that `ConfigRecord`. Brokers re-enter become-leader, create the local log if needed, and start the consolidation fetcher. The bump is skipped when consolidation is off, the topic is not already diskless, or the same request starts a classic-to-diskless switch.
A WAL-gap fetch after that arm can land below the diskless WAL start while RLMM is still catching up. The endpoint proves a remote prefix from RLMM (or a cheap local positive): present-true returns `OFFSET_MOVED_TO_TIERED_STORAGE`, present-false leaves never-tiered `OFFSET_OUT_OF_RANGE`, and empty returns `NOT_LEADER_OR_FOLLOWER` so the fetcher retries at debug. `ConsolidationRemotePrefixUnknown` is 1 during that wait and 0 once decided. Register resets the latch; fetcher removal clears it without tearing down the lag gauges. Alert on a sustained non-zero after `__remote_log_metadata` catch-up; a leadership transition or an overlapping fetch can dip or briefly restore the value.
## Test plan
- [ ] Controller: late `remote.storage.enable=true` on an already-diskless topic emits epoch-bump `PartitionChangeRecord` objects in the same batch as the `ConfigRecord`.
- [ ] Controller: no bump when consolidation is off, the topic is classic, or the same request starts a switch.
- [ ] Broker: become-leader after the bump arms the consolidation fetcher for a born-diskless topic.
- [ ] WAL-gap fetch: RLMM present-true → `OFFSET_MOVED`; present-false → `OFFSET_OUT_OF_RANGE`; empty → `NOT_LEADER_OR_FOLLOWER`.
- [ ] Metrics: `registerPartition` resets `ConsolidationRemotePrefixUnknown`; fetcher removal clears the latch and leaves lag gauges registered.


core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -185,6 +207,24 @@ class DisklessLeaderEndPoint(
...
+   * Returns evidence of a remote-tier prefix the OFFSET_MOVED path can rebuild from.
+   * Present-true: copy has confirmed data (`highestOffsetInRemoteStorage >= 0`), a committed
+   * seal, or RLMM lists a live (not delete-finished) remote segment.
+   * Present-false: RLMM is ready and has no such segment, so the fetch stays OFFSET_OUT_OF_RANGE.
+   * Empty: RLMM is not ready or the query failed; the caller retries.
+   * `crossTierEarliestOffset` and a reported `remote_log_start_offset` can look like a prefix
+   * on a never-tiered born-diskless topic, so this method does not use them.
+   */
+  private def remoteConsolidatedPrefixEvidence(topicPartition: TopicPartition, localLog: UnifiedLog): Optional[java.lang.Boolean] = {
+    if (localLog.highestOffsetInRemoteStorage() >= 0 ||
```
giuseppelillo: The positive evidence is not tied to the offset that the tier-state rebuild needs. `TierStateMachine` looks up the remote segment containing `disklessStartOffset - 1`, but `highestOffsetInRemoteStorage >= 0`, a committed seal, or any unrelated segment can all return true without covering that boundary. For example, a legacy switched topic can have seal 100 and WAL start 200 after pure-diskless retention; enabling remote storage later makes the seal select `OFFSET_MOVED_TO_TIERED_STORAGE`, although offset 199 was never tiered, so recovery retries indefinitely. Check for readable remote coverage at `disklessStartOffset - 1` instead; the local highest-offset shortcut is only sufficient when it reaches that boundary.

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
```diff
@@ -6579,6 +6579,70 @@ class ReplicaManagerInklessTest {
...
+          oneReplica,
+          leaderEpoch = 0,
+        )
+        replicaManager.applyDelta(firstDelta, imageFromTopics(firstDelta.apply()))
+        verify(mockCfm, never()).addFetcherForPartitions(any())
+
+        when(replicaManager.inklessMetadataView().isConsolidatingDisklessTopic(topic)).thenReturn(true)
+        when(replicaManager.inklessMetadataView().isRemoteStorageEnabled(topic)).thenReturn(true)
+
+        val bumpDelta = createLeaderDelta(
```
giuseppelillo: This does not exercise the same-leader epoch-bump regression. `createLeaderDelta` always starts from `TopicsImage.EMPTY` and replays a complete `PartitionRecord`, so `bumpDelta.localChanges` treats the partition as newly created and reports it as a leader regardless of the bump. Build the second `TopicsDelta` from `firstDelta.apply()` and replay the actual `PartitionChangeRecord` with the existing leader. That verifies that the controller record increments the epochs, appears in `TopicDelta.localChanges`, and re-enters the broker activation path.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
viktorsomogyi: Thanks, good catch, addressed it in the latest set of patches.

core/src/test/scala/unit/kafka/server/ReplicaManagerInklessTest.scala
viktorsomogyi: Thanks, addressed this one too in the latest patch set.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -159,22 +159,44 @@ class DisklessLeaderEndPoint(
...
-                logger.debug("Offset {} for {} is below the diskless WAL start {} but at/above the whole-log start {}; " +
-                  "signalling OFFSET_MOVED_TO_TIERED_STORAGE to rebuild the consolidated remote prefix.",
-                  requestedOffset, tp.topicPartition, disklessStartOffset, logStartOffset)
-                fetchResponseData.setErrorCode(Errors.OFFSET_MOVED_TO_TIERED_STORAGE.code)
-                fetchResponseData.setRecords(MemoryRecords.EMPTY)
+              if (logStartOffset != UnifiedLog.UNKNOWN_OFFSET &&
+                  requestedOffset >= logStartOffset &&
+                  requestedOffset < disklessStartOffset) {
+                localLogOpt.filter(_.remoteLogEnabled()).foreach { _ =>
+                  val generation = replicaManager.consolidationRemotePrefixGeneration(tp.topicPartition)
```
giuseppelillo: The generation is captured after the potentially blocking fetch completes. A fetch can start, removal can bump the generation and clear the gauge while that fetch is blocked, and then the response path captures the new generation. Its stale `unknown = true` update is accepted, so the original race remains. Capture the generation before starting the fetch and carry it into response processing, or validate that the partition still has an active fetcher before publishing the metric. Add a test that removes the fetcher while the fetch is blocked.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationMetrics.scala
```diff
@@ -100,26 +116,70 @@ class ConsolidationMetrics extends Closeable {
...
+    Option(remotePrefixGenerationByPartition.get(tp)).exists(_.get == generation)
+
   def unregisterPartition(tp: TopicPartition): Unit = {
     val tags = Map("topic" -> tp.topic, "partition" -> tp.partition.toString).asJava
     totalLagByPartition.remove(tp)
     localLagByPartition.remove(tp)
     deletableByPartition.remove(tp)
     oversizedBatchByPartition.remove(tp)
+    remotePrefixUnknownByPartition.remove(tp)
+    remotePrefixGenerationByPartition.remove(tp)
```
giuseppelillo: `unregisterPartition` deletes the generation counter, so re-registration can reuse an old generation number. For example, an old fetch captures `1`; removal advances it to `2`; unregister deletes it; and re-register creates `1` again. When the old fetch completes, its generation matches the new registration and it can set the gauge incorrectly. Keep generations monotonic across re-registration or use globally unique registration tokens.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
viktorsomogyi: Attempted a fix for this.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationMetrics.scala
viktorsomogyi: Added a fix attempt.

core/src/main/scala/io/aiven/inkless/consolidation/DisklessLeaderEndPoint.scala
```diff
@@ -88,6 +88,12 @@ class DisklessLeaderEndPoint(
       topicNames.put(topic.topicId, topic.topic)
     }
     val fetchInfos = request.fetchData(topicNames.asJava)
+    // Capture before awaitDelayedFetch. The delayed op parks up to
+    // diskless.consolidation.fetch.max.wait.ms, and removeFetcherForPartitions
+    // can bump the generation while this thread is blocked.
+    val remotePrefixGenerations = fetchInfos.asScala.keys.iterator.map { tidp =>
```
giuseppelillo, line 97: This still leaves a race before the generation snapshot. `AbstractFetcherThread` builds the request while holding `partitionMapLock`, releases the lock, and only then calls `leader.fetch`. Removal can happen in that gap and bump the generation. The already-built stale request then enters here, captures the new generation, and its metric update is accepted. Capture the generation while building the request under `partitionMapLock` and carry it into response processing. Add a test that removes the partition after `buildFetch` returns but before `leader.fetch` starts.

viktorsomogyi: Added a fix for this.

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationMetrics.scala
```diff
@@ -100,26 +116,73 @@ class ConsolidationMetrics extends Closeable {
...
+    }
+  }
+
+  private def generationMatches(tp: TopicPartition, generation: Long): Boolean =
+    Option(remotePrefixGenerationByPartition.get(tp)).exists(_.get == generation)
+
   def unregisterPartition(tp: TopicPartition): Unit = {
+    // Bump and keep the generation counter. Deleting it would let re-register reuse 1
+    // and an in-flight fetch that captured 1 could restore the latch after stopPartitions.
+    bumpRemotePrefixGeneration(tp)
```
giuseppelillo, line 160: Non-blocking: Keeping the counter fixes generation reuse, but it retains the `TopicPartition` and topic name for every partition until broker shutdown. Repeated topic churn can therefore grow this map without bound. A broker-wide `AtomicLong` could assign globally unique generation tokens, allowing `unregisterPartition` to remove the per-partition entry without letting a later registration reuse an old token. This can also be addressed in a follow-up PR.

viktorsomogyi: Given the PR size, I think we should do it as a follow-up.

giuseppelillo (approved):

core/src/main/scala/io/aiven/inkless/consolidation/ConsolidationMetrics.scala
giuseppelillo: agreed, approved

PR #806, opened 2026-09-18
fix(inkless:consolidation): recover late enablement after WAL pruning

giuseppelillo:
When consolidation starts after pure-diskless retention advances the WAL start, an empty local log can persist `remote_log_start_offset=0`. The consolidation fetcher then loops between `OFFSET_OUT_OF_RANGE` and an `EARLIEST` response of 0.

This change defers the initial remote-log start report until remote data exists. If a born-diskless partition has no remote prefix, the broker repairs a previously persisted 0 by advancing the cross-tier start to the surviving WAL boundary. Switched partitions remain unchanged because their classic prefix can still exist outside remote storage.

## Test plan
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :storage:test --tests org.apache.kafka.server.log.remote.storage.RemoteLogManagerTest --no-daemon --no-parallel`
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :core:test --tests io.aiven.inkless.consolidation.DisklessLeaderEndPointTest --no-daemon --no-parallel`
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :core:test --tests kafka.server.ReplicaManagerInklessTest.testAdvanceBornDisklessCrossTierStartWritesThroughAndRejectsSwitchedPartition --no-daemon --no-parallel`


storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
```diff
@@ -992,6 +1003,22 @@ private void maybeUpdateLogStartOffsetOnBecomingLeader(UnifiedLog log) throws Re
...
 
+        private boolean hasRemoteLogSegmentInLeaderEpochHistory(UnifiedLog log) throws RemoteStorageException {
+            LeaderEpochFileCache leaderEpochCache = log.leaderEpochCache();
+            OptionalInt epoch = leaderEpochCache.earliestEntry()
+                    .map(entry -> OptionalInt.of(entry.epoch()))
+                    .orElseGet(OptionalInt::empty);
+            while (epoch.isPresent()) {
+                if (remoteLogMetadataManagerPlugin.get()
+                        .listRemoteLogSegments(topicIdPartition, epoch.getAsInt())
+                        .hasNext()) {
```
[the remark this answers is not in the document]
giuseppelillo: Fixed in f631e6b01a. The bootstrap now performs one leader-epoch scan, ignores transitional metadata, and reports the exact earliest COPY_SEGMENT_FINISHED start from that scan. It defers when no readable segment exists. The regression test covers empty, transitional-only, and mixed transitional/readable metadata.

storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
```diff
@@ -982,8 +982,22 @@ private void maybeUpdateLogStartOffsetOnBecomingLeader(UnifiedLog log) throws Re
                 // when the remote start is unreported, which would lock the wrong value in via the
                 // forward-only advance. No-op for classic topics.
                 OptionalLong override = logStartOffsetOverride.apply(topicIdPartition.topicPartition());
-                long logStartOffset = override.isPresent()
-                        ? override.getAsLong()
+                OptionalLong readableRemoteStart = OptionalLong.empty();
+                if (override.isEmpty() && isConsolidatingDisklessPartition.test(topicIdPartition.topicPartition())) {
```
viktorsomogyi: The condition covers every consolidating partition, including switched topics. If remote metadata has no readable segment, the method skips reporting the local classic start. Diskless retention can then advance the control-plane fallback while remote_log_start_offset remains unset, causing EARLIEST to skip the still-readable classic prefix. Even after a successful first copy, reporting waits until the next RLM cycle.

AI Recommendation: Distinguish partition origin and seal state. Report the local earliest for switched partitions that still own a classic prefix, while retaining deferral for late-enabled born-diskless partitions and rebuilt switched partitions without a local prefix. Add a switched-topic regression test with empty remote metadata.

giuseppelillo: Fixed in c950bc181d. RLM still prefers a readable remote start. When remote metadata is empty, it now uses the local start only if the partition has a committed switch seal, localLogStartOffset is below that seal, and the local high watermark has reached it. Born-diskless partitions, incomplete prefixes, and rebuilt switched replicas whose local start is at the seal continue to defer. Added RLM and ReplicaManager regression coverage for these states.

storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
```diff
@@ -982,8 +987,32 @@ private void maybeUpdateLogStartOffsetOnBecomingLeader(UnifiedLog log) throws Re
                 // when the remote start is unreported, which would lock the wrong value in via the
                 // forward-only advance. No-op for classic topics.
                 OptionalLong override = logStartOffsetOverride.apply(topicIdPartition.topicPartition());
-                long logStartOffset = override.isPresent()
-                        ? override.getAsLong()
+                OptionalLong readableBootstrapStart = OptionalLong.empty();
+                if (override.isEmpty() && isConsolidatingDisklessPartition.test(topicIdPartition.topicPartition())) {
+                    readableBootstrapStart = findReadableRemoteLogStartOffset(log);
```
viktorsomogyi: Seems like there is still an issue where a later finished segment can lock the start above an uncopied classic prefix. 

AI review comment: If the scan returns the first COPY_SEGMENT_FINISHED start and then skips the local-prefix check. listRemoteLogSegments(tp, epoch) is ordered by startOffset and includes unreferenced COPY_SEGMENT_STARTED and DELETE_SEGMENT_* entries, so both [STARTED 0–99, FINISHED 100–199] and [FINISHED 100–199] alone report 100. The second case is a switched leader that still has [localStart, seal) locally while an earlier leader, whose log already started at the seal, finished a suffix segment. updateRemoteLogStartOffset enqueues 100 on the forward-only control-plane path and calls updateLogStartOffsetFromRemoteTier. maybeIncrementLogStartOffset moves the log start to 100, and isSegmentEligibleForDeletion then allows those local segments to be deleted because the log start moved, even though they were never copied. ListOffsets(EARLIEST) cannot return below 100 after that.

giuseppelillo: Fixed in the follow-up commit 9b88d8d5ed without amending the previous commit. Bootstrap now computes the minimum readable start across the complete local classic prefix and all COPY_SEGMENT_FINISHED metadata. It also tracks the lowest COPY_SEGMENT_STARTED offset and defers if that offset is below the readable candidate, since the control-plane value cannot move backward. Regression coverage now verifies both cases: an in-progress prefix before a finished suffix, and a local classic prefix before a finished remote suffix.

storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java
```diff
@@ -992,6 +1020,34 @@ private void maybeUpdateLogStartOffsetOnBecomingLeader(UnifiedLog log) throws Re
             }
         }
 
+        private OptionalLong findReadableBootstrapStart(UnifiedLog log, OptionalLong localStart)
```
viktorsomogyi: A comment from grok 4.7:
Any COPY_SEGMENT_STARTED segment whose start is below the local prefix or the first finished segment suppresses the report on every copy tick. A new leader does not resume another leader's copy, and segmentIdsBeingCopied only tracks copies this task currently owns. The expiration task removes DELETE_SEGMENT_STARTED segments immediately, but it leaves COPY_SEGMENT_STARTED in place unless retention already breaches them. Retention will not move the log start past that segment while this bootstrap keeps remote_log_start_offset unset. A copy left in COPY_SEGMENT_STARTED by a crashed leader can therefore defer forever, and ListOffsets(EARLIEST) keeps coalescing to the WAL start.

storage/inkless/src/test/java/io/aiven/inkless/control_plane/AbstractControlPlaneTest.java
```diff
@@ -1905,6 +1905,67 @@ void realTimestamp() {
         }
     }
 
+    @Nested
+    class PruneCrossTierBootstrap {
```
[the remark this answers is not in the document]
giuseppelillo: Added in 289bb19bf4. The shared control-plane contract now initializes a switched partition with a classic start below the seal and covers both requested shapes: a remaining batch based at the seal, and an empty WAL with its high watermark at the seal. Both PostgreSQL and memory verify that the raw cross-tier start and `EARLIEST` remain at the classic start while `log_start_offset` and `EARLIEST_LOCAL` advance to the seal. `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./gradlew :storage:inkless:test --tests '*SwitchedPruneCrossTierBootstrap*' --no-daemon --no-parallel` passes all four cases.

viktorsomogyi (approved):

